PR #21, opened 2020-02-29
Improvements to install.sh

kolbe:
Fixes to some quoting errors and POSIX problems in install.sh.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,23 +1,33 @@
 #!/bin/sh
 
 repo='https://tiup-mirrors.pingcap.com'
-case $(uname) in
+case $(uname -s) in
     Linux|linux) os=linux ;;
     Darwin|darwin) os=darwin ;;
-    *) os=unknown ;;
+    *) os= ;;
 esac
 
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -1,31 +1,55 @@
...
+        echo "$PROFILE has been modified to to add tiup to PATH"
+        echo "open a new terminal or ${bold}source ${PROFILE}${sgr0} to use it"
+        ;;
+esac
 
-echo "tiup is installed in ${TIUP_HOME}/bin/tiup"
-echo "we have modify ${PROFILE} to add tiup to PATH"
-echo "you can open a new terminal or \033[1msource ${PROFILE}\033[0m to use it"
\ No newline at end of file
+echo "tiup is installed in $bin_dir/tiup"
```
lonng: @breeswish suggests we print `tiup run playground` to the console at the end of the installation.

kolbe: Ok! That sounds like a separate PR to me 🙂

lonng: Thanks

PR #43, opened 2020-03-07
Created and moved some variables to facilitate setting build-time defaults

kolbe:
This PR changes the way the profile directory is populated. The new behavior should match the old behavior for the default case.

`TIUP_HOME` was not being propagated correctly because of the order of environment assignment; that's fixed in this PR.

This PR adds DefaultTiupHome and ProfileDirName variables to pkg/localdata/constant.go to allow a build-time ldflags -X flag to set the variable. The Makefile adds EXTRA_LDFLAGS to further support this. This is useful for environments like homebrew where the tiup binary itself should be built to have a default directory outside the home directory of the installing user.

This new behavior can be achieved by executing something like this:
```
make EXTRA_LDFLAGS='-X "github.com/pingcap-incubator/tiup/pkg/localdata.DefaultTiupHome=/usr/local/var/tiup"'
```


kolbe:
I also split out "cmd" and "check" in Makefile, because builds were breaking in the homebrew environment where golint wasn't installed. "default" still builds check and cmd, so no change unless you explicitly do "make cmd", which will now skip "check".

lonng:
@kolbe Thanks for your contribution.

PR #61, opened 2020-03-14
mod: add wizard for playground

c4pt0r:
### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/components/playground/main.go
+++ b/components/playground/main.go
@@ -22,6 +22,10 @@ import (
 	"github.com/spf13/cobra"
 )
 
+type ValicationFunc func(interface{}) error
+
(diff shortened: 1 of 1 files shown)


components/playground/main.go
```diff
@@ -20,6 +22,10 @@ import (
 	"github.com/spf13/cobra"
 )
 
+type ValicationFunc func(interface{}) error
```
lonng:
```suggestion
type ValidationFunc func(interface{}) error
```

PR #70, opened 2020-03-17
clusterinfo support: Try to register prometheus.

mapleFU:
Signed-off-by: mapleFU <1506118561@qq.com>

Register Prometheus message to pd's etcd. Fix this issue: https://github.com/pingcap-incubator/tidb-dashboard/issues/222

Add codes in playground.

### Check List 

Tests 

 - No code

Code changes

Side effects

 - Increased code complexity


components/playground/main.go
```diff
@@ -17,6 +17,11 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"go.uber.org/zap"
```
lonng: Please optimize the imports.

components/playground/main.go
```diff
@@ -287,6 +289,12 @@ func bootCluster(version string, pdNum, tidbNum, tikvNum int, host string, monit
 			return err
 		}
 	}
```
breeswish: Line 275 ~ Line 283 can be removed?

mapleFU: Yes, I'll remove them.

mapleFU: Updated, PTAL.

Issue #64, opened 2020-03-15
UCP: Support specify configuration file of TiDB/TiKV/PD when starting playground


lonng:
## Description

Currently, we cannot specify the configuration file for the internal component of the playground, we should support this feature.

e.g `tiup playground --db.config ~/conf/db.config.toml --kv.config ~/conf/kv.config.toml`

## Score

* 100

## Mentor(s)

* @lonng

Contact the mentors: **#tidb-challenge-program** channel in [TiDB Community](https://join.slack.com/t/tidbcommunity/shared_invite/enQtNzc0MzI4ODExMDc4LWYwYmIzMjZkYzJiNDUxMmZlN2FiMGJkZjAyMzQ5NGU0NGY0NzI3NTYwMjAyNGQ1N2I2ZjAxNzc1OGUwYWM0NzE)  Slack Workspace

## Recommended Skills

* Go

lonng:
@chenlx0 Sorry for that our bot hasn't been configured correctly, we have fixed it. Please pick this issue again.

chenlx0:
Hi, i'd like to check some points here.

In the configuration file, we just need to specify host, port, status port and number of TiDB/TiKV/pd. Is that right?

lonng:
The configuration file is specified by user, eg:
```
tiup playground --kv 3 --db 3 --pd 3 --pd.config ~/custom.pd.toml
```

This task allows the users to specify the configuration file for the specific component.

PR #80, opened 2020-03-24
Support specify configuration file of TiDB/TiKV/PD when starting playground

chenlx0:
UCP #64 

Support specify configuration file of TiDB/TiKV/PD when starting playground. for #64 

Add flags `db.config` `pd.config` `kv.config` for playground, and pass their values to TiKV/TiDB/pd command as configuration file path when starting playground. If these flags are not specified, use default config.

### Check List 

Tests 

 - No code


chenlx0:
PTAL @lonng

components/playground/main.go
```diff
@@ -73,23 +73,27 @@ func execute() error {
...
 if you don't specified a version.
 
 Examples:
-  $ tiup playground nightly                         # Start a TiDB nightly version local cluster
-  $ tiup playground v3.0.10 --db 3 --pd 3 --kv 3    # Start a local cluster with 10 nodes
-  $ tiup playground nightly --monitor               # Start a local cluster with monitor system`,
+  $ tiup playground nightly                                                           # Start a TiDB nightly version local cluster
+  $ tiup playground v3.0.10 --db 3 --pd 3 --kv 3                                      # Start a local cluster with 10 nodes
+  $ tiup playground nightly --monitor                                                 # Start a local cluster with monitor system
+  $ tiup playground --pd.config ./pd.toml --db.config ./db.toml --kv.config ./kv.toml # Start a local cluster with specified configuration file`,
```
lonng: I think we don't need to pass all config flags in examples. e.g
```
tiup playground --pd.config ./pd.tom
```

components/playground/instance/pd.go
```diff
@@ -61,12 +62,16 @@ func (inst *PDInstance) Start(ctx context.Context, version repository.Version) e
...
 		"tiup", compVersion("pd", version),
 		"--name=" + uid,
-		fmt.Sprintf("--data-dir=%s", filepath.Join(inst.Dir, "data")),
-		fmt.Sprintf("--peer-urls=http://%s:%d", inst.Host, inst.Port),
-		fmt.Sprintf("--advertise-peer-urls=http://%s:%d", inst.Host, inst.Port),
-		fmt.Sprintf("--client-urls=http://%s:%d", inst.Host, inst.StatusPort),
-		fmt.Sprintf("--advertise-client-urls=http://%s:%d", inst.Host, inst.StatusPort),
-		fmt.Sprintf("--log-file=%s", filepath.Join(inst.Dir, "pd.log")),
+	}
+	if inst.ConfigPath != "" {
```
lonng: We can remove other flags even if we specify a configuration file. We should keep the original flags and append `--config` if the config flag exists.

components/playground/instance/tidb.go
```diff
@@ -59,12 +60,16 @@ func (inst *TiDBInstance) Start(ctx context.Context, version repository.Version)
...
 	args := []string{
 		"tiup", compVersion("tidb", version),
-		"-P", strconv.Itoa(inst.Port),
-		"--store=tikv",
-		fmt.Sprintf("--host=%s", inst.Host),
-		fmt.Sprintf("--status=%d", inst.StatusPort),
-		fmt.Sprintf("--path=%s", strings.Join(endpoints, ",")),
-		fmt.Sprintf("--log-file=%s", filepath.Join(inst.Dir, "tidb.log")),
+	}
+	if inst.ConfigPath != "" {
```
lonng: ditto

components/playground/main.go
chenlx0: Fixed.

components/playground/instance/pd.go
chenlx0: Thank you. Fixed.

components/playground/main.go
```diff
@@ -82,14 +85,15 @@ if you don't specified a version.
...
+  $ tiup playground nightly --monitor               # Start a local cluster with monitor system
+  $ tiup playground --pd.config ~/config/pd.toml    # Start a local cluster with specified configuration file`,
 		SilenceUsage: true,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			version := ""
 			if len(args) > 0 {
 				version = args[0]
 			}
-			return bootCluster(version, pdNum, tidbNum, tikvNum, host, monitor)
+			return bootCluster(version, pdConfigPath, dbConfigPath, kvConfigPath, pdNum, tidbNum, tikvNum, host, monitor)
```
lonng: The signature of function `bootCluster` too complex to hard to maintain, you can define a struct, eg.
```go
type bootOptions {
    version...
}
```

components/playground/main.go
```diff
@@ -147,7 +154,7 @@ func hasDashboard(pdAddr string) bool {
 	return false
 }
 
-func bootCluster(version string, pdNum, tidbNum, tikvNum int, host string, monitor bool) error {
+func bootCluster(version, pdConfigPath, dbConfigPath, kvConfigPath string, pdNum, tidbNum, tikvNum int, host string, monitor bool) error {
```
lonng: I think we should make further handling for the `pdConfigPath/dbConfigPath/kvConfigPath` and convert it to an absolute path. 
According to https://github.com/pingcap-incubator/tiup/blob/e08d28e360c6aee5044a9308c74eb994281be4e4/cmd/run.go#L198. Them will be related to the playground work directory, which is not same as the working directory of the user running `tiup playground`.

We can achieve our intention via the following code snippet.```
if !strings.HasPrefix(pdConfigPath, "/") && strings.HasPrefix(pdConfigPath, "~") {
    pdConfigPath = filepath.Join(os.Getenv(localdata.EnvNameWorkDir), pdConfigPath)
}
```

components/playground/main.go
chenlx0: OK, bootOptions added.

components/playground/main.go
chenlx0: @fredchenbj define a method `getAbsolutePath` in #82 , i think i can reuse that method here. But redefine that method here seems redundant and could cause conflicts. Maybe it is better to fix this problem after #82 merged so that i can use `getAbsolutePath` here. Is that OK?

lonng: Yes, merge #82 first.

chenlx0:
All fixed, @lonng PTAL

PR #82, opened 2020-03-25
specify component binary path

fredchenbj:
UCP https://github.com/pingcap-incubator/tiup/issues/77

Support specifies the binary path when running a component.

it runs like `tiup playground --kv 3 --db 3 --pd 3 --db.binpath /xx/tidb-server --kv.binpath /xx/tikv-server --pd.binpath /xx/pd-server`

Tests 

 - Manual test

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -15,14 +15,12 @@ package cmd
 
 import (
 	"fmt"
-	"os"
-	"strings"
-
 	"github.com/fatih/color"
 	"github.com/pingcap-incubator/tiup/pkg/meta"
 	"github.com/pingcap-incubator/tiup/pkg/repository"
 	"github.com/pingcap-incubator/tiup/pkg/version"
 	"github.com/spf13/cobra"
+	"os"
 )
 
 var rootCmd *cobra.Command
@@ -82,15 +80,7 @@ the latest stable version will be downloaded from the repository.
 				componentSpec := args[0]
 				for i, arg := range os.Args {
 					if arg == componentSpec {
-						var num = 1
-						if len(os.Args) > i+1 && strings.HasPrefix(os.Args[i+1], "--binpath=") {
-							num += 1
-							flagLen := len("--binpath=")
-							if len(os.Args[i+1]) > flagLen {
-								binPath = os.Args[i+1][flagLen:]
-							}
-						}
-						transparentParams = os.Args[i+num:]
+						transparentParams = os.Args[i+1:]
 						break
 					}
 				}
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -16,6 +16,7 @@ package cmd
 import (
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/fatih/color"
 	"github.com/pingcap-incubator/tiup/pkg/meta"
@@ -31,6 +32,7 @@ func init() {
 
 	var (
 		binary   string
+		binPath  string
 		tag      string
 		rm       bool
 		repoOpts repository.Options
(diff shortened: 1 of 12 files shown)


fredchenbj:
@lonng PTAL

lonng:
I think we shouldn't change the protocal component:version. Prefer to use a flag --binpath to specify the binary path.

The protocal component:version:binpath could make the user confused. And will introduce more corner case like component:binpath etc...

The protocal component:version can be used in install/update/uninstall.

fredchenbj:
> I think we shouldn't change the protocal component:version. Prefer to use a flag --binpath to specify the binary path.
> 
> The protocal component:version:binpath could make the user confused. And will introduce more corner case like component:binpath etc...
> 
> The protocal component:version can be used in install/update/uninstall.

I have modified it, please review again. @lonng

cmd/root.go
```diff
@@ -75,15 +77,24 @@ the latest stable version will be downloaded from the repository.
...
 				// e.g: tiup --tag mytag --rm playground --db 3 --pd 3 --kv 4
 				//   => run "playground" with parameters "--db 3 --pd 3 --kv 4"
+				// tiup --tag mytag --binpath /xxx/tikv-server tikv
 				var transparentParams []string
 				componentSpec := args[0]
 				for i, arg := range os.Args {
 					if arg == componentSpec {
-						transparentParams = os.Args[i+1:]
+						var num = 1
+						if len(os.Args) > i+1 && strings.HasPrefix(os.Args[i+1], "--binpath=") {
```
lonng: I think we should not filter the `--binpath` in this way because of the `--binpath` has been parsed by the cobra.
The syntax of running a component is `tiup [flags] <component> [args]`.

components/playground/instance/pd.go
```diff
@@ -53,13 +53,14 @@ func (inst *PDInstance) Join(pds []*PDInstance) *PDInstance {
...
 // Start calls set inst.cmd and Start
-func (inst *PDInstance) Start(ctx context.Context, version repository.Version) error {
+func (inst *PDInstance) Start(ctx context.Context, version repository.Version, binPath string) error {
 	if err := os.MkdirAll(inst.Dir, 0755); err != nil {
 		return err
 	}
 	uid := fmt.Sprintf("pd-%d", inst.ID)
 	args := []string{
 		"tiup", compVersion("pd", version),
+		fmt.Sprintf("--binpath=%s", binPath),
```
lonng: The order or arguments should be `"tiup", fmt.Sprintf("--binpath=%s", binPath), compVersion("pd", version)`

components/playground/main.go
```diff
@@ -147,19 +155,33 @@ func hasDashboard(pdAddr string) bool {
...
 }
 
-func bootCluster(version string, pdNum, tidbNum, tikvNum int, host string, monitor bool) error {
+func bootCluster(version string, pdNum, tidbNum, tikvNum int, host string, monitor bool, tidbBinPath, tikvBinPath, pdBinPath string) error {
 	if pdNum < 1 || tidbNum < 1 || tikvNum < 1 {
 		return fmt.Errorf("all components count must be great than 0 (tidb=%v, tikv=%v, pd=%v)",
 			tidbNum, tikvNum, pdNum)
 	}
 
+	var pathMap = make(map[string]string)
```
lonng: I think we should make further handling for the `*path` and convert it to an absolute path. 
According to https://github.com/pingcap-incubator/tiup/blob/e08d28e360c6aee5044a9308c74eb994281be4e4/cmd/run.go#L198. Them will be related to the playground work directory, which is not same as the working directory of the user running `tiup playground`.

We can achieve our intention via the following code snippet.
```
if !strings.HasPrefix(tidbBinPath, "/") && !strings.HasPrefix(pdConfigPath, "~") {
    tidbBinPath = filepath.Join(os.Getenv(localdata.EnvNameWorkDir), tidbBinPath)
}
```

fredchenbj: fixed, please look again.

components/playground/main.go
```diff
@@ -82,14 +85,16 @@ if you don't specified a version.
 Examples:
   $ tiup playground nightly                         # Start a TiDB nightly version local cluster
   $ tiup playground v3.0.10 --db 3 --pd 3 --kv 3    # Start a local cluster with 10 nodes
-  $ tiup playground nightly --monitor               # Start a local cluster with monitor system`,
+  $ tiup playground nightly --monitor               # Start a local cluster with monitor system
+  $ tiup playground --db.binpath /xx/tidb-server --pd.binpath /xx/pd-server --kv.binpath /xx/tikv-server
```
lonng:
```suggestion
  $ tiup playground --db.binpath /xx/tidb-server
```
I think we don't need to demo all flags in the examples, the users can learn from `tiup help playground`.

components/playground/main.go
```diff
@@ -147,19 +155,40 @@ func hasDashboard(pdAddr string) bool {
 	return false
 }
 
-func bootCluster(version string, pdNum, tidbNum, tikvNum int, host string, monitor bool) error {
+func getAbsolutePath(binPath string) string {
+	if !strings.HasPrefix(binPath, "/") {
```
lonng: I think we shouldn't join with `localdata.EnvNameWorkDir` if the user specifies the directory start with `~`, it is relative to home directory.

lonng:
@fredchenbj Thanks for your contribution, almost LGTM. It's better to add some integration tests in `tests` directory, you just should add some case to `tests/cases.json` and put the expected result to `tests/expected` directory.

fredchenbj:
> @fredchenbj Thanks for your contribution, almost LGTM. It's better to add some integration tests in `tests` directory, you just should add some case to `tests/cases.json` and put the expected result to `tests/expected` directory.

Thanks for your review. Adding integration test is ok, but I don't know the binary path or binary version of component when ci test.

PR #85, opened 2020-03-28
add message after updating

c4pt0r:


lonng:
Please add message for uninstall subcommand too, thanks.

PR #96, opened 2020-04-02
README: add some overview docs and instructions for building

nrc:
Add some docs for contributors to README.md

PTAL @lonng


lonng:
@nrc Thanks

PR #99, opened 2020-04-03
Refactor repository to enable more unitary unit tests

nrc:

The repository and mirror abstractions mixed IO and business logic. That means that unit tests were more like integration tests since they had to refer to files on disk and tested the whole pipeline of the repository.

Refactored repository and mirror (details in first commit message). Refactored their tests to be more unit-like.

### Check List 

 - Unit test

Code changes


pkg/repository/fileSource.go
```diff
@@ -0,0 +1,131 @@
+// Copyright 2020 PingCAP, Inc.
```
lonng, line 1: Please rename the file name to `file_source.go`, it's more Golang idiomatic.

pkg/repository/fileSource.go
```diff
@@ -0,0 +1,131 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package repository
+
+import (
+	"crypto/sha1"
+	"encoding/hex"
+	"encoding/json"
+	"github.com/pingcap-incubator/tiup/pkg/utils"
```
lonng, line 20: Please separate the 3rd dependencies into the individual import group.

pkg/repository/fileSource_test.go
```diff
@@ -0,0 +1,49 @@
+// Copyright 2020 PingCAP, Inc.
```
lonng, line 1: ditto

pkg/repository/fileSource.go
```diff
@@ -0,0 +1,131 @@
...
+		return errors.Trace(err)
+	}
+
+	if !expand {
+		return nil
+	}
+
+	tarFile, ok := tarReader.(*os.File)
+	if !ok {
+		return errors.Errorf("Expected file, found %v", tarReader)
```
lonng, line 109: Error string should start with lower case.

nrc:
All comments addressed, PTAL @lonng

Issue #108, opened 2020-04-08
When using --kv.config to specify a custom config file, it will be overwritten by the default config


HunDunDM:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?

`tiup playground nightly --kv.config=~/tiup/tikv.toml`

2. What did you expect to see?

Start a custom configured TiKV.

3. What did you see instead?

The custom config file is overwritten by the default config.

4. What version of TiUP are you using (`tiup --version`)?
tiup version v0.0.3 tiup

lucklove:
It's fixed in newest version (v1.0.7)

Issue #106, opened 2020-04-07
The cluster cannot be shutdown properly when using nohup


HunDunDM:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?

Execute in terminal
``` bash
nohup tiup playground nightly >tiup.log 2>&1 &
echo "$\!" >tiup.pid
kill `cat tiup.pid`
```

2. What did you expect to see?

```
Got signal interrupt (Component: tidb ; PID: XXX)
Got signal interrupt (Component: tikv ; PID: XXX)
Got signal interrupt (Component: playground ; PID: XXX)
Got signal interrupt (Component: pd ; PID: XXX)
```

3. What did you see instead?

Only the playground was stopped normally. PD stopped unexpectedly, TiDB and TiKV are still running.
```
Got signal terminated (Component: playground ; PID: XXX)
Error: start `~/.tiup/components/pd/nightly/pd-server` (wd:~/.tiup/data/RvTgZcK/pd-0) failed: exit status 1
```
They cannot be stopped by `tiup clean --all`, only by `kill <PID>`.

4. What version of TiUP are you using (`tiup --version`)?
tiup version v0.0.3 tiup

lonng:
Closed via #127

PR #127, opened 2020-04-08
Prevent tikv config from being overwritten and  Gracefully stop

HunDunDM:

* #106 
* #108

* Prevent tikv config from being overwritten
* Gracefully stop `tiup run`
* Process the signal in `playground`

### Check List 

Tests 

* Manual test (add detailed scripts or steps below)

1. sh run.sh
```bash
nohup tiup playground nightly --kv.config=./tikv.toml --db=1 --pd=1 --kv=1 >tiup.log 2>&1 &
echo "$!" > tiup.pid
```
2. sh stop.sh
```bash
kill `cat tiup.pid`
rm -f tiup.pid
```

### DIFF
--- a/components/playground/instance/tikv.go
+++ b/components/playground/instance/tikv.go
@@ -55,19 +55,9 @@ func (inst *TiKVInstance) Start(ctx context.Context, version repository.Version,
 	if err := os.MkdirAll(inst.Dir, 0755); err != nil {
 		return err
 	}
-	configPath := path.Join(inst.Dir, "tikv.toml")
-	if inst.ConfigPath != "" {
-		configPath = inst.ConfigPath
+	if err := inst.checkConfig(); err != nil {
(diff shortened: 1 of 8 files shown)


components/playground/instance/tikv.go
```diff
@@ -100,3 +90,29 @@ func (inst *TiKVInstance) Wait() error {
...
+		inst.ConfigPath = path.Join(inst.Dir, "tikv.toml")
+	}
+
+	_, err := os.Stat(inst.ConfigPath)
+	if err == nil || os.IsExist(err) {
+		return nil
+	}
+	if !os.IsNotExist(err) {
+		return errors.Trace(err)
+	}
```
lonng, line 105: Seems we can simplify this part to `utils.IsExist(inst.ConfigPath)`

Issue #111, opened 2020-04-08
Unit tests should only test single actions


nrc:
Some of our unit tests are more like integration tests at the moment in that they test more than one stage in the TiUp pipeline. It is better practice to test this sort of thing in integration tests and for unit tests to test small, modular actions. In particular, unit tests should almost never do IO. Doing this will require some mocking of interfaces and extracting interfaces where there are currently only concrete types

lucklove:
+1

PR #176, opened 2020-04-18
Refactoring for better testing

nrc:

This PR is a step towards making unit tests more unit-y and a better long-term test framework

cc #111

PTAL @lonng 

The first commit refactors the list component so that it can be unit tested without also testing the generic running framework or the table writing code.

The second commit refactors the `profile` and `repo` globals into an `Environment` struct. This removes global, mutable state and will allow for tests to run concurrently using different environments.

### Check List 

Tests 

 - Unit test
 - Integration test
 
(existing tests, all work is refactoring)

Code changes

Side effects

none

Related changes

none

### DIFF
--- a/components/playground/instance/instance.go
+++ b/components/playground/instance/instance.go
@@ -17,6 +17,8 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/pingcap-incubator/tiup/pkg/localdata"
+
 	"github.com/pingcap-incubator/tiup/pkg/repository"
 )
 
@@ -32,7 +34,7 @@ type instance struct {
 // Instance represent running component
 type Instance interface {
 	Pid() int
-	Start(ctx context.Context, version repository.Version, binPath string) error
+	Start(ctx context.Context, version repository.Version, binPath string, profile *localdata.Profile) error
 	Wait() error
 }
 
--- a/pkg/meta/meta.go
+++ b/pkg/meta/meta.go
@@ -16,42 +16,14 @@ package meta
 import (
 	"fmt"
 	"os"
-	"os/user"
 	"path/filepath"
 	"sort"
 	"strings"
 
 	"github.com/pingcap-incubator/tiup/pkg/localdata"
 	"github.com/pingcap-incubator/tiup/pkg/repository"
-	"github.com/pingcap/errors"
 	"golang.org/x/mod/semver"
 )
-
-var (
-	// profile represents the TiUP local profile
-	profile *localdata.Profile
-	// repo represents the components repository of TiUP, it can be a
-	// local file system or a HTTP URL
-	repo *repository.Repository
-)
-
-func init() {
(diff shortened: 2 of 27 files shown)


nrc:
If I understand the test output, the failure is due to code coverage shrinking, but this is intentional because the unit tests are testing less. `make test` passes locally

lonng:
@nrc 
image

components/playground/instance/instance.go
```diff
@@ -16,6 +16,7 @@ package instance
 import (
 	"context"
 	"fmt"
+	"github.com/pingcap-incubator/tiup/pkg/localdata"
```
lonng, line 19: Please format the imports.

pkg/meta/meta.go
```diff
@@ -62,110 +34,63 @@ func Mirror() string {
...
-func Repository() *repository.Repository {
-	return repo
+func (env Environment) Repository() *repository.Repository {
+	return env.repo
 }
 
 // SetRepository exports for test
-func SetRepository(r *repository.Repository) {
-	repo = r
+func (env Environment) SetRepository(r *repository.Repository) {
```
lonng, line 68:
```suggestion
func (env *Environment) SetRepository(r *repository.Repository) {
```

pkg/meta/meta.go
```diff
@@ -62,110 +34,63 @@ func Mirror() string {
...
-func Profile() *localdata.Profile {
-	return profile
+func (env Environment) Profile() *localdata.Profile {
+	return env.profile
 }
 
 // SetProfile exports for test
-func SetProfile(p *localdata.Profile) {
-	profile = p
+func (env Environment) SetProfile(p *localdata.Profile) {
```
lonng, line 78: The receiver must a pointer, or else the `set` will take no effect.
```suggestion
func (env *Environment) SetProfile(p *localdata.Profile) {
```

lonng:
Seems there are some integration tests unstable.
image

The expectation and the got result indeed the same. It's weird.

components/playground/instance/instance.go
nrc: Sorry, I always forget to do this. I've added goimports to `make fmt`, I hope that helps.

nrc:
The doc help failure is due to 00d033b0ec8b85d29ae0bb73024fc7d45781acc9, there is a fix in the new 3rd commit. I think the instability is due to using `Environment` and not `*Environment`.

pkg/meta/meta.go
```diff
@@ -14,45 +14,17 @@
 package meta
 
 import (
+	"github.com/pingcap-incubator/tiup/pkg/localdata"
```
lonng, line 17: Place all `std` library imports at the front of the 3rd will be more idiomatic.

nrc: I wish goimports did this, or even better, gofmt.

Issue #118, opened 2020-04-08
Collect versions of different components together as channels or releases


nrc:
Provide some abstraction for a group of versioned components that work well together (e.g., TiDB 4.0, TiKV 4.0, TiFlash 1.0, ... ). This would minimise version consistency errors and better support releases.

lucklove:
The [offline-package](https://pingcap.com/download-cn/) can do this now.

PR #182, opened 2020-04-21
Add requirements doc

nrc:
This PR is a start to the design docs section, adding a document for goals, requirements, and vision for the project. The doc is incomplete, I've just added my opinions on some areas. I'd really input from others with more context about how the tool is intended to be used.

PTAL @lonng @c4pt0r @lucklove @AstroProfundis

### DIFF
--- a/doc/design/requirements.md
+++ b/doc/design/requirements.md
@@ -12,24 +12,99 @@ We should distinguish between goals and requirements for the long term vs an ini
 * Help TiDB and TiKV developers run and test their code in a production-like environment
 * Make TiDB an easy to use database
 * Be the standard way for TiDB users to manage their clusters
-* Make it easy to try out TiDB, both on a single machine and a cluster
+* Make it easy to try out TiDB, on a single machine and a cluster
+
+## Use cases
+
+* A new user with no experience with TiDB can try out TiDB with minimal friction.
+* A developer of TiDB or TiKV can setup a local environment to test, benchmark, and debug their code.
+* An oncall DBA or developer can quickly setup a cluster matching one where a bug was reported.
+* A user reporting a bug can provide clear steps to reproduce based on a TiUp cluster configuration
+* A TiDB DBA can manage a production cluster, including TODO
+
 
 ## Requirements
 
+TODO Do we need to support different platforms? (E.g., mac/linux/ARM)
+
 ### Manage components
 
-* Install and update from a remote or local server, or from user-provided binaries
+* Install and update components
 * Uninstall/remove
 * Provide information about installed components and how to use them to the user
+* How must components be written? Restrictions on language, API? Sandboxing?
+* Components can be run directly from TiUp, components may run other components if they are dependencies (see versioning).
+  - E.g., running tidb can launch pd and tikv components
+
+#### Versioning
+
+* Components have an explicit version
+* Components may depend on other components and specify versions of those dependencies with which they are compatible
+* Components should use strict semver versioning
+* Component publishers will provide all versions
+* Version and semver compatibility is trusted, not verified
+* Users can specify which versions of each component to install
+* Component versions can be 'yanked', this means they are not listed and if the user tries to install them, they will be prompted for confirmation
+* Component versions cannot be deleted from servers
+* Component versions can be deleted locally (without removing the whole component)
+* Update
+  - patch revisions will be updated by default (i.e., the user must opt-in to keeping older versions)
+  - major and minor revisions will be duplicated by default (i.e., old versions are kept)
+  - when updating a component, by default, dependent components will be updated if necessary
+* Users can have multiple versions of components installed and can specify which version will be launched by default
+* Users can run any version of a component that they have installed
+* By default, TiUp will not let components run incompatible components, this may be forced by the user
+* Components may not specify which version of a dependent component they run.
+* TiUP will select which version to run (whether run by the user or another component) as follows:
+  - if the version is constrained by the user, that constraint has highest priority
+  - if the version is constrained by a (transitive-)dependee component, that constraint has second priority
+    - these constraints can be ignored if specified by the user
+  - if an override is specified by the user that overrides any version which fits the above constraints, then it is run
+  - otherwise, the most recent installed version of a component which satisfies the constraints is run
+  - if the constraints cannot be satisfied by any installed version, TiUp will show an error
+* The user may specify overrides of a component
+  - any component may be overridden
+  - an override may apply to any version, a specific versions, or a range of versions
+  - an override may use another version or another component (of the same or different version) or a local binary
+  - overrides apply to ways of running a component
+  - when the user runs a component explicitly, the user can cause the override to be ignored
+
+### Third-party components
+
+* Users should be able to publish their own components to public and private servers
+* Users can install third-party components and use them like the standard components
+* TiUp should manage version compatibility between components
+* Users should be able to search for components
+* Users who have published a component should be able to hide it (but not delete it), or do so with specific versions
+* Only the user who published a component should be able to publish new versions of that component
+* Users should be able to publish multiple components and be able to list them
 
 ### Self-update
 
-* Update TiUp from remote server
-* Uninstall
+* Update TiUp from remote and local servers
+* Uninstall TiUp
 
 ### Deployment and managing a deployed cluster
 
 TODO
+
+Better error messages when setting up and managing a cluster
+
+### Sources
+
+Also called mirrors and servers
+
+* PingCAP will host a publicly available server
+* Any third party may host a public or private server
+  - Users must opt-in to using a source
+  - Sources are strictly ordered for a given user (and can be changed by the user)
+  - Sources cannot be specified per-component
+  - All sources are checked in turn for a component
+  - If updating a component causes it to be installed from a different source, the user should be prompted
+* The user may create a local server, that server is created empty
+* Components can be cloned from any source to a local source, either a specific version, all versions, or a range of versions
+* TiUp can manage a server (possibly with a non-standard component); users should not have to manually edit any part of a server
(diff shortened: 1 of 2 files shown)


nrc:
Tests are failing due to bugs already in the codebase. They are fixed in https://github.com/pingcap-incubator/tiup/pull/176.

doc/design/requirements.md
```diff
@@ -0,0 +1,69 @@
+# Requirements and goals
+
+Includes the overlapping projects of TiUp, tiup-cluster, and tiup-playground.
```
lonng, line 3: We use the word `TiUP` when treated it as a production (consistent with TiDB TiKV), and use the word `tiup` when used it in the command line context.

nrc:
One question: what are the requirements around version management? Other than using a user-supplied binary (and similar things like building from source), I think there are some options, from simplest to most complex:

* only support the latest version of each component
* only support a single version of each component, but any version can be installed
* support multiple versions
* support multiple versions on distinct channels. Channels here could mean many things, e.g., 3.0 vs 3.1 vs 4.0, latest stable vs latest beta vs latest commit (or nightly builds if we want to produce nightly builds).

Relatedly, there is the idea of bundling versions together (see #118), if a user switches from TiDB 3 to TiDB 4, should we automatically switch TiKV too? Or allow mix and match? (And there are options for defaults as well as the permitted behaviour).

lonng:
@nrc I prefer to support multiple versions. Currently, the TiUP support `tiup comp[:version]` to run s specific version of a component. I think there is no easy to support channel like v3/v4 because not all components can be bonded to the TiDB release version.

e.g:
We will implement the `tiup publish` component to allow the community to publish their component to our registry.

And another question is: can I use the v3 channel's component if I set the current channel to v4. If not, the user must remember which component belongs to which channel. I think this will not easy if there are many components.

I think the better way is to place all components into the same channel, and user uses a different version via `tiup comp[:version]`. And in most scenario, the user just needs to use the latest stable version via `tiup comp`.

nrc:
> We will implement the tiup publish component to allow the community to publish their component to our registry.

Hmm, this is a fairly big thing - it means we are kind of a TiDB version of Rustup + Cargo, not just Rustup. Good to know this in advance, we should think in detail how this will work before we start specifying the manifest, etc.

doc/design/requirements.md
```diff
@@ -0,0 +1,69 @@
...
+
+We should distinguish between goals and requirements for the long term vs an initial release.
+
+## Vision and high-level goals
+
+* Manage and deploy TiDB and TiKV clusters
```
lonng, line 9: * Expand the TiDB ecology (including but not limited to manage and deploy TiDB clusters)

doc/design/requirements.md
```diff
@@ -0,0 +1,69 @@
...
+* Provide information about installed components and how to use them to the user
+
+### Self-update
+
+* Update TiUp from remote server
+* Uninstall
+
+### Deployment and managing a deployed cluster
+
+TODO
```
lonng, line 32: There some documents but haven't been translated to English, I think we can link to the documents after we finished translating.

doc/design/requirements.md
```diff
@@ -0,0 +1,69 @@
...
+
+* Manage and deploy TiDB and TiKV clusters
+* Open source project, with a community of users and contributors
+* Help TiDB and TiKV developers run and test their code in a production-like environment
+* Make TiDB an easy to use database
+* Be the standard way for TiDB users to manage their clusters
+* Make it easy to try out TiDB, both on a single machine and a cluster
+* Extensible
+
+## Requirements
```
lonng, line 17: I think the requirements can be expand the following core ideas:
1. We hope users can try it out as soon as they know TiDB.
2. Some developers can set up a local debug environment to adapt their application.
3. We can quickly set up a cluster in any version to reproduce the BUG reported by the ONCALL or GitHub issue.
4. The users can provide a reproduce easily via `tiup`, ref: https://github.com/pingcap/tidb/issues/737.
5. We should provide a component to make operate production clusters easier.
6. We should output a more human-friendly error message when some unexpected occur.

doc/design/requirements.md
nrc: Sorry, I don't understand "Expand the TiDB ecology"

doc/design/requirements.md
nrc: "We should output a more human-friendly error message when some unexpected occur." - could you give an example of this please?

lonng: The initial idea we want to re-implement the `tiup-cluster` is that the ansible cannot emit a human-friendly error message, and the users cannot figure out the root cause of why the procedure abort.

@breeswish introduces a new error message framework (https://github.com/pingcap-incubator/tiup-cluster/blob/master/cmd/cluster/command/root.go#L108 /  https://github.com/pingcap-incubator/tiup-cluster/pull/116). We can attach more details and clear error message context, and give some fixed steps to help the users learn the reason for errors and how to fix them.

doc/design/requirements.md
lonng: I hope we can build a bigger TiDB ecology base on the TiUP because we allow the community to upload components developed by themself, and we think some ideas we cannot think of, but someone in the community can.

doc/design/requirements.md
nrc: Why TiUP not TiUp? I think the second is better since 'up' is one word like 'Ti', not an acronym like 'KV' or 'DB'

nrc:
Added another commit with some more details

doc/design/requirements.md
```diff
@@ -0,0 +1,142 @@
...
+* Be the standard way for TiDB users to manage their clusters
+* Make it easy to try out TiDB, on a single machine and a cluster
+
+## Use cases
+
+* A new user with no experience with TiDB can try out TiDB with minimal friction.
+* A developer of TiDB or TiKV can setup a local environment to test, benchmark, and debug their code.
+* An oncall DBA or developer can quickly setup a cluster matching one where a bug was reported.
+* A user reporting a bug can provide clear steps to reproduce based on a TiUp cluster configuration
+* A TiDB DBA can manage a production cluster, including TODO
```
lonng, line 23: Is it WIP?

nrc: I don't know all the requirements for tiup-cluster or what people do at the moment to manage their clusters, so I hope somebody else can help with this part.

doc/design/requirements.md
lucklove: I think we have n’t considered it carefully enough and use the word `TiUP` directly in all documents... Nick is right but it ’s weird to fix this because that many users already know this product and it's name.

doc/design/requirements.md
```diff
@@ -0,0 +1,142 @@
...
+## Requirements
+
+### Manage components
+
+* Install and update components
+* Uninstall/remove
+* Provide information about installed components and how to use them to the user
+* How must components be written? Restrictions on language, API? Sandboxing?
+* Components can be run directly from TiUp, components may run other components if they are dependencies (see versioning).
+  - E.g., running tidb can launch pd and tikv components
```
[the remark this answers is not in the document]
nrc: This document is meant to describe what the software must do, not how it must do it, so running `tiup playground` and have it run tidb and its dependents satisfies this requirement (for tidb).

doc/design/requirements.md
```diff
@@ -0,0 +1,142 @@
...
+* Component publishers will provide all versions
+* Version and semver compatibility is trusted, not verified
+* Users can specify which versions of each component to install
+* Component versions can be 'yanked', this means they are not listed and if the user tries to install them, they will be prompted for confirmation
+* Component versions cannot be deleted from servers
+* Component versions can be deleted locally (without removing the whole component)
+* Update
+  - patch revisions will be updated by default (i.e., the user must opt-in to keeping older versions)
+  - major and minor revisions will be duplicated by default (i.e., old versions are kept)
+  - when updating a component, by default, dependent components will be updated if necessary
```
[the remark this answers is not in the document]
nrc: I don't think this introduces complexity, since we already need to keep track of dependencies for running the components or telling the user and we want the ability to update multiple components in any case too (update all), so supporting update of a dependency tree seems not too much extra. It is also fine for this to be low priority and to not support it for some time (I have not tried to prioritise requirements at this stage).

doc/design/requirements.md
```diff
@@ -0,0 +1,146 @@
...
+* Scale - should support our largest users, i.e., 1000s of nodes in a cluster and nodes in different data centers.
+* Backwards compatibility - TODO how much do we have to preserve workflows and scripts between versions?
+
+### Non-goals
+
+Things we intend not to do, either for now or ever.
+
+* Designed to be primarily used by humans, secondarily by scripts. Not intended as an API.
+* No GUI.
+* Project is specific to TiDB and TiKV, not a generic database or software management tool.
```
lucklove, line 144: I'm not sure about this. If the user provides a component that can manage MySQL or other software, it can manage these Softwares. FYI @lonng

doc/design/requirements.md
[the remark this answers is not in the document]
lucklove: Currently we have not tracked dependencies yet. TiUP doesn't know `playgrouond` dependent on `tidb/tikv/pd`. When you execute `tiup playground`, it's playground itself that download tidb/tikv/pd.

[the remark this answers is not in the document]
nrc: How does playground know where to download from? Is there API in TiUp to get the location of the mirror?

I'll rephrase this to make it more implementation agnostic (i.e., it can be implemented in the component or TiUp).

doc/design/requirements.md
nrc: That sounds fine. The way I would frame it is that if this works we don't try and stop it, but for now it is not a goal. Does that sound ok?

nrc:
I've addressed comments and added some explicit priorities. PTAL @lucklove @lonng

Issue #121, opened 2020-04-08
Permanent URLs


nrc:
All artifacts (other than the top-level manifest) should have permanent URLs, that is once a resource is available at a URL, it should never change, nor should it ever disappear.

See [post about similar issues in Rustup](https://internals.rust-lang.org/t/future-updates-to-the-rustup-distribution-format/4196)

lonng:
@nrc Now this method is the same for offline mirroring and online mirroring, and it is relatively simple. If we use Permanent URLs, this seems not easy to handle.

lucklove:
For the same mirror, it does have permanent URLs, I think. The cloned offline-mirror is a different mirror from the online one, so their URLs are not the same.

Issue #120, opened 2020-04-08
Manifest structure


nrc:
We should properly design the various manifests. We might want to follow TUF (see #114). In any case we should consider the following points:

* Manifests should contain the URLs of components
* Manifests should contain the hashes of components
* Manifests should be read-only: they should have permanent URLs and should not be updated (other than the top-level manifest which specifies the current versions and should be as small as possible).

See also #117 #119

lucklove:
I think this is finished

Issue #119, opened 2020-04-08
Manifest version numbers


nrc:
To better support backwards compatibility, every manifest should have a version number.

lucklove:
We have it now

Issue #117, opened 2020-04-08
Manifests should be .json files instead of .index


nrc:
Since they are json.

PR #193, opened 2020-05-05
Manifest and server organisation design doc

nrc:
~~Still some TODOs to complete and needs a good check over. Uploading early because it is taking quite a while.~~

cc #114, #117, #119, #120, #121,

### DIFF
--- a/doc/design/manifest.md
+++ b/doc/design/manifest.md
@@ -1,0 +1,589 @@
+# Manifest format and repository layout
+
+See [#120](https://github.com/pingcap-incubator/tiup/issues/120).
+
+## The Update Framework
+
+[The Update Framework (TUF)](https://theupdateframework.io/overview/) is the state of the art for update systems. It has a strong focus on security.
+
+TUF is a [specification](https://github.com/theupdateframework/specification/blob/master/tuf-spec.md#the-update-framework-specification) for building update systems, not an implementation (though reference implementations exist). It does not fully specify how such a system should work, e.g., it does not discuss how updates are applied or how servers are managed. However, it defines what metadata should be kept, how data and metadata should be signed, and how to ensure downloads are secure.
+
+We build on and adapt TUF, all places where we deviate from TUF are enumerated below. I believe we still fulfil all TUF security guarantees.
+
+
+## Components
+
+A component is a program which can be run by TiUp. In this document, we do not consider what constitutes a component, for our purposes it is an opaque file.
+
+A component may have many versions. We do not discuss how versions are specified or applied, only that a client may specify a version to download.
+
+A component may be supplied by PingCAP or a third party, we detail how components are *uploaded*, but how components are packaged and checked by the client is out of scope. We assume that a component is not 'opened', i.e., a component is uploaded to a server and later downloaded, the server does not inspect the contents of the component. Note that this design document only addresses part of the security issue with handling third party code or binaries - we ensure that the code that a user downloads is the same code that the publisher uploads, however, we cannot ensure that the code uploaded is not malicious.
+
+We do not support channels or groupings of components. We expect components will support something here, but it will be up to the client to implement.
+
+TiUp itself can be treated as a component and I think does not need to be treated any differently on the server side.
+
+## Other terms
+
+* repository: a source of components and manifests (aka server, mirror)
+* manifest: a JSON metadata file
+* platform: the hardware platform and operating system where a binary will run (c.f., built).
+
+
+## Principles
+
+* URLs are persistent.
+* All data is immutable
+* Most metadata is immutable, achieved using versioning.
+* Components cannot be removed or deleted.
+* Zero-trust access control.
+* Minimise the number of keys and the frequency with which they must be replaced.
+* Uploads and downloads should be secure.
+* All operations should be idempotent or at least repeatable. If an operation is interupted, the user can just retry until it succeeds.
+* There should be no 'transaction' mechanism between client and server, groups of operations should not require atomicity.
+* Minimise required knowledge in the client - the metadata should direct the client how to perform operations with minimal knowledge of the server's data layout, etc.
+* We should not require consistency between files in the repo (beyond hashes and keys).
+
+
+## Assumptions
+
+* Updates are per-component, updating all components is interpreted by the client.
+* There is a secure channel to distribute initial and updated keys for each server
+* Don't need to change the owner of a component.
+* Don't need to delegate ownership of a component (owners can delegate themselves with client-side tooling).
+
+
+## Repository organisation
+
+We require snapshot.json and root.json to exist at the root of the repository. There will be many n.root.json files, these must also be at the repository root.
+
+The rest of the repository organisation can be changed freely since all other locations are specified by URLs. All files and manifests should be in the repository root, i.e., there is a single directory.
+
+Each non-manifest file has the form `id-version-platform.ext`, where `ext` is `tar.gz` for tarballs, `id` is the component id, `platform` is a target triple, and `version` is a semver version containing exactly two periods (`.`, e.g., `1.0.0`).
+
+
+## Manifests
+
+Each manifest is a [canonical json](http://wiki.laptop.org/go/Canonical_JSON) file with a `.json` extension.
+
+Manifest format:
+
+```
+{
+    "signatures": [
+        {
+            "keyid": "",
+            "sig": "",
+        }
+    ],
+    "signed": {
+        "_type": "TY",
+        "spec_version": "",
+        "expires": "TS",
+        // version is only present for versioned manifests.
+        "version": n,
+        ROLE
+    },
+}
+```
+
+where
+
+* `""` is a string.
+* `n` is an integer > 0.
+* `"TS"` is a timestamp (date and time) in ISO 8601 format; timezone should always be `Z`.
+* `"TY"` is one of `"root"`, `"index"`, `"component"`, `"snapshot"`, or `"timestamp"`.
+* `ROLE` is specific to the type of manifest.
+* The value of `"spec_version"` should be valid semver.
+
+The timestamp and snapshot manifests should expire after one day. All other manifests should expire after one year.
+
+URLs are relative to the repository root, which allows a repository to be cloned to a new location without changing URLs.
+
+The spec version should be the same for all files in a given snapshot, if the client finds a spec version which is inconsistent, or that the client cannot interpret, it should abort its operation.
+
+### Keys
+
+A key definition is:
+
+```json
+"KEYID": {
+    "keyid_hash_algorithms": [
+        "sha256",
+        "sha512",
+    ],
+    "keytype": "ed25519",
+    "keyval": {
+        "public": "KEY",
+    },
+    "scheme": "ed25519",
+},
+```
+
+where
+
+* `"KEYID"` is a globally unique identifier (i.e., keys must not share key ids); it is the hexdigest of the SHA-256 hash of the canonical JSON form of the key.
+* We support `"sha256"` and `"sha512"` hashing (only SHA256 in first iteration).
+* We support `"rsa"`, `"ed25519"`, and `"ecdsa-sha2-nistp256"` key types (only `"rsa"` in first iteration).
+* We support `"rsassa-pss-sha256"`, `"ed25519"`, and `"ecdsa-sha2-nistp256"` schemes (only `"rsassa-pss-sha256"` in first iteration)
+  - "rsassa-pss-sha256" : RSA Probabilistic signature scheme with appendix. The underlying hash function is SHA256. https://tools.ietf.org/html/rfc3447#page-29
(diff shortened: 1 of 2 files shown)


nrc:
And cc #113, #118

nrc:
Branch is based on https://github.com/pingcap-incubator/tiup/pull/182

nrc:
Ready for review; PTAL @lonng @lucklove @brson

doc/design/manifest.md
```diff
@@ -0,0 +1,585 @@
...
+                },
+                "scheme": "ed25519",
+            },
+        },
+    },
+},
+"components": {
+    "tidb": {
+        "name": "Name of the Component",
+        "yanked": false,
```
lucklove, line 241: What is the meaning of the word `yanked`?

doc/design/manifest.md
```diff
@@ -0,0 +1,585 @@
...
+
+Follow steps for one component workflow.
+
+### Add a new owner
+
+Probably an implicit action when a user publishes their first component.
+
+* Generate a new key pair in the client. The private key is stored securely.
+* Update root.json
+* Send new owner request from TiUp client to server including the client's public key but not private key; request must be sent over TLS to ensure that the public key is not tampered with.
```
lucklove, line 422: Where the public key comes from? And where we store it?

doc/design/manifest.md
nrc: It means 'pulled', a component or version which is yanked still exists on the server but is not listed in search functions so is not discoverable. It means that a component or version can be removed without breaking any users who depend on it.

doc/design/manifest.md
nrc: Line 420 - the client generates it; "key pair" = a public key and a private key. The public key is stored in index.json (the private key is stored by the client)

doc/design/requirements.md
```diff
@@ -0,0 +1,144 @@
...
+* A new user with no experience with TiDB can try out TiDB with minimal friction.
+* A developer of TiDB or TiKV can setup a local environment to test, benchmark, and debug their code.
+* An oncall DBA or developer can quickly setup a cluster matching one where a bug was reported.
+* A user reporting a bug can provide clear steps to reproduce based on a TiUp cluster configuration
+* A TiDB DBA can manage a production cluster, including TODO
+
+
+## Requirements
+
+TODO Do we need to support different platforms? (E.g., mac/linux/ARM)
```
AstroProfundis, line 28: For the client side, at least darwin and linux need to be supported.

doc/design/requirements.md
```diff
@@ -0,0 +1,144 @@
...
+This section is getting beyond requirements and should be removed, but might be useful for deriving requirements
+
+* SQL client
+* playground
+* cluster
+
+### Non-functional requirements
+
+* Security - we must be able to securely download files and manage clusters since the DB is likely to be security-critical for users. Download binaries is likely to be over the internet. Cluster management might include cross-data center communication on a public network. No need to protect data at rest.
+* Reliability - could bugs in cluster management cause severe issues? That might be important. I think delaying update is not critical as long as it is not for too long, perhaps <1hr?
```
AstroProfundis, line 129: I think this may be not much a critical issue, many users trends to delay upgrading of the software in production, especially infrastructures like DB server.

doc/design/manifest.md
lucklove: What if the user lost the key (eg. reinstall tiup)? Can he control his old packages?

nrc: There is a flow for restoring access, I should specify that keys may be backed up

doc/design/manifest.md
```diff
@@ -0,0 +1,585 @@
...
+
+Rotating keys is specified by [TUF](https://github.com/theupdateframework/specification/blob/master/tuf-spec.md#6-usage). Neither TUF nor TiUp will dictate how often this should happen. The TiUp server software should provide a facility for doing so, both for the root keys and for owner keys. The TiUp client should automatically rotate owner keys on a regular basis (question: how often?). TiUp server should ensure that all keys are unique, this should be verified by the client whenever it downloads a new manifest.
+
+### root.json and n.root.json
+
+```
+"_type": "root"
+```
+
+Indexes other top-level manifests and roles. n is the version, versioned files are immutable, root.json is a copy of the most recent version of n.root.json and is mutable.
```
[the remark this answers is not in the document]
nrc: Seems fine to me, assuming the web server handles symlinks correctly (in particularly they are not cached).

doc/design/manifest.md
```diff
@@ -0,0 +1,585 @@
...
+                "sha512": "ef5beafa16041bcdd2937140afebd485296cd54f7348ecd5a4d035c09759608de467a7ac0eb58753d0242df873c305e8bffad2454aa48f44480f15efae1cacd0"
+            },
+            "length": 1001000499,
+            "dependencies": {
+                "foo": "1.0.0",
+            },
+        },
+        "0.2.0": { ... },
+    },
+    "aarch64-unknown-linux-gnu": { ... },
```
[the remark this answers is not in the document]
nrc: We definitely don't need the toolchain! I'll remove that. The target triple is a fairly standard way to specify platforms.

doc/design/manifest.md
```diff
@@ -0,0 +1,585 @@
...
+* Use timestamp.json to download and verify snapshot.json
+* Use snapshot.json to download and verify index.json
+* Verify that the user's entry in index.json is as expected.
+* Client signs manifest and sends manifest and tar ball to the server
+* Server verifies the manifest using the clients public key from most recent index.json.
+* Verify that the component id is unique and that the platform is valid.
+* Create a component directory and platform sub-directory.
+* Store the component tar ball and manifest (with version number 1) in the new directories.
+* Return success to the client.
+* Client downloads timestamp.json, snapshot.json, index.json, and the new component manifest and verifies they are as expected.
```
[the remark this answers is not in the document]
nrc: It can retry or error out, depending on the circumstance, I'll add a note.

nrc:
Addressed comments

doc/design/manifest.md
```diff
@@ -0,0 +1,589 @@
...
+### n.xxx.json
+
+```
+"_type": "component"
+```
+
+Where xxx is the id of the component, n is the version of the manifest, file is immutable.
+
+```
+"name": "Name of the Component",
```
lonng, line 266: Prefer
```
"name": "Name of the Component",
"description": "This is a component",
"versions": {
    "0.0.1": {
        "amd64-darwin": {
            "yanked": false,
            "url": "/name-of-amd64-darwin-v0.0.1.tar.gz",
            "hashes": {
                "sha256": "141f740f53781d1ca54b8a50af22cbf74e44c21a998fa2a8a05aaac2c002886b",
                "sha512": "ef5beafa16041bcdd2937140afebd485296cd54f7348ecd5a4d035c09759608de467a7ac0eb58753d0242df873c305e8bffad2454aa48f44480f15efae1cacd0"
            },
            "length": 1001000499,
            "dependencies": {
                "foo": "1.0.0",
            },
        },
        "amd64-linux": {
            "yanked": false,
            "url": "/name-of/name-of_x86_64-apple-darwin_0.0.1.tar.gz",
            "hashes": {
                "sha256": "141f740f53781d1ca54b8a50af22cbf74e44c21a998fa2a8a05aaac2c002886b",
                "sha512": "ef5beafa16041bcdd2937140afebd485296cd54f7348ecd5a4d035c09759608de467a7ac0eb58753d0242df873c305e8bffad2454aa48f44480f15efae1cacd0"
            },
            "length": 1001000499,
            "dependencies": {
                "foo": "1.0.0",
            },
        }
    },
    "0.0.2":{ ... },
}
```

The reason why I prefer:
1. The CDN which we used doesn't support sub-directory and the additional advantage is we can keep the filename backward compatible (maybe we should support the current solution in short-terms).
2. Grouping distributed by version instead of platforms.

doc/design/manifest.md
```diff
@@ -0,0 +1,589 @@
...
+
+## Repository organisation
+
+We require snapshot.json and root.json to exist at the root of the repository. There will be many n.root.json files, these must also be at the repository root.
+
+The rest of the repository organisation is here for completeness but can be changed freely since all other locations are specified by URLs.
+
+n.index.json files and snapshot.json are at the root of the repository.
+
+For each component there is a directory named with the component's id. Within this directory are the n.id.json files (where id is the id of the component; alternative: all these manifests could be called n.component.json). Within the component directory, there is a sub-directory for each platform named with the target triple. All data files are within these platform subdirectories. Each file has the form version.platform.id.tar.gz for tarballs and version.platform.id.ext or version.platform.id for other single files, where id is the component id, platform is a target triple, and version is a semver version containing exactly two periods (`.`, e.g., `1.0.0`).
```
brson, line 64: The version-first naming scheme seems very strange. Id-first provides a more natural sorted order. I can't think of an advantage to version-first. `id.version.platform` makes more sense to me.

The subdirectory-per platform is redundant with the file-naming scheme. That's not necessarily a problem, since it could e.g. aid human comprehension.

Some rationale for both these choices would help me.

brson: Perhaps the version-first scheme is related to ease of parsing?

doc/design/manifest.md
```diff
@@ -0,0 +1,589 @@
...
+
+where
+
+* `"KEYID"` is a globally unique identifier (i.e., keys must not share key ids); it is the hexdigest of the SHA-256 hash of the canonical JSON form of the key.
+* We support `"sha256"` and `"sha512"` hashing (only SHA256 in first iteration).
+* We support `"rsa"`, `"ed25519"`, and `"ecdsa-sha2-nistp256"` key types (only `"rsa"` in first iteration).
+* We support `"rsassa-pss-sha256"`, `"ed25519"`, and `"ecdsa-sha2-nistp256"` schemes (only `"rsassa-pss-sha256"` in first iteration)
+  - "rsassa-pss-sha256" : RSA Probabilistic signature scheme with appendix. The underlying hash function is SHA256. https://tools.ietf.org/html/rfc3447#page-29
+  - "ed25519" : Elliptic curve digital signature algorithm based on Twisted Edwards curves. https://ed25519.cr.yp.to/
+  - "ecdsa-sha2-nistp256" : Elliptic Curve Digital Signature Algorithm with NIST P-256 curve signing and SHA-256 hashing. https://en.wikipedia.org/wiki/Elliptic_Curve_Digital_Signature_Algorithm
```
brson, line 133: There's no need to claim to support all these options, unless perhaps we're developing this as a general-purpose library for others to use.

brson:
It looks good in general, simply by following tuf. I haven't carefully reviewed all the flows, but the parts I have read seem plausibly correct.

The lack of channels and grouping seems like a big problem, but from the language in the intro I guess you see that as something to layer on top of the base protocol. e.g. I might have a meta-component called that is nothing but a manifest of component-versions.

doc/design/manifest.md
brson: There are conceivably platformless components too. Consider how a component with no target-triple would fit into this scheme.

nrc:
> The lack of channels and grouping seems like a big problem, but from the language in the intro I guess you see that as something to layer on top of the base protocol. e.g. I might have a meta-component that is nothing but a manifest of component-versions.

Yeah, I'm thinking that dependencies will be handled more like Cargo than like Rustup

doc/design/manifest.md
nrc: > The version-first naming scheme seems very strange. Id-first provides a more natural sorted order. I can't think of an advantage to version-first. id.version.platform makes more sense to me.

Good point, there was an older version where it was important for the version to be first but now it doesn't matter and the ordering is good.

> The subdirectory-per platform is redundant with the file-naming scheme. That's not necessarily a problem, since it could e.g. aid human comprehension.

Yep, rationale is purely to make it easier for manual administration, but it sounds like using just one directory is better for the CDN.

doc/design/manifest.md
nrc: I wasn't sure why TUF chose to support these three or their relative merits, so I erred on the side of caution. Supporting just one seems like a good idea, but I didn't know which one to choose! Even if we only choose one, I think it is good to have the flexibility to support others in case we need it in the future (it is not very much extra complexity).

doc/design/manifest.md
nrc: > The CDN which we used doesn't support sub-directory and the additional advantage is we can keep the filename backward compatible

That seems fine, using subdirectories is just to make it easier to look at, it is not necessary and can be changed by changing the URLs pretty easily.

> Grouping distributed by version instead of platforms.

The reason for platform first is that it makes things marginally easier to search since the platform is fixed - the client is only interested in one platform at a time - but the client must search the versions, e.g., find the most recent, so it is nice to have them together. Of course since there are unlikely to be thousands of versions, it is easy enough to convert in software.

Why do you prefer versions first?

Issue #201, opened 2020-05-10
refactor manifest handling


nrc:
Refactor manifest and component handling to follow the design in #193

Soft-blocked on: merging https://github.com/pingcap-incubator/tiup/pull/193

Tasks:

Group 1:

- [x] Tool to generate repo and manifests (#200, #210, #216)
- [x] Download and access root manifest (#224)
- [x] Update root manifest (#224)

Group 2:

- [x] High level manifest handling and shared ser/de (#205)
- [x] Save and load manifests from local storage (#209)
- [x] Download and update timestamp and snapshot manifests (#213)
- [x] Download and verify index.json; representation of the index in code(#219)
- [x] Download and represent component manifests(#219)
- [x] Use the new component manifest to find and verify components (#222)

Group 3:

- [x] Keys, key ids, signing, and hashing (#216)
- [x] Verify signing
- [x] Use canonical JSON (#216)

nrc:
Group 3 above is implementing all crypto parts of the design. That includes:

* reading the keys section of a manifest
* hashing a file using SHA256
* signing a string and verifying a string and its signature
* generating public/private key pairs

july2993:
I am taking a look at thinks about `Download and access root manifest` and will help implement it

AstroProfundis:
I'm working on `Update root manifest` in group 1 now

lonng:
Done

PR #205, opened 2020-05-11
repository: create some basic manifest infrastructure

nrc:
cc #201 this is ~~most of~~ the first task in group 2. ~~It is missing serialisation though.~~

PTAL @AstroProfundis 

It seems that we had not been running unit tests due to a bad argument to `go test`, this PR fixes that and adds TODOs where tests need fixing.


nrc:
Tests seem to hit https://github.com/pingcap/failpoint/issues/33

PR #210, opened 2020-05-12
cmd: implement basic entries of repository management

### LINKED ISSUE #200: Initial server-side admin component
Be able to generate file structure and manifests on the server; implement basic admin actions such as yanking components.

In the short term this lets us manage the repo, in the long term this will be the basis for the server-side code

~~Blocked on: https://github.com/pingcap-incubator/tiup/pull/193~~

Tasks:

- [ ] Create and initialise repo (#210)
- [ ] Update a component
- [ ] Yank a component
- [ ] Yank a version

AstroProfundis:
Close part of #200 

The whole feature is still in an early stage of development and is supposed not to work yet

Add a `repo` sub command for repository management.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes


cmd/repo_add.go
```diff
@@ -0,0 +1,42 @@
...
+
+func newRepoAddCompCmd(env *meta.Environment) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "add <component-id> <platform> <version> <file>",
+		Short: "Add a file to a component",
+		Long:  `Add a file to a component, and set its metadata of platform ID and version.`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) < 4 {
+				cmd.Help()
+				return nil
```
lonng, line 29: Prefer `return cmd.Help()`

cmd/repo_init.go
```diff
@@ -0,0 +1,152 @@
...
+
+	cmd.Flags().StringVar(&pubKey, "pubkey", "", "Path to the public key file")
+	cmd.Flags().StringVar(&privKey, "privkey", "", "Path to the private key file")
+
+	return cmd
+}
+
+func initRepo(path string) error {
+	currTime := time.Now()
+	repoManifests := localdata.NewManifests(path)
```
nrc, line 71: You should avoid using `Manifests` here - the way the way that manifests are saved locally is not the same as the way they are saved in a repo.

cmd/repo_init.go
```diff
@@ -0,0 +1,152 @@
...
+
+	// initial manifests
+	newManifests := make([]*repository.Manifest, 0)
+
+	// init the root manifest
+	root := &repository.Root{
+		SignedBase: repository.SignedBase{
+			Ty:          "root",
+			SpecVersion: "TODO",
+			Expires:     currTime.UTC().Add(time.Hour * 24 * 365).Format(time.RFC3339), // 1y
```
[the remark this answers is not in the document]
nrc: Small: you could move `.UTC()` to the declaration of `currTime`

cmd/repo_init.go
```diff
@@ -0,0 +1,152 @@
...
+		},
+	}
+
+	cmd.Flags().StringVar(&pubKey, "pubkey", "", "Path to the public key file")
+	cmd.Flags().StringVar(&privKey, "privkey", "", "Path to the private key file")
+
+	return cmd
+}
+
+func initRepo(path string) error {
```
nrc, line 69: I think it is good for the functions in `cmd` to do the minimum and mostly just call out to higher-level functions, in this case in manifest.go, that makes it easier to change the interface or add alternate interfaces and probably easier to test too.

With the changes I suggested, this function might end up looking much smaller, but if not, could you try and move more to manifest.go please?

cmd/repo_add.go
```diff
@@ -0,0 +1,41 @@
+// Copyright 2020 PingCAP, Inc.
```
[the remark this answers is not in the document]
AstroProfundis: I'm worring that if we implement much logic in the cmd/*.go file (it's highly possible IMO), the `repo.go` would be too long. But it's also ok to put those don't have long functions be in one file, I think we could put them in one file at first, and split to multiple if any of those sub-subcommands get long.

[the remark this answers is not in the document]
nrc: Yes, it’s true. I hope that if we move most of the logic to the repository package and just leave cli logic here, then it will not be too big

cmd/repo_init.go
```diff
@@ -0,0 +1,152 @@
...
+	currTime := time.Now()
+	repoManifests := localdata.NewManifests(path)
+	// TODO: set key store
+
+	// initial manifests
+	newManifests := make([]*repository.Manifest, 0)
```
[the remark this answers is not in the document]
AstroProfundis: I'm leaving the file writing part as it is for now, I think we could do that in a separate PR and include more changes than initializing repo, e.g., when making changes to an exist repo.

pkg/repository/manifest.go
```diff
@@ -56,27 +58,98 @@ type ValidManifest interface {
...
+		expire:    time.Hour * 24, // 1d
+		threshold: 1,
+	},
 }
 var knownVersions = map[string]struct{}{"0.1.0": {}}
 
+// Role is the metadata of a manifest
+type Role struct {
+	URL       string               `json:"url"`
+	Keys      map[string]*KeyStore `json:"keys"`
```
nrc, line 103: This shouldn't be a map to KeyStore, it should be some new KeyInfo struct or something like that

pkg/repository/manifest.go
```diff
@@ -56,27 +58,98 @@ type ValidManifest interface {
...
-	// TODO
+	Roles map[string]*Role `json:"roles"`
+}
+
+// NewRoot creates a Root object
+func NewRoot(initTime time.Time) *Root {
+	return &Root{
+		SignedBase: SignedBase{
+			Ty:          "root",
+			SpecVersion: "TODO",
```
nrc, line 118: spec version can be `0.1.0` everywhere

pkg/repository/manifest.go
```diff
@@ -280,6 +424,20 @@ func readTimestampManifest(input io.Reader, keys *KeyStore) (*Timestamp, error)
 	return &ts, nil
 }
 
+// BatchSaveManifests write a series of manifests to disk
+func BatchSaveManifests(dst string, manifestList []ValidManifest) error {
+	for _, m := range manifestList {
+		writer, err := os.OpenFile(filepath.Join(dst, m.Filename()), os.O_WRONLY|os.O_CREATE, 0644)
```
lonng, line 430:
```suggestion
		writer, err := os.OpenFile(filepath.Join(dst, m.Filename()), os.O_WRONLY|os.O_TRUNC|os.O_CREATE, 0644)
```

pkg/repository/manifest.go
```diff
@@ -243,6 +340,53 @@ func (manifest *Timestamp) Filename() string {
 	return "timestamp.json"
 }
 
+// SetRoles populates role list in the root manifest
+func (manifest *Root) SetRoles(manifestList []ValidManifest) {
+	if manifest.Roles == nil {
+		manifest.Roles = make(map[string]*Role)
+	}
+	for _, m := range manifestList {
+		manifest.Roles[m.Filename()] = &Role{
```
nrc, line 349: I think it is better to iterate over `types` rather than passing in a manifest list (skip the component type because `filename == ""`). This is because in TUF, roles are not exactly the same as manifests. Alos, once there are component manifests around, there won't be a manifest list.

pkg/repository/manifest.go
```diff
@@ -280,6 +424,20 @@ func readTimestampManifest(input io.Reader, keys *KeyStore) (*Timestamp, error)
...
 }
 
+// BatchSaveManifests write a series of manifests to disk
+func BatchSaveManifests(dst string, manifestList []ValidManifest) error {
+	for _, m := range manifestList {
+		writer, err := os.OpenFile(filepath.Join(dst, m.Filename()), os.O_WRONLY|os.O_CREATE, 0644)
+		if err != nil {
+			return err
+		}
+		if err = signAndWrite(writer, m); err != nil {
```
lonng, line 434: I have never seen there is a snippet to close the writer.

nrc, line 434: Should `defer writer.Close()` before this `if`

pkg/repository/manifest.go
```diff
@@ -91,12 +164,36 @@ type Snapshot struct {
 	Meta map[string]FileVersion `json:"meta"`
 }
 
+// NewSnapshot creates a Snapshot object.
+func NewSnapshot(initTime time.Time) *Snapshot {
+	return &Snapshot{
+		SignedBase: SignedBase{
+			Ty:          "snapshot",
```
lonng, line 171: Can we define some constants for those literals?

nrc: Yes, my fault - I've been meaning to do that

pkg/repository/manifest.go
nrc: I'm no expert on this, I would think that all files need closing. In this case there is a `close` in the example of https://golang.org/pkg/os/#OpenFile, but it does not use defer.

PR #213, opened 2020-05-14
Download and save timestamp and snapshot manifests

nrc:
This required a bunch of refactoring, mostly to the new manifest stuff, also to add a size limit when we fetch resources (and I fixed a TODO at the same time). I've added mock local manifest stores and mirrors for testing.

PTAL @AstroProfundis


pkg/repository/local_manifests.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+// MockManifests is a LocalManifests implementation for testing.
+type MockManifests struct {
+	Manifests map[string]ValidManifest
+}
+
+// SaveManifest implements LocalManifests.
+func (ms *MockManifests) SaveManifest(manifest *Manifest) error {
+	return nil
```
lonng, line 105: Seems we should save the `manifest` at `ms. Manifests`.

PR #215, opened 2020-05-15
Add RSA signature

lucklove:

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/pkg/repository/crypto/keys.go
+++ b/pkg/repository/crypto/keys.go
@@ -1,0 +1,75 @@
+// Copyright 2020 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package crypto
+
+import (
+	"errors"
+)
+
+var (
+	// ErrorKeyUninitialized will be present when key is used before Deserialize called
+	ErrorKeyUninitialized = errors.New("Key not initialized, call Deserialize first")
+	// ErrorDeserializeKey means the key format is not valid
+	ErrorDeserializeKey = errors.New("Error on deserialize key, check if the key is valid")
+)
+
+// KeyStore is the collection of all public keys available to TiUp.
+type KeyStore interface {
+	// Add put new key into KeyStore
+	Put(string, PubKey) KeyStore
+
+	// Get return a key by it's id
+	Get(string) PubKey
+}
+
+// Serializable represents object that can be serialized and deserialized
+type Serializable interface {
+	// Translate the key to the format that can be stored
+	Serialize() ([]byte, error)
+
+	// Deserialize a key from data
+	Deserialize([]byte) error
+}
+
+// PubKey is a public key available to TiUp
+type PubKey interface {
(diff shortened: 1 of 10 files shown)


pkg/repository/crypto/keys.go
```diff
@@ -0,0 +1,75 @@
...
+	// ErrorKeyUninitialized will be present when key is used before Deserialize called
+	ErrorKeyUninitialized = errors.New("Key not initialized, call Deserialize first")
+	// ErrorDeserializeKey means the key format is not valid
+	ErrorDeserializeKey = errors.New("Error on deserialize key, check if the key is valid")
+)
+
+// KeyStore is the collection of all public keys available to TiUp.
+type KeyStore interface {
+	// Add put new key into KeyStore
+	Add(string, PubKey) KeyStore
```
lonng, line 30: Prefer `Put`

pkg/repository/crypto/keys.go
```diff
@@ -0,0 +1,75 @@
...
+
+// PubKey is a public key available to TiUp
+type PubKey interface {
+	Serializable
+	// Verify check the signature is right
+	Verify(payload []byte, sig string) error
+}
+
+// PriKey is the private key that provide signature method
+type PriKey interface {
```
AstroProfundis, line 53: I think `PrivKey` might be better.

pkg/repository/crypto/rsa.go
```diff
@@ -0,0 +1,129 @@
...
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"encoding/asn1"
+	"encoding/base64"
+	"encoding/pem"
+)
+
+// RsaPair generate a pair of rsa keys
+func RsaPair() (*RsaPubKey, *RsaPriKey, error) {
```
AstroProfundis, line 27: Should be `RSA` with all tree captive letters

pkg/repository/crypto/rsa.go
```diff
@@ -0,0 +1,129 @@
...
+	key, err := rsa.GenerateKey(rand.Reader, 2048)
```
AstroProfundis, line 28: Put the key length to a const at top? We may need to change the value (or even support different values) in the future

pkg/repository/crypto/keys.go
```diff
@@ -0,0 +1,75 @@
...
+type Serializable interface {
+	// Translate the key to the format that can be stored
+	Serialize() ([]byte, error)
+
+	// Deserialize a key from data
+	Deserialize([]byte) error
+}
+
+// PubKey is a public key available to TiUp
+type PubKey interface {
```
AstroProfundis, line 46: Should also add a method to get fingerprint of the key

pkg/sign/sign.go
```diff
@@ -0,0 +1,27 @@
+package sign
```
lonng, line 1: Please add the license header.

pkg/repository/crypto/keys.go
lucklove: Why need this now?

pkg/sign/sign.go
lucklove: This file is of no use, deleted already

pkg/repository/crypto/keys.go
lucklove: The key-id has the same function as this, but it should be added into KeyInfo, I think

AstroProfundis: I believe the value of key id is fingerprint of that key, I didn't see that part, maybe I missed something?

PR #217, opened 2020-05-15
Some followups to #216

nrc:
The first commit separate manifest package into two to keep code for old (v0) and new (v1) manifests distinct. The second commit removes `description` from `ComponentItem`. The description is kept only in the component manifest, not the index so that the field is not duplicated and so that an owner can update a component including its description without changing the index.

PTAL @lonng


cmd/run.go
```diff
@@ -17,6 +17,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/pingcap-incubator/tiup/pkg/repository/v0manifest"
```
lonng, line 20: Please optimize all imports.

tools/migrate/main.go
```diff
@@ -254,22 +255,21 @@ func migrate(srcDir, dstDir string) error {
 			return errors.Trace(err)
 		}
 
-		index.Components[comp.Name] = manifest.ComponentItem{
-			Name:        comp.Name,
-			Description: comp.Desc,
```
lonng, line 259: The description is used at `tiup list`, we must fetch all component manifest to display the description if we remove it from `ComponentItem`.

nrc: I think it is ok - we will have all the manifests locally and we can cache the descriptions locally too.

cmd/run.go
nrc: is there an argument to goimports which will make it do this?

lonng: I think `make fmt` will optimize the imports.

nrc: It only organises them within groups :-(

nrc:
@lonng please take another look

PR #218, opened 2020-05-15
Add basic implementation of component

AstroProfundis:
Add basic implementation of `Component` manifest, and add a function to generate new component.

Also implemented signing of manifests during init.

An `AddComponent()` is added but not used yet, commands will be added in a following PR.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/pkg/repository/crypto/rsa.go
+++ b/pkg/repository/crypto/rsa.go
@@ -15,11 +15,13 @@ package crypto
 
 import (
 	"crypto"
+	"crypto/md5"
 	"crypto/rand"
 	"crypto/rsa"
 	"crypto/x509"
 	"encoding/asn1"
 	"encoding/base64"
+	"encoding/hex"
 	"encoding/pem"
 )
 
@@ -83,6 +85,16 @@ func (k *RSAPubKey) Verify(payload []byte, sig string) error {
 	}
 
 	return rsa.VerifyPSS(k.key, crypto.SHA256, hashed, b64decSig, nil)
+}
+
+// Fingerprint returns the MD5 hash of the public key
+func (k *RSAPubKey) Fingerprint() (string, error) {
+	bytes, err := x509.MarshalPKIXPublicKey(k.key)
+	if err != nil {
+		return "", err
+	}
+	hash := md5.Sum(bytes)
+	return hex.EncodeToString(hash[:]), nil
 }
 
 // RSAPrivKey represents the private key of RSA
--- a/pkg/repository/crypto/rsa.go
+++ b/pkg/repository/crypto/rsa.go
@@ -15,21 +15,19 @@ package crypto
 
 import (
 	"crypto"
-	"crypto/md5"
 	"crypto/rand"
 	"crypto/rsa"
 	"crypto/x509"
 	"encoding/asn1"
 	"encoding/base64"
-	"encoding/hex"
 	"encoding/pem"
 )
 
 // RSAKeyLength define the length of RSA keys
 const RSAKeyLength = 2048
 
-// RsaPair generate a pair of rsa keys
-func RsaPair() (*RSAPubKey, *RSAPrivKey, error) {
+// RSAPair generate a pair of rsa keys
+func RSAPair() (*RSAPubKey, *RSAPrivKey, error) {
 	key, err := rsa.GenerateKey(rand.Reader, RSAKeyLength)
 	if err != nil {
 		return nil, nil, err
@@ -87,16 +85,6 @@ func (k *RSAPubKey) Verify(payload []byte, sig string) error {
 	return rsa.VerifyPSS(k.key, crypto.SHA256, hashed, b64decSig, nil)
 }
 
-// Fingerprint returns the MD5 hash of the public key
-func (k *RSAPubKey) Fingerprint() (string, error) {
-	bytes, err := x509.MarshalPKIXPublicKey(k.key)
-	if err != nil {
-		return "", err
-	}
-	hash := md5.Sum(bytes)
-	return hex.EncodeToString(hash[:]), nil
-}
-
 // RSAPrivKey represents the private key of RSA
 type RSAPrivKey struct {
 	key *rsa.PrivateKey
@@ -141,4 +129,11 @@ func (k *RSAPrivKey) Signature(payload []byte) (string, error) {
 		return "", err
 	}
 	return base64.StdEncoding.EncodeToString(sig), nil
+}
+
+// Public returns public key of the PrivKey
+func (k *RSAPrivKey) Public() PubKey {
+	return &RSAPubKey{
+		key: &k.key.PublicKey,
+	}
 }
--- a/pkg/repository/v1manifest/repo.go
+++ b/pkg/repository/v1manifest/repo.go
@@ -53,7 +53,10 @@ func Init(dst string, initTime time.Time, priv string) error {
 
 	// snapshot and timestamp are the last two manifests to be initialized
 	// init snapshot
-	manifests[ManifestTypeSnapshot] = NewSnapshot(initTime).SetVersions(manifests)
+	manifests[ManifestTypeSnapshot], err = NewSnapshot(initTime).SetVersions(manifests)
+	if err != nil {
+		return err
+	}
 
 	// init timestamp
 	timestamp, err := NewTimestamp(initTime).SetSnapshot(manifests[ManifestTypeSnapshot].(*Snapshot))
@@ -129,7 +132,10 @@ func AddComponent(id, name, desc, owner, repo string, isDefault bool, pub, priv 
 	index.Version += 1 // bump index version
 
 	// update snapshot
-	snapshot := manifests[ManifestTypeSnapshot].(*Snapshot).SetVersions(manifests)
+	snapshot, err := manifests[ManifestTypeSnapshot].(*Snapshot).SetVersions(manifests)
+	if err != nil {
+		return err
+	}
 	snapshot.Expires = currTime.Add(ManifestsConfig[ManifestTypeSnapshot].Expire).Format(time.RFC3339)
 
 	// update timestamp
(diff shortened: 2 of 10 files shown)


AstroProfundis:
For local testing, keypair generated with the following commands are confirmed to work:
```
openssl genrsa -out priv 2048
openssl rsa -in priv -pubout -out pub
```
And then `./bin/tiup repo init path/to/repo --pubkey pub --privkey priv`

pkg/repository/crypto/rsa.go
```diff
@@ -85,6 +87,16 @@ func (k *RSAPubKey) Verify(payload []byte, sig string) error {
 	return rsa.VerifyPSS(k.key, crypto.SHA256, hashed, b64decSig, nil)
 }
 
+// Fingerprint returns the MD5 hash of the public key
+func (k *RSAPubKey) Fingerprint() (string, error) {
+	bytes, err := x509.MarshalPKIXPublicKey(k.key)
+	if err != nil {
+		return "", err
+	}
+	hash := md5.Sum(bytes)
```
lonng, line 96: It should be sha256 of bytes according to the design document.

pkg/repository/manifest/repo.go
```diff
@@ -16,16 +16,32 @@ package manifest
...
+	// read key files
+	privBytes, err := ioutil.ReadFile(priv)
+	if err != nil {
+		return err
+	}
+	privKey := &crypto.RSAPrivKey{}
+	if err = privKey.Deserialize(privBytes); err != nil {
+		return err
+	}
+	pubBytes, err := ioutil.ReadFile(pub)
```
lonng, line 36: Why do we need to store the public key? we can obtain the public key from the private key directly.

pkg/repository/manifest/repo.go
```diff
@@ -57,7 +78,75 @@ func Init(dst string, initTime time.Time) error {
 		return fmt.Errorf("manifest '%s' not initialized porperly", ty)
 	}
 
-	return BatchSaveManifests(dst, manifests)
+	return BatchSaveManifests(dst, manifests, pubKey, privKey)
```
lonng, line 81: Seems we don't need `pubKey` anymore, because we can obtain it via `privKey`.

pkg/repository/manifest/repo.go
```diff
@@ -57,7 +78,75 @@ func Init(dst string, initTime time.Time) error {
...
+	// read key files
+	privBytes, err := ioutil.ReadFile(priv)
+	if err != nil {
+		return err
+	}
+	privKey := &crypto.RSAPrivKey{}
+	if err = privKey.Deserialize(privBytes); err != nil {
+		return err
+	}
+	pubBytes, err := ioutil.ReadFile(pub)
```
lonng, line 95: ditto

pkg/repository/manifest/manifest.go
```diff
@@ -222,16 +248,45 @@ func readTimestampManifest(input io.Reader, keys crypto.KeyStore) (*Timestamp, e
 	return &ts, nil
 }
 
+// SignAndWrite creates a manifest and writes it to out.
+func SignAndWrite(out io.Writer, role ValidManifest, keyID string, privKey crypto.PrivKey) error {
```
lonng, line 252: Seems we can remove the `keyID` parameter if we can obtain the fingerprint from public key.

pkg/repository/manifest/repo.go
```diff
@@ -57,7 +78,75 @@ func Init(dst string, initTime time.Time) error {
...
+	}
+
+	// read manifest files from disk
+	manifests, err := ReadManifestDir(repo)
+	if err != nil {
+		return nil
+	}
+
+	// check id conflicts
+	if _, found := manifests[ManifestTypeIndex].(*Index).Components[id]; found {
```
lonng, line 111: Name conflicts with `index/timestamp/snapshot/root` should be checked also.

pkg/repository/crypto/rsa.go
```diff
@@ -85,6 +87,16 @@ func (k *RSAPubKey) Verify(payload []byte, sig string) error {
 	return rsa.VerifyPSS(k.key, crypto.SHA256, hashed, b64decSig, nil)
 }
 
+// Fingerprint returns the MD5 hash of the public key
+func (k *RSAPubKey) Fingerprint() (string, error) {
```
nrc, line 91: The key id is the hash of the whole key info, not just the public key

pkg/repository/manifest/repo.go
```diff
@@ -57,7 +78,75 @@ func Init(dst string, initTime time.Time) error {
...
+	bytes, err := json.Marshal(comp)
+	if err != nil {
+		return err
+	}
+	compInfo := ComponentItem{
+		Name:      comp.Name,
+		Owner:     owner,
+		URL:       fmt.Sprintf("/%s", comp.Filename()),
+		Length:    uint(len(bytes)),
+		Threshold: 1, // TODO: make this configurable
```
nrc, line 129: It shouldn't be configurable

pkg/repository/manifest/repo.go
```diff
@@ -16,16 +16,32 @@ package manifest
...
 	"github.com/pingcap-incubator/tiup/pkg/repository/crypto"
-	"github.com/pingcap/errors"
 )
 
-// Init creates and initializes an empty reposityro
-func Init(dst string, initTime time.Time) error {
+// Init creates and initializes an empty repository
+func Init(dst string, initTime time.Time, pub, priv string) error {
+	// read key files
+	privBytes, err := ioutil.ReadFile(priv)
```
nrc, line 28: we should not take keys from the user, instead we generate them ourselves and print the private key

pkg/repository/manifest/types.go
```diff
@@ -90,15 +79,16 @@ type Owner struct {
...
 type VersionItem struct {
-	Yanked       bool              `json:"yanked"`
-	URL          string            `json:"url"`
-	Hashes       map[string]string `json:"hashes"`
-	Entry        string            `json:"entry"`
-	Released     string            `json:"released"`
-	Length       int64             `json:"length"`
-	Dependencies []string          `json:"dependencies"`
+	URL          string   `json:"url"`
+	Length       uint     `json:"length"`
```
nrc, line 85: I think Length is in FileHash

tools/migrate/main.go
```diff
@@ -255,13 +257,12 @@ func migrate(srcDir, dstDir string) error {
...
-			Owner:       "pingcap",
-			URL:         fmt.Sprintf("/%s", name),
-			Length:      stat.Size(),
-			Threshold:   0,
+			Name:      comp.Name,
+			Yanked:    false,
+			Owner:     "pingcap",
+			URL:       fmt.Sprintf("/%s", name),
+			Length:    uint(stat.Size()),
+			Threshold: 0,
```
nrc, line 265: Threshold should be 1

pkg/repository/manifest/repo.go
```diff
@@ -57,7 +78,75 @@ func Init(dst string, initTime time.Time) error {
...
+	if err != nil {
+		return err
+	}
+	pubKey := &crypto.RSAPubKey{}
+	if err = pubKey.Deserialize(pubBytes); err != nil {
+		return err
+	}
+
+	// read manifest files from disk
+	manifests, err := ReadManifestDir(repo)
```
nrc, line 105: I think you should only read the index manifest here. Note that we don't save index.json, only n.index.json and you'll need to use the correct version from snapshot.json

pkg/repository/manifest/repo.go
```diff
@@ -57,7 +78,75 @@ func Init(dst string, initTime time.Time) error {
...
+	}
+	index := manifests[ManifestTypeIndex].(*Index)
+	index.Components[id] = compInfo
+	if isDefault {
+		index.DefaultComponents = append(index.DefaultComponents, id)
+	}
+	index.Version += 1 // bump index version
+
+	// update snapshot
+	snapshot := manifests[ManifestTypeSnapshot].(*Snapshot).SetVersions(manifests)
```
nrc, line 139: I think it is better to explicitly update the version for the index and component entries, not all manifests

pkg/repository/manifest/repo.go
```diff
@@ -57,7 +78,75 @@ func Init(dst string, initTime time.Time) error {
...
+	snapshot.Expires = currTime.Add(ManifestsConfig[ManifestTypeSnapshot].Expire).Format(time.RFC3339)
+
+	// update timestamp
+	timestamp, err := manifests[ManifestTypeTimestamp].(*Timestamp).SetSnapshot(snapshot)
+	if err != nil {
+		return err
+	}
+	timestamp.Expires = currTime.Add(ManifestsConfig[ManifestTypeTimestamp].Expire).Format(time.RFC3339)
```
nrc, line 147: Should create a new timestamp manifest rather than updating the old one. Also note that the version number needs to incremented

pkg/repository/manifest/manifest.go
```diff
@@ -212,6 +216,28 @@ func ReadManifest(input io.Reader, role ValidManifest, keys crypto.KeyStore) (*M
 	return &m, m.Signed.isValid()
 }
 
+// ReadManifestDir reads manifests from a dir
```
nrc, line 219: Could you move this to repo.go please?

pkg/repository/manifest/repo.go
AstroProfundis: This is intended for easier local developing and testing, key generating and storing will be added in another PR that implement `genkey` subcommand.

AstroProfundis:
I'll add test cases along with implementing `repo component` sub command in a following PR, as the functions in this PR is not fully working yet and may need to adjust then.

pkg/repository/v1manifest/repo.go
```diff
@@ -113,43 +202,84 @@ func NewTimestamp(initTime time.Time) *Timestamp {
...
+}
+
+// SetSnapshot hashes a snapshot manifest and update the timestamp manifest
+func (manifest *Timestamp) SetSnapshot(s *Snapshot) (*Timestamp, error) {
+	bytes, err := cjson.Marshal(s)
+	if err != nil {
+		return manifest, err
+	}
+
+	hash := sha256.Sum256(bytes)
```
lonng, line 272: Both `sha256/sha512` hashes should be provided.

pkg/repository/v1manifest/repo.go
```diff
@@ -113,43 +202,84 @@ func NewTimestamp(initTime time.Time) *Timestamp {
...
+
+	if manifest.Meta == nil {
+		manifest.Meta = make(map[string]FileHash)
+	}
+	manifest.Meta[s.Base().Filename()] = FileHash{
+		Hashes: map[string]string{"sha256": hex.EncodeToString(hash[:])},
+		Length: uint(len(bytes)),
```
lonng, line 279: I don't think the length is correct. file length? or just signed length?

pkg/repository/v1manifest/repo.go
```diff
@@ -58,7 +82,72 @@ func Init(dst string, initTime time.Time) error {
 		return fmt.Errorf("manifest '%s' not initialized porperly", ty)
 	}
 
-	return BatchSaveManifests(dst, manifests)
+	return BatchSaveManifests(dst, manifests, privKey)
+}
+
+// AddComponent adds a new component to an existing repository
+func AddComponent(id, name, desc, owner, repo string, isDefault bool, pub, priv string) error {
+	id = strings.ToLower(id)
```
nrc, line 90: Don't do this, the id should be taken literally and can include case differences

pkg/repository/v1manifest/repo.go
```diff
@@ -113,43 +202,84 @@ func NewTimestamp(initTime time.Time) *Timestamp {
...
 	}
 	for _, m := range manifestList {
+		bytes, err := cjson.Marshal(m)
+		if err != nil {
+			return nil, err
+		}
 		manifest.Meta[m.Filename()] = FileVersion{
 			Version: m.Base().Version,
-			// TODO length
+			Length:  uint(len(bytes)),
```
nrc, line 259: the length should be the length of the whole manifest, not just the signed field

pkg/repository/v1manifest/repo.go
```diff
@@ -113,43 +202,84 @@ func NewTimestamp(initTime time.Time) *Timestamp {
...
-func (manifest *Snapshot) SetVersions(manifestList map[string]ValidManifest) *Snapshot {
+func (manifest *Snapshot) SetVersions(manifestList map[string]ValidManifest) (*Snapshot, error) {
 	if manifest.Meta == nil {
 		manifest.Meta = make(map[string]FileVersion)
 	}
 	for _, m := range manifestList {
+		bytes, err := cjson.Marshal(m)
+		if err != nil {
+			return nil, err
+		}
```
nrc, line 256: Don't need to propagate the error - there is no way to recover from it and it should never happen.

pkg/repository/v1manifest/repo.go
nrc: Yeah, it should be the whole file

PR #220, opened 2020-05-18
manifest: refine key generating and manifest signing

lucklove:
### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes


cmd/repo.go
```diff
@@ -155,13 +157,28 @@ func newRepoGenkeyCmd(env *meta.Environment) *cobra.Command {
 		Short: "Generate a new key pair",
 		Long:  `Generate a new key pair that can be used to sign components.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
+			privPath := env.Profile().Path("private.pem")
+			pubPath := env.Profile().Path("public.pem")
```
lonng, line 161: I don't think we should store the public to the disk, because we can obtain the public key from private key directly.

cmd/repo.go
```diff
@@ -206,20 +233,32 @@ current working directory (".") will be used.`,
...
 			}
 
-			return initRepo(repoPath)
+			return initRepo(repoPath, keyDir)
 		},
 	}
 
-	cmd.Flags().StringVar(&pubKey, "pubkey", "", "Path to the public key file")
-	cmd.Flags().StringVar(&privKey, "privkey", "", "Path to the private key file")
+	cmd.Flags().StringVar(&keyDir, "key-dir", "", "Path to write the private key file")
```
lonng, line 240: I think the `-i` is enough if we don't need to specify the public key.

pkg/repository/crypto/keys.go
```diff
@@ -21,11 +21,26 @@ var (
 	// ErrorKeyUninitialized will be present when key is used before Deserialize called
 	ErrorKeyUninitialized = errors.New("key not initialized, call Deserialize first")
 	// ErrorDeserializeKey means the key format is not valid
-	ErrorDeserializeKey = errors.New("error on deserialize key, check if the key is valid")
+	ErrorDeserializeKey = errors.New("Error on deserialize key, check if the key is valid")
```
lonng, line 24: The error message should be started with the lower case (which is more Golang idiom).

pkg/repository/crypto/keys.go
```diff
@@ -73,3 +98,59 @@ func (s *keychain) Put(id string, key PubKey) KeyStore {
 	(*s)[id] = key
 	return s
 }
+
+// Keys returns all key ids
+func (s *keychain) Keys() []string {
```
lonng, line 103: Prefer to `KeyIDs`

pkg/repository/crypto/keys.go
```diff
@@ -21,11 +21,26 @@ var (
...
+	ErrorDeserializeKey = errors.New("Error on deserialize key, check if the key is valid")
+	// ErrorUnsupportedKeyType means we don't supported this type of key
+	ErrorUnsupportedKeyType = errors.New("Provided key type not supported")
+	// ErrorUnsupportedKeySchema means we don't support this schema
+	ErrorUnsupportedKeySchema = errors.New("Provided schema not supported")
+)
+
+const (
+	// KeyTypeRSA represents the RSA type of keys
+	KeyTypeRSA = "RSA"
```
lonng, line 33: Prefer to lower case and keep consistent with JSON content.

pkg/repository/crypto/keys.go
```diff
@@ -73,3 +98,59 @@ func (s *keychain) Put(id string, key PubKey) KeyStore {
...
+	// We only support rsassa-pss-sha256 now
+	if keyScheme != KeySchemeRSASSAPSSSHA256 {
+		return nil, nil, ErrorUnsupportedKeySchema
+	}
+
+	return RSAPair()
+}
+
+// NewPrivKey return PrivKey
+func NewPrivKey(keyType, keyScheme string, payload []byte) (PrivKey, error) {
```
lonng, line 127:
```suggestion
func NewPrivKey(keyType, keyScheme string, key []byte) (PrivKey, error) {
```

pkg/repository/crypto/keys.go
```diff
@@ -73,3 +98,59 @@ func (s *keychain) Put(id string, key PubKey) KeyStore {
...
+	if keyScheme != KeySchemeRSASSAPSSSHA256 {
+		return nil, ErrorUnsupportedKeySchema
+	}
+
+	priv := &RSAPrivKey{}
+	return priv, priv.Deserialize(payload)
+}
+
+// NewPubKey return PrivKey
+func NewPubKey(keyType, keyScheme string, payload []byte) (PubKey, error) {
```
lonng, line 143:
```suggestion
func NewPubKey(keyType, keyScheme string, key []byte) (PubKey, error) {
```

cmd/repo.go
lucklove: Yes

cmd/repo.go
```diff
@@ -155,13 +157,28 @@ func newRepoGenkeyCmd(env *meta.Environment) *cobra.Command {
 		Short: "Generate a new key pair",
 		Long:  `Generate a new key pair that can be used to sign components.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
+			privPath := env.Profile().Path("private.pem")
+			pubPath := env.Profile().Path("public.pem")
+			if utils.IsExist(privPath) && utils.IsExist(pubPath) {
```
AstroProfundis: We could just remove the public key reading logic

pkg/repository/v1manifest/keys.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	payload, err := cjson.Marshal(KeyInfo{
+		Algorithms: ki.Algorithms,
+		Type:       ki.Type,
+		Scheme:     ki.Scheme,
+		Value:      value,
+	})
+	if err != nil {
+		// XXX: maybe we can assume that the error should always be nil since the KeyInfo struct is valid
```
AstroProfundis, line 78: yes this error can be ignored, it should never be an error

pkg/repository/v1manifest/keys.go
```diff
@@ -0,0 +1,147 @@
...
+		priv, err := ki.privateKey()
+		if err != nil {
+			return nil, err
+		}
+		return priv.Public(), nil
+	}
+	return crypto.NewPubKey(ki.Type, ki.Scheme, []byte(ki.Value["public"]))
+}
+
+// privateKey return PrivKey if the key is private key
```
lonng, line 140:
```suggestion
// privateKey returns PrivKey if the key is private key
```

pkg/repository/v1manifest/repo.go
```diff
@@ -318,7 +290,7 @@ func (manifest *Timestamp) SetSnapshot(s *Manifest) (*Timestamp, error) {
 }
 
 // SetRole populates role list in the root manifest
-func (manifest *Root) SetRole(m ValidManifest) {
+func (manifest *Root) SetRole(m ValidManifest) error {
```
AstroProfundis, line 293: What did the `error` represent? It seems this function doesn't produce any error at all.

PR #222, opened 2020-05-19
Implement fetching and checking of components using v1 manifests

nrc:
Also does some minor refactoring.

PTAL @lucklove

### DIFF
--- a/pkg/repository/v1_repository.go
+++ b/pkg/repository/v1_repository.go
@@ -14,6 +14,7 @@
 package repository
 
 import (
+	"bytes"
 	"fmt"
 	"io"
 	"runtime"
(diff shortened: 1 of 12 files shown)


pkg/repository/v1_repository.go
```diff
@@ -14,7 +14,10 @@
 package repository
 
 import (
+	"bytes"
 	"fmt"
+	"github.com/pingcap-incubator/tiup/pkg/utils"
```
lonng, line 19: Please fix the imports.

pkg/repository/v1_repository.go
```diff
@@ -174,42 +179,79 @@ func (r *V1Repository) updateComponentManifest(local v1manifest.LocalManifests,
...
+// downloadComponent downloads a component using the relevant manifest and checks its hash.
+// Precondition: the component's manifest is up to date and the version and platform are valid.
+// The caller is responsible for closing the returned reader.
+func (r *V1Repository) downloadComponent(id string, platform string, version string) (io.Reader, error) {
+	manifest, err := r.local.LoadComponentManifest(v1manifest.ComponentFilename(id))
+	if err != nil {
+		return nil, err
+	}
+
+	item := manifest.Platforms[platform][version]
```
lonng, line 199: Do we need to check whether `platform/version` exists?

pkg/repository/v1_repository.go
nrc: Gah, sorry, my IDE just sticks them in anywhere and then hides the imports so I don't notice :-(

pkg/repository/v1_repository.go
nrc: its up to the caller to ensure. I think the caller should know from their flow, so it would be a waste to check here (see also the precondition in the comment)

nrc:
Comments addressed, PTAL @lonng

PR #223, opened 2020-05-19
Generate local keys

lucklove:

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/tools/migrate/main.go
+++ b/tools/migrate/main.go
@@ -137,6 +137,21 @@ func migrate(srcDir, dstDir string) error {
 		index    = v1manifest.NewIndex(initTime)
 	)
 
+	// TODO: bootstrap a server instead of generating key
+	keys := map[string][]*v1manifest.KeyInfo{}
+	tys := []string{
+		v1manifest.ManifestTypeRoot,
+		v1manifest.ManifestTypeIndex,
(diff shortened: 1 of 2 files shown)


tools/migrate/main.go
```diff
@@ -137,6 +137,21 @@ func migrate(srcDir, dstDir string) error {
...
+	keys := map[string][]*v1manifest.KeyInfo{}
+	tys := []string{
+		v1manifest.ManifestTypeRoot,
+		v1manifest.ManifestTypeIndex,
+		v1manifest.ManifestTypeSnapshot,
+		v1manifest.ManifestTypeTimestamp,
+	}
+	keyDir := filepath.Join(dstDir, "keys")
+	for _, ty := range tys {
+		if err := v1manifest.GenAndSaveKeys(keys, ty, int(v1manifest.ManifestsConfig[ty].Threshold), keyDir); err != nil {
```
AstroProfundis, line 150: (as a note) We should avoid generating new keys during migrating manifest but using the real root keys instead. Since we don't have any "real" root keys, this process is ok, but we need to replace this testing repo entirely at GA.

PR #224, opened 2020-05-19
implement update root

july2993:
implement update root

Seems we should VerifySignature according to the threshold of the manifest type but not just all signatures?
Also we should update the keys of `LocalRoot` when we get or save a new root manifest

### DIFF
--- a/pkg/repository/v1_repository.go
+++ b/pkg/repository/v1_repository.go
@@ -14,8 +14,11 @@
 package repository
 
 import (
+	"encoding/json"
 	"fmt"
+	"path/filepath"
 	"runtime"
+	"strconv"
 
 	"github.com/pingcap-incubator/tiup/pkg/repository/crypto"
 	"github.com/pingcap-incubator/tiup/pkg/repository/v1manifest"
@@ -44,6 +47,7 @@ func NewV1Repo(mirror Mirror, opts Options) *V1Repository {
 }
 
 const maxTimeStampSize uint = 1024
+const maxRootSize uint = 1024 * 1024
 
 // If the snapshot has been updated, we return the new snapshot, if not we return nil.
 // Postcondition: if returned error is nil, then the local snapshot and timestamp are up to date.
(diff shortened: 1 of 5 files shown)


pkg/repository/v1_repository.go
```diff
@@ -88,13 +92,100 @@ func (r *V1Repository) updateLocalSnapshot() (*v1manifest.Snapshot, error) {
...
+	}
+
+	var newRoots []v1manifest.Manifest
+
+	for version := root1.Version + 1; ; version++ {
+		var root2 v1manifest.Root
+		fname := fnameWithVersion(v1manifest.ManifestURLRoot, version)
+
+		reader, err := r.mirror.Fetch(fname, int64(maxRootSize))
+		if err != nil {
```
lonng, line 117: I think we break just if the error is `NotFound`, and other errors should propagate to the caller.

pkg/repository/v1_repository.go
```diff
@@ -88,13 +92,100 @@ func (r *V1Repository) updateLocalSnapshot() (*v1manifest.Snapshot, error) {
...
+		return err
+	}
+
+	var newRoots []v1manifest.Manifest
+
+	for version := root1.Version + 1; ; version++ {
+		var root2 v1manifest.Root
+		fname := fnameWithVersion(v1manifest.ManifestURLRoot, version)
+
+		reader, err := r.mirror.Fetch(fname, int64(maxRootSize))
```
nrc, line 116: Could `maxRootSize` be an `int64` and skip this conversion

pkg/repository/v1_repository.go
```diff
@@ -88,13 +92,100 @@ func (r *V1Repository) updateLocalSnapshot() (*v1manifest.Snapshot, error) {
...
+			break
+		}
+
+		decoder := json.NewDecoder(reader)
+		var m v1manifest.Manifest
+		m.Signed = &root2
+		err = decoder.Decode(&m)
+		if err != nil {
+			return err
+		}
```
nrc, line 129: I think decoding and basic verification should be done in manifest.go.

pkg/repository/v1manifest/manifest.go
```diff
@@ -95,31 +95,50 @@ var ManifestsConfig = map[string]ty{
...
-// verifySignature ensures that each signature in manifest::signatures is a valid signature of manifest::signed.
-func (manifest *Manifest) verifySignature(keys crypto.KeyStore) error {
+// VerifySignature ensures that threshold signature in manifest::signatures is a valid signature of manifest::signed.
+func (manifest *Manifest) VerifySignature(threshold uint, keys crypto.KeyStore) error {
 	if keys == nil {
 		return nil
 	}
 
+	if threshold == 0 {
+		return nil
```
nrc, line 105: This should be an error

pkg/repository/v1manifest/manifest.go
```diff
@@ -95,31 +95,50 @@ var ManifestsConfig = map[string]ty{
...
+		// TODO use SignatureError
+		if err != nil {
+			errs = append(errs, err)
+		} else {
+			validCount++
 		}
 	}
 
-	return nil
+	if validCount >= threshold {
```
nrc, line 137: I think all signatures must be correct, not just the threshold number. Threshold just dictates the number of signatures required.

PR #225, opened 2020-05-19
repository: fix tarball downloading

AstroProfundis:
When downloading tarball, it fails to decompress

There was a `FIXME` comment that suggests to store downloaded files in memory other than files to reduce disk IO, but after fixing that, the behavior of handling tarballs are not correct.

In `pkg/repository/file_source.go`, we check file hash before decompress it, so the returned reader was read twice if `expand` is true, the 2nd read can only get `EOF`. There is a `Seek()` call to reset the offset before decompressing, however, this only works for `os.File`, but not other reader types.

This PR partially revert #213 and changed the process back to storing downloaded files on disk in a temp dir. (This also avoids the risk that downloading large file may lead to too much memory usage)

### Check List 

Tests 

 - Unit test

Code changes

 - Has persistent data change

Side effects

Related changes

### DIFF
--- a/pkg/repository/mirror.go
+++ b/pkg/repository/mirror.go
@@ -17,13 +17,16 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"math/rand"
 	"net/http"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"time"
 
 	"github.com/cavaliercoder/grab"
+	"github.com/pingcap-incubator/tiup/pkg/utils"
 	"github.com/pingcap/errors"
 )
 
@@ -138,11 +141,17 @@ func (l *localFilesystem) Close() error {
 
 type httpMirror struct {
 	server  string
+	tmpDir  string
 	options MirrorOptions
 }
 
 // Open implements the Mirror interface
 func (l *httpMirror) Open() error {
+	tmpDir := filepath.Join(os.TempDir(), strconv.Itoa(rand.Int()))
+	if err := os.MkdirAll(tmpDir, os.ModePerm); err != nil {
+		return errors.Trace(err)
+	}
+	l.tmpDir = tmpDir
 	return nil
 }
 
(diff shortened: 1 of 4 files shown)


pkg/repository/mirror.go
```diff
@@ -205,11 +213,24 @@ func (l *httpMirror) Download(resource, targetDir string) error {
 
 // Fetch implements the Mirror interface
 func (l *httpMirror) Fetch(resource string, maxSize int64) (io.ReadCloser, error) {
-	return l.download(l.prepareURL(resource), "", maxSize)
+	r, err := l.download(l.prepareURL(resource), filepath.Join(l.tmpDir, resource), maxSize)
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	defer r.Close()
```
lonng, line 220: The returned `file` has been closed before the caller using it.

pkg/repository/mirror.go
```diff
@@ -205,11 +213,24 @@ func (l *httpMirror) Download(resource, targetDir string) error {
...
+	path := filepath.Join(l.tmpDir, resource)
+	file, err := os.OpenFile(path, os.O_RDONLY, os.ModePerm)
```
lonng, line 222: It seems the `file` references to the same under-level file `r` returned from `l.download`.

AstroProfundis:
> My impression is that we should not write to disk in fetch, only in download

That make sense to me, so we should not use `Fetch()` when downloading tarballs from the begining

pkg/utils/ioutil.go
```diff
@@ -120,3 +120,13 @@ func Copy(src, dst string) error {
 	}
 	return out.Close()
 }
+
+// Move moves a file from src to dst, this is done by copying the file and then
+// delete the old one. Use os.Rename() to rename file within the same filesystem
+// instead this, it's more lightweight but can not be used accross devices.
+func Move(src, dst string) error {
+	if err := Copy(src, dst); err != nil {
```
nrc, line 128: Could we check if src and dst can be renamed and do that if possible, the copy and delete if not? Doesn't need to be done now, can be later.

AstroProfundis: Yes it's possible to try rename first and check for the error, I'll see if there's a simple way of doing so.

PR #229, opened 2020-05-19
Init keys of the FsManifests

july2993:
Init keys of the FsManifests

ms.keys is never set before and will be used to check the signature of manifests.

### DIFF
--- a/pkg/repository/v1manifest/local_manifests.go
+++ b/pkg/repository/v1manifest/local_manifests.go
@@ -21,6 +21,7 @@ import (
 	"path/filepath"
 
 	"github.com/pingcap-incubator/tiup/pkg/repository/crypto"
+	"github.com/pingcap/errors"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/repository/v1manifest/local_manifests.go
```diff
@@ -51,14 +59,39 @@ type LocalManifests interface {
...
 
-	return ioutil.WriteFile(filepath.Join(ms.root, filename), bytes, 0644)
+	err = ioutil.WriteFile(filepath.Join(ms.root, filename), bytes, 0644)
+	if err != nil {
+		return err
+	}
+
+	if manifest.Signed.Base().Ty == ManifestTypeRoot {
+		return ms.initKeys()
+	}
```
nrc, line 92: I think this should be in v1repository rather than here - the behaviour should be the same independent of where the manifests are coming from

PR #230, opened 2020-05-20
Update and install components using v1 manifests

nrc:
WIP

### DIFF
--- a/pkg/localdata/profile.go
+++ b/pkg/localdata/profile.go
@@ -16,6 +16,7 @@ package localdata
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/pingcap-incubator/tiup/pkg/repository/v1manifest"
 	"io/ioutil"
 	"log"
 	"os"
@@ -278,4 +279,9 @@ func (p *Profile) SelectInstalledVersion(component string, version v0manifest.Ve
 		return "", errInstallFirst
 	}
 	return version, nil
+}
+
+// LocalManifests creates a LocalManifests object representing the v1 manifests stored locally.
+func (p *Profile) LocalManifests() (v1manifest.LocalManifests, error) {
+	return v1manifest.NewManifests(p.root)
 }
--- a/pkg/localdata/profile.go
+++ b/pkg/localdata/profile.go
@@ -16,7 +16,6 @@ package localdata
 import (
 	"encoding/json"
 	"fmt"
-	"github.com/pingcap-incubator/tiup/pkg/repository/v1manifest"
 	"io/ioutil"
 	"log"
 	"os"
@@ -249,6 +248,20 @@ func (p *Profile) InstalledVersions(component string) ([]string, error) {
 	return versions, nil
 }
 
+// VersionIsInstalled returns true if exactly version of component is installed.
+func (p *Profile) VersionIsInstalled(component, version string) (bool, error) {
+	installed, err := p.InstalledVersions(component)
+	if err != nil {
+		return false, err
+	}
+	for _, v := range installed {
+		if v == version {
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
 // SelectInstalledVersion selects the installed versions and the latest release version
 // will be chosen if there is an empty version
 func (p *Profile) SelectInstalledVersion(component string, version v0manifest.Version) (v0manifest.Version, error) {
@@ -279,9 +292,4 @@ func (p *Profile) SelectInstalledVersion(component string, version v0manifest.Ve
 		return "", errInstallFirst
 	}
 	return version, nil
-}
-
-// LocalManifests creates a LocalManifests object representing the v1 manifests stored locally.
-func (p *Profile) LocalManifests() (v1manifest.LocalManifests, error) {
-	return v1manifest.NewManifests(p.root)
 }
--- a/pkg/repository/v1_repository.go
+++ b/pkg/repository/v1_repository.go
@@ -19,6 +19,7 @@ import (
 	"io"
 	"path/filepath"
 	"runtime"
+	"sort"
 	"strconv"
 
 	"github.com/pingcap-incubator/tiup/pkg/repository/v1manifest"
@@ -31,12 +32,21 @@ type V1Repository struct {
 	Options
 	mirror Mirror
 	local  v1manifest.LocalManifests
-	root   *v1manifest.Root
+}
+
+// ComponentSpec describes a component a user would like to have or use.
+type ComponentSpec struct {
+	// ID is the id of the component
+	ID string
+	// Version describes the versions which are desirable; "" = use the most recent, compatible version.
+	Version string
+	// Force is true means overwrite any existing installation.
+	Force bool
 }
 
 // NewV1Repo creates a new v1 repository from the given mirror
 // local must exists a trusted root.
-func NewV1Repo(mirror Mirror, opts Options, local v1manifest.LocalManifests) (*V1Repository, error) {
+func NewV1Repo(mirror Mirror, opts Options, local v1manifest.LocalManifests) *V1Repository {
 	if opts.GOOS == "" {
 		opts.GOOS = runtime.GOOS
 	}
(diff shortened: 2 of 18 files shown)


nrc:
Closes https://github.com/pingcap-incubator/tiup/issues/202

pkg/repository/v1_repository.go
```diff
@@ -54,6 +65,87 @@ func NewV1Repo(mirror Mirror, opts Options, local v1manifest.LocalManifests) *V1
...
+		if err != nil {
+			// TODO could report error and continue
+			return err
+		}
+	}
+	return nil
+}
+
+func (r *V1Repository) ensureManifests() error {
+	// TODO
```
july2993, line 122: this will ensure we can the fresh and right snapshot&index?

pkg/repository/v1_repository.go
```diff
@@ -54,6 +65,87 @@ func NewV1Repo(mirror Mirror, opts Options, local v1manifest.LocalManifests) *V1
...
 
+// UpdateComponents updates the components described by specs.
+// TODO tests
+func (r *V1Repository) UpdateComponents(specs []ComponentSpec) error {
+	err := r.ensureManifests()
+	if err != nil {
+		return err
+	}
+	for _, spec := range specs {
+		manifest, err := r.updateComponentManifest(spec.ID)
```
july2993, line 76: `updateComponentManifest` assume we can get a higher version manifest from the mirror than the local one,
seems we can not call this directly

nrc: Good catch!

pkg/repository/v1_repository.go
nrc: It will ensure that we have an up to date index, root, and snapshot, i.e., that we can call updateComponentManifest

nrc:
PTAL @july2993 The test  for `updateComponents` is still not done, but I think I won't finish it today and this PR is getting large and I don't everyone's branches diverging too much.

pkg/repository/v1manifest/manifest.go
```diff
@@ -258,18 +286,18 @@ func (manifest *Timestamp) SnapshotHash() FileHash {
 	return manifest.Meta[ManifestURLSnapshot]
 }
 
-// VersionedURL looks up url in the snapshot and returns a modified url with the version prefix
-func (manifest *Snapshot) VersionedURL(url string) (string, error) {
+// VersionedURL looks up url in the snapshot and returns a modified url with the version prefix, and that file's length.
+func (manifest *Snapshot) VersionedURL(url string) (string, *FileVersion, error) {
 	entry, ok := manifest.Meta[url]
```
july2993, line 291: seems key of Meta if just a file name. like `index.json`

july2993: we should first extract the filename from the url parameter?

pkg/repository/v1manifest/manifest.go
```diff
@@ -198,7 +224,9 @@ func (s *SignedBase) isValid() error {
 	// When updating root, we only check the newest version is not expire.
 	// This checking should be done by the update root flow.
 	if s.Ty != ManifestTypeRoot {
-		if err := CheckExpire(s.Expires); err != nil {
+		if err := CheckExpiry(s.Expires); err != nil {
+			fmt.Println("foo5")
```
july2993, line 228: need to be removed finaly

pkg/repository/v1manifest/manifest.go
```diff
@@ -185,6 +191,26 @@ func (s *SignedBase) expiryTime() (time.Time, error) {
...
+// ExpiresAfter checks that manifest 1 expires after manifest 2 (or are equal) and returns an error otherwise.
+func ExpiresAfter(m1, m2 ValidManifest) error {
+	time1, err := time.Parse(time.RFC3339, m1.Base().Expires)
+	if err != nil {
+		fmt.Println("foo2")
+		return errors.AddStack(err)
+	}
+	time2, err := time.Parse(time.RFC3339, m2.Base().Expires)
+	if err != nil {
+		fmt.Println("foo3")
```
july2993, line 203: ditto

pkg/repository/v1manifest/manifest.go
```diff
@@ -185,6 +191,26 @@ func (s *SignedBase) expiryTime() (time.Time, error) {
 	return time.Parse(time.RFC3339, s.Expires)
 }
 
+// ExpiresAfter checks that manifest 1 expires after manifest 2 (or are equal) and returns an error otherwise.
+func ExpiresAfter(m1, m2 ValidManifest) error {
+	time1, err := time.Parse(time.RFC3339, m1.Base().Expires)
+	if err != nil {
+		fmt.Println("foo2")
```
july2993, line 198: ditto
we can remove this king of log after as follow up of this PR?

pkg/localdata/profile.go
```diff
@@ -248,6 +248,20 @@ func (p *Profile) InstalledVersions(component string) ([]string, error) {
...
+	installed, err := p.InstalledVersions(component)
+	if err != nil {
+		return false, err
+	}
+	for _, v := range installed {
+		if v == version {
+			return true, nil
+		}
+	}
+	return false, err
```
july2993, line 262: `return false, nil`   will be more clear for me

pkg/repository/v1_repository.go
```diff
@@ -50,32 +60,134 @@ func NewV1Repo(mirror Mirror, opts Options, local v1manifest.LocalManifests) (*V
...
+			if err != nil {
+				return err
+			}
+			if installed {
+				// TODO could report error and continue
+				return fmt.Errorf("component %s version %s is already installed", spec.ID, version)
+			}
+		}
+
+		reader, err := r.downloadComponent(versionItem)
```
july2993, line 109: defer reader.Close()

pkg/repository/v1_repository.go
```diff
@@ -50,32 +60,134 @@ func NewV1Repo(mirror Mirror, opts Options, local v1manifest.LocalManifests) (*V
...
+	if err != nil {
+		return false, err
+	}
+
+	// Check that the version of root we have is the same as declared in the snapshot.
+	root, err := r.loadRoot()
+	if err != nil {
+		return false, err
+	}
+	snapRootVersion := snapshot.Meta[v1manifest.ManifestURLRoot].Version
```
july2993, line 147: should use `ManifestFilenameRoot` ?

pkg/repository/v1_repository.go
```diff
@@ -50,32 +60,134 @@ func NewV1Repo(mirror Mirror, opts Options, local v1manifest.LocalManifests) (*V
...
+		return false, fmt.Errorf("root version mismatch. Expected: %v, found: %v", snapRootVersion, root.Version)
+	}
+
+	// Update index (if needed).
+	var index v1manifest.Index
+	exists, err := r.local.LoadManifest(&index)
+	if err != nil {
+		return false, err
+	}
+	snapIndexVersion := snapshot.Meta[v1manifest.ManifestURLIndex].Version
```
july2993, line 158: the same here, the key of Meta is the file name, so can not access by url

pkg/repository/v1_repository.go
```diff
@@ -121,194 +226,173 @@ func fnameWithVersion(fname string, version uint) string {
...
-	}
-
-	return &index, nil
+	return r.local.SaveManifest(manifest, v1manifest.ManifestFilenameIndex)
 }
 
-// Precondition: the snapshot manifest exists and is up to date
+// Precondition: the snapshot and index manifests exist and are up to date.
+// Returns nil if the component manifest is up to date.
 func (r *V1Repository) updateComponentManifest(id string) (*v1manifest.Component, error) {
```
july2993, line 329: do you want to return the Component event the local one is up to date?
how can we get the v1manifest.Component?  will add some other kind of method?

and i think `UpdateComponents` still can not call this directly if it return `nil, nil` when the manifest is up to date

pkg/repository/v1manifest/manifest.go
nrc: I think all fields which identify a file in any manifest should be a URL, not just a filename. The reason is that if in the future we add directories, then the filename is not enough.

pkg/repository/v1_repository.go
nrc: This is just a Reader, not a ReadCloser and doesn't need (and can't be) closed

pkg/repository/v1_repository.go
nrc: The update functions leave the manifests on disk, so after calling `updateComponentManifest` and there is no error, then the guarantee is that the manifest on disk is the most recent (and validated). To access the manifest, use LocalManifests. You are correct, there is a bug in `UpdateManifests` such that we won't install a component if the manifest is up to date, I'll fix that...

nrc:
@july2993 thanks for the review! I've addressed most of the comments and rebased. I think that we disagree on whether a file key should be a filename or url. See https://github.com/pingcap-incubator/tiup/pull/230#discussion_r428552779

pkg/repository/v1manifest/manifest.go
july2993: in https://github.com/pingcap-incubator/tiup/blob/master/doc/design/manifest.md#snapshotjson it's a filename,
also `tools/migrate` will generate as filename.
then we need to fix both

july2993:
> @july2993 thanks for the review! I've addressed most of the comments and rebased. I think that we disagree on whether a file key should be a filename or url. See [#230 (comment)](https://github.com/pingcap-incubator/tiup/pull/230#discussion_r428552779)

in https://github.com/pingcap-incubator/tiup/blob/master/doc/design/manifest.md#snapshotjson it's a filename,
also tools/migrate will generate as a filename.
if using url then we need to fix both of this later.

see https://github.com/pingcap-incubator/tiup/pull/230#discussion_r428558605
I think should have some method like `GetComponentManifest` to make sure we can get the manifest(the internal make sure the returning one is update to date), the current `updateComponentManifest` looks should only be used when we get a new snapshot and update the componentManifest locally if need.

PR #231, opened 2020-05-20
Refine genkey with keyinfo instead of pem file

lucklove:

- genkey: store keyinfo as ~/.tiup/private.json instead of ~/.tiup/private.pem
- genkey: support -p flag to show public key content:
```
KeyID: 79fa37bc6b41d52f0980caf1574768a001478e1d11af07c876bf70f06076f258
KeyContent: 
{
        "keyid_hash_algorithms": [
                "sha256",
                "sha512"
        ],
        "keytype": "rsa",
        "keyval": {
                "public": "-----BEGIN PUBLIC KEY-----\nMIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEA0kgU3v3SxYzm5J6P+GPP\nmy6toBnKYtTViAVmpUJiIEjdZ9NLpoJU0na9q0CD8sIgo2Js/W/owJUvSj6rm8us\nsu/Ve5KsoJN6zca2am1uZ5IKnc48i0mCv76WXawCxM+NFGqSCMJcltlhj3fC/GDS\ngu+BiIbrgR1PgJf6Jk6l7uMJdN3TL6JJQcEC4lz+2hj5zoVNYkq06ZC79j2tPDCI\nkTAYGF/TAAVLH08/kGH5ZeRPlVKJ7cwW3OniLM5NeFnS8+shRNb6AYr7xju3Ikbw\nDo14ipIghBI0iAxn6Lvr/iilc7TM7RWJ4OiTrmK3SQSJ+U6H2N2/I5OGEHBEKzbA\nOQIDAQAB\n-----END PUBLIC KEY-----\n"
        },
        "scheme": "rsassa-pss-sha256"
}
```

### Check List 

Tests 

Code changes

Side effects

Related changes

### DIFF
--- a/cmd/repo.go
+++ b/cmd/repo.go
@@ -14,13 +14,12 @@
 package cmd
 
 import (
+	"encoding/json"
 	"fmt"
-	"io/ioutil"
 	"os"
 	"time"
 
 	"github.com/pingcap-incubator/tiup/pkg/meta"
-	"github.com/pingcap-incubator/tiup/pkg/repository/crypto"
 	"github.com/pingcap-incubator/tiup/pkg/repository/v1manifest"
 	"github.com/pingcap-incubator/tiup/pkg/utils"
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


lucklove:
This is used to support manually sign root.json, since it need you to add your public key into it.

nrc:
> This is used to support manually sign root.json, since it need you to add your public key into it.

The way I was thinking that signing would work is that TiUp would generate all keys necessary, write the public keys to the manifests and give return the private keys to the user, probably by printing to the command line (saving to disk is potentially a security issue). The user should never have to supply there own keys (because they should not be reusing keys and we need lots of them), plus we have to support key rotation.

Do you have a different idea for things will work? Or is this a temporary solution?

PR #237, opened 2020-05-21
Use SignatureError error

july2993:
Use SignatureError error


pkg/repository/v1_repository.go
```diff
@@ -78,7 +78,7 @@ func (r *V1Repository) loadRoot() error {
 // Postcondition: if returned error is nil, then the local snapshot and timestamp are up to date.
 func (r *V1Repository) updateLocalSnapshot() (*v1manifest.Snapshot, error) {
 	hash, err := r.checkTimestamp()
-	if _, ok := err.(*v1manifest.SignatureError); ok {
+	if v1manifest.IsSignatureError(errors.Cause(err)) {
```
lonng, line 81: I think it's better to call `errors.Cause` in function `v1manifest.IsSignatureError`.

PR #238, opened 2020-05-21
Fix migrate

july2993:
1. Fix wrong length(less then the finally file write).
2. Fix losing snapshot role in root.json

### DIFF
--- a/tools/migrate/main.go
+++ b/tools/migrate/main.go
@@ -34,6 +34,9 @@ import (
 	"github.com/spf13/cobra"
 )
 
+// for simplify just set the length of manifest file a value lager than the true value
+var limitLength uint = 1024 * 1024
+
 func main() {
 	cmd := &cobra.Command{
 		Use:           "migrate <src-dir> <dst-dir>",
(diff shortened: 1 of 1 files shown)


tools/migrate/main.go
```diff
@@ -309,14 +305,24 @@ func migrate(srcDir, dstDir string) error {
...
+
+	hash, _, err := hashManifest(signedManifests[v1manifest.ManifestTypeSnapshot])
 	if err != nil {
 		return errors.Trace(err)
 	}
 	timestamp.Meta = map[string]v1manifest.FileHash{
 		v1manifest.ManifestFilenameSnapshot: {
 			Hashes: hash,
-			Length: n,
+			Length: limitLength,
```
lonng, line 325: Is it correct? I think the length should be the real file length of snapshot.json.

july2993: >= the length of snapshot.sjon is fine.
even zero should be ok which means no limit.

nrc: 0 is not ok (might permit DOS-style attacks), but the same or greater (within reason) is ok.

july2993: when 0, client should use a limit size at the client side to protest this attack, then I think it's ok?
seems I read this in the TUF spec before, but I am not sure.

nrc: I think it is true, but we have not implemented this

PR #241, opened 2020-05-21
test repo using manifest generated by tools/migrate

july2993:
Add `testdata/manifest`  which generate by tools/migrate as testdata used by unit-test

add test for using that manifest dir as a mirror make sure we can load the manifest correctly
todo: test for we can load the component manifest too.


pkg/repository/v1_repository_test.go
```diff
@@ -566,3 +570,56 @@ func hash(s string) string {
 
 	return hex.EncodeToString(shaWriter.Sum(nil))
 }
+
+func TestWithMigrate(t *testing.T) {
```
nrc, line 574: Could you add a comment to this test to explain what is happening? It is a bit less obvious than the other tests I think.

july2993: added by https://github.com/pingcap-incubator/tiup/pull/241/commits/a7f51bf6196078be0b138acb81b9cd3ec08ff8ba

PR #249, opened 2020-05-24
manifest caching in local_manifests

nrc:
Avoids repeated file reads by caching successful loads. We never clear the cache since we'll error out on an invalid or out of date manifest.

Probably don't want to land this until we fix the integration test which is commented out, since that is the way that this code will get tested.

PTAL @july2993


july2993:
@nrc I try to uncomment the test and looks good in https://github.com/pingcap-incubator/tiup/pull/249/commits/886d57886066b16dcf29ff2569d2c70c9ca56300, just forget to init the map, I will merge this

PR #251, opened 2020-05-24
Set high expire time

july2993:
support to set expires time for migrate new, also can sign the root

to generate:
- tiup mirrors /Users/huangjiahao/local-mirrors --full   (this take times...will download all **.tar.gz)
- ./migrate new --expires 100000h --signed-root /Users/huangjiahao/local-mirrors ./out2
- copy out2/manefest to testdata/


tools/migrate/main.go
```diff
@@ -255,6 +268,9 @@ func migrate(srcDir, dstDir string) error {
 	}
 
 	snapshot := v1manifest.NewSnapshot(initTime)
+	if expires != 0 {
```
lonng, line 271: Do we need to set different expiration for different roles?

july2993: only used to set as a very high value for the test.
if not settled, it will used different value in New**

PR #253, opened 2020-05-24
Add back functions to keep tiup-cluster working

lucklove:
Some functions are removed, that broken tiup-cluster. This pull request adds them back and add more functions to support tiup-cluster


lucklove:
> Could you add a comment to the new functions for the v1 stuff that they are used by tiup-cluster please? Otherwise they might get removed or refactored.
> 
> Otherwise, lgtm

Ok

PR #254, opened 2020-05-25
Check role is correct for keys and refactor key store

nrc:
This also changes the manifest format, so we'll need to rebuild any old manifests.

PTAL @july2993


july2993:
please resolve the conflicts

nrc:
@july2993 rebased, PTAL

pkg/repository/v1_repository.go
```diff
@@ -176,11 +181,17 @@ func (r *V1Repository) selectVersion(id string, versions map[string]v1manifest.V
...
-		sort.Strings(versionSlice)
-		// TODO I think it is not actually correct semver order because versions can have a text suffix.
-		version := versionSlice[len(versionSlice)-1]
-		item := versions[version]
-		return version, &item, nil
+
+		var latest string
+		var latestItem v1manifest.VersionItem
+		for version, item := range versions {
+			if latest == "" || semver.Compare("v"+version, "v"+latest) > 0 {
```
july2993, line 188: the current generate `version` has include the "v" actually

pkg/repository/v1manifest/key_store.go
```diff
@@ -0,0 +1,137 @@
...
+	keys, ok := (*s)[role]
+	if !ok {
+		return errors.Errorf("Unknown role %s", role)
+	}
+
+	var validSigs uint = 0
+	for _, sig := range signatures {
+		key, ok := keys.keys[sig.KeyID]
+		if !ok {
+			continue
```
july2993, line 124: then this is not "all supplied signatures are correct for signed" ?

nrc: I’ll change the comment. This is for handling root manifest old and new keys

pkg/repository/v1manifest/local_manifests.go
```diff
@@ -59,12 +64,16 @@ type FsManifests struct {
 // NewManifests creates a new FsManifests with local store at root.
 // There must exists the trusted root.json.
 func NewManifests(profile *localdata.Profile) *FsManifests {
-	return &FsManifests{profile: profile, cache: make(map[string]string)}
+	return &FsManifests{profile: profile, keys: NewKeyStore(), cache: make(map[string]string)}
```
july2993, line 67: need a way to init keys by loading the trusted root locally?

july2993: guess it's caused by this
```
➜  tiup git:(role) ✗ tt tivk
The component `tivk` is not installed; downloading from repository.
Failed to start component `tivk`
Error: Unknown role timestamp
➜  tiup git:(role) ✗
```

lonng:
@nrc Please resolve the conflicts again 😢

pkg/repository/v1_repository.go
nrc: I don't understand, versions should come straight from the manifest and therefore not have the "v". Where is it added?

pkg/repository/v1manifest/local_manifests.go
nrc: I think I've addressed this in the last commit, when creating a new FsManifests, we now do some initialisation by reading in the root manifest and adding its keys to the key store.

pkg/repository/v1_repository.go
july2993: the origin manifest already have the "v" generated by `tools/migrate`

nrc:
@july2993 comments addressed and rebased, please take another look (also I had a question in https://github.com/pingcap-incubator/tiup/pull/254#discussion_r429845034).

july2993:
> @july2993 comments addressed and rebased, please take another look (also I had a question in [#254 (comment)](https://github.com/pingcap-incubator/tiup/pull/254#discussion_r429845034)).

I have replied in the comment.

pkg/repository/v1_repository.go
```diff
@@ -180,15 +191,16 @@ func (r *V1Repository) selectVersion(id string, versions map[string]v1manifest.V
 	// TODO we should check what version the user has currently installed and only update to the same semver major version unless they force upgrade.
 
 	if target == "" {
-		versionSlice := make([]string, 0, len(versions))
-		for v := range versions {
-			versionSlice = append(versionSlice, v)
+		var latest string
+		var latestItem v1manifest.VersionItem
+		for version, item := range versions {
+			if latest == "" || semver.Compare("v"+version, "v"+latest) > 0 {
```
july2993, line 197:
```suggestion
			if latest == "" || semver.Compare(version, latest) > 0 {
```

pkg/repository/v1_repository.go
july2993: I change this line to remove the "v" and fix the related tests, others look good. I will merge this first.
If we want to use it as without the "v", we should change it both later.

PR #257, opened 2020-05-25
Self update

july2993:
support self update
TODO: make sure the server supports  tiup as a component too and test it.


pkg/repository/v1_repository.go
```diff
@@ -551,6 +554,16 @@ func (r *V1Repository) FetchIndex() (index *v1manifest.Index, err error) {
 	return index, nil
 }
 
+// DownloadTiup downloads the tiup tarball and expands it into targetDir
+func (r *V1Repository) DownloadTiup(targetDir string) error {
+	var spec = ComponentSpec{
```
lonng, line 559: The `targetDir` useless? I think you forget to assign the `targetDir` to `ComponentSpec.TargetDir`.

july2993: fixed in https://github.com/pingcap-incubator/tiup/pull/257/commits/11beea8feb1616ed3125ea23ab942913d25b45f0

PR #258, opened 2020-05-25
Refactor tiup list subcommand to use v1 repository 

lucklove:
Switch repository from v0 to V1

Use env.V1Repository instead of env.Repository

### Check List 

Tests 

Code changes

Side effects

Related changes


july2993:
please fix resolve conflicts and fix the tests

pkg/repository/v1manifest/index.go
```diff
@@ -0,0 +1 @@
+package v1manifest
```
nrc, line 1: I think this file can be removed

pkg/repository/v1manifest/component.go
```diff
@@ -0,0 +1,7 @@
+package v1manifest
+
+// ContainsVersion returns if the versions contain the specific version
+func (manifest *Component) ContainsVersion(version Version) bool {
```
nrc, line 4: I think this can be removed - it is not used

pkg/repository/v1manifest/types.go
```diff
@@ -116,6 +116,8 @@ type Component struct {
 
 // ComponentItem object
 type ComponentItem struct {
+	Name      string `json:"name"`
+	Desc      string `json:"description"`
```
nrc, line 120: I think it is better not to add these here since they can be changed when a component is updated and then we would need to touch index.json.

pkg/repository/v1manifest/version.go
```diff
@@ -0,0 +1,10 @@
+package v1manifest
+
+import (
+	"github.com/pingcap-incubator/tiup/pkg/repository/v0manifest"
+)
+
+type (
+	// Version represents a version string, like: v3.1.2
+	Version = v0manifest.Version
```
nrc, line 9:
```suggestion
	Version string
```

I like to keep the v0 and v1 manifest systems separate. However, I think it is also ok to just use a `string` for the version rather than create a type because we don't ensure any further invariants when we use the `Version` type.

pkg/repository/v1manifest/types.go
lucklove: Yes. But it introduces another problem: the tiup list must fetch all componet.json to display component discription

cmd/list.go
```diff
@@ -82,102 +79,53 @@ func (lr *listResult) print() {
...
-			return nil, err
-		}
-		err = env.Profile().SaveManifest(manifest)
-		if err != nil {
-			return nil, err
-		}
-	}
-
-	installed, err := env.Profile().InstalledComponents()
+	index, err := env.V1Repository().FetchIndexManifest()
```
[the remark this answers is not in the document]
lonng: Agree with @nrc, I think we should use the locally cached manifests, the `update` to the latest manifests if the user pass the `--refresh` flag.

pkg/meta/meta.go
```diff
@@ -63,12 +63,12 @@ func InitEnv(options repository.Options) (*Environment, error) {
 	var v1repo *repository.V1Repository
 	var err error
 
-	if os.Getenv(EnvNameV1) == "" {
-		repo, err = repository.NewRepository(mirror, options)
-	} else {
+	if os.Getenv(EnvNameV0) == "" {
```
lonng, line 66:
```suggestion
	if env := os.Getenv(EnvNameV0); env == "" || env == "disable" || env = "false" {
```

pkg/repository/v1manifest/types.go
nrc: Hmmm, true. I think this is OK for now, but long term maybe we can move them to snapshot.json.

cmd/list.go
```diff
@@ -82,102 +80,93 @@ func (lr *listResult) print() {
...
 
 	installed, err := env.Profile().InstalledComponents()
 	if err != nil {
 		return nil, err
 	}
-	manifest := env.Profile().Manifest()
+
 	var cmpTable [][]string
-	cmpTable = append(cmpTable, []string{"Name", "Installed", "Platforms", "Description"})
+	cmpTable = append(cmpTable, []string{"Name", "Owner", "Installed", "Platforms", "Description"})
```
nrc, line 94: I think it would be good for `tiup list` to give a very simple list of component names and `tiup list --verbose` to give this view. It is good info, but I think it might be overwhelming for new users.

cmd/list.go
```diff
@@ -82,102 +80,93 @@ func (lr *listResult) print() {
...
-		}
-		if onlyInstalled && !localComponents.Exist(comp.Name) {
+	for name, comp := range index.Components {
+		if onlyInstalled && !localComponents.Exist(name) {
 			continue
 		}
 		installStatus := ""
-		if localComponents.Exist(comp.Name) {
-			versions, err := env.Profile().InstalledVersions(comp.Name)
+		if localComponents.Exist(name) {
```
nrc, line 102: The 'name' here is actually the id of the component, but after seeing this PR I think it is a mistake to have both name and id for components and we should remove name (I can do that if you agree, it doesn't have to be done in this PR).

cmd/list.go
```diff
@@ -82,102 +80,93 @@ func (lr *listResult) print() {
...
+			versions, err := env.Profile().InstalledVersions(name)
 			if err != nil {
 				return nil, err
 			}
 			installStatus = fmt.Sprintf("YES(%s)", strings.Join(versions, ","))
```
nrc, line 107: I think rather than using `YES(versions)` we can just display the installed versions, it should be clearer.

cmd/list.go
```diff
@@ -82,102 +80,93 @@ func (lr *listResult) print() {
...
+		}
+
 		cmpTable = append(cmpTable, []string{
-			comp.Name,
+			name,
+			comp.Owner,
 			installStatus,
-			strings.Join(comp.Platforms, ","),
-			comp.Desc,
+			strings.Join(platforms, ","),
```
nrc, line 124:
```suggestion
			strings.Join(platforms, ", "),
```

cmd/list.go
nrc: This also has the advantage that `tiup list` would be very fast because it would only need the index manifest and installed versions.

cmd/list.go
nrc: #267

pkg/repository/v1manifest/version.go
lucklove: This file is deleted now.

cmd/list.go
lucklove: what do you think about this? @lonng

lucklove: @nrc We will have `verbose` in the next minor version

pkg/repository/v1manifest/types.go
lucklove: OK

cmd/list.go
lucklove: OK

cmd/list.go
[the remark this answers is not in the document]
lucklove: Supported --refresh now

july2993:
@lucklove 
pls fix ci (maybe just run `go mod tidy`)

```
[2020-05-26T10:37:09.602Z] go: extracting gopkg.in/fsnotify.v1 v1.4.7

[2020-05-26T10:37:09.602Z] go: downloading github.com/fsnotify/fsnotify v1.4.7

[2020-05-26T10:37:09.602Z] go: extracting github.com/fsnotify/fsnotify v1.4.7

[2020-05-26T10:37:09.602Z] go: extracting github.com/facebookgo/subset v0.0.0-20200203212716-c811ad88dec4

[2020-05-26T10:37:09.603Z] go: extracting github.com/chzyer/logex v1.1.10

[2020-05-26T10:37:09.603Z] go: extracting golang.org/x/text v0.3.2

[2020-05-26T10:37:10.168Z] Files go.sum and /tmp/go.sum.before differ

[2020-05-26T10:37:10.168Z] make: *** [tidy] Error 1

[2020-05-26T10:37:10.168Z] + set +x

[2020-05-26T10:37:10.168Z] GO111MODULE=off go get github.com/wadey/gocovmerge

[2020-05-26T10:37:25.168Z] gocovmerge cover/cov.* | grep -vE ".*.pb.go|.*__failpoint_binding__.go|mock.go" > "cover/all_cov.out"

[2020-05-26T10:37:25.168Z] bash: cover/all_cov.out: No such file or directory

[2020-05-26T10:37:25.168Z] 2020/05/26 18:37:23 failed to parse profiles: open cover/cov.*: no such file or directory

[2020-05-26T10:37:25.168Z] make: *** [coverage] Error 1

script returned exit code 2
```

PR #259, opened 2020-05-25
Change the behavior of 'tiup uninstall --self' and keep user data

lonng:

Fix https://github.com/pingcap-incubator/tiup/issues/150

Change the behavior of `tiup uninstall --self` and keep user data.

### DIFF
--- a/cmd/uninstall.go
+++ b/cmd/uninstall.go
@@ -55,7 +55,7 @@ which is used to uninstall tiup.
 					}
 					fmt.Printf("Remove directory '%s' successfully!\n", env.Profile().Path(dir))
 				}
-				fmt.Println("Uninstalled self successfully!")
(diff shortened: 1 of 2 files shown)


cmd/uninstall.go
```diff
@@ -48,10 +48,14 @@ which is used to uninstall tiup.
   tiup uninstall --all`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if self {
-				if err := os.RemoveAll(env.Profile().Root()); err != nil {
-					return errors.Trace(err)
+				deletable := []string{"bin", "manifest", "components"}
```
AstroProfundis, line 51: so the packages download by `cluster` are still there, how about just not delete anything by default, and prompt user an option to delete everything?

PR #261, opened 2020-05-25
*: add basic support of nightly versions

AstroProfundis:
Add basic support of "nightly" versions to manifest definition and the migrate tool

 - Add a "nightly" section to component manifest, that points to the latest daily build
 - Add support of migrating nightly builds from v0 manifest

### Check List 

Tests 

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/cmd/repo.go
+++ b/cmd/repo.go
@@ -96,6 +96,7 @@ func newRepoSignCmd(env *meta.Environment) *cobra.Command {
 
 // the `repo add` sub command
 func newRepoAddCompCmd(env *meta.Environment) *cobra.Command {
+	var nightly bool // if this is a nightly version
 	cmd := &cobra.Command{
 		Use:   "add <component-id> <platform> <version> <file>",
 		Short: "Add a file to a component",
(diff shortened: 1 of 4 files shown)


cmd/repo.go
```diff
@@ -105,14 +106,18 @@ func newRepoAddCompCmd(env *meta.Environment) *cobra.Command {
 				return cmd.Help()
 			}
 
-			return addCompFile(repoPath, args[0], args[1], args[2], args[3])
+			return addCompFile(repoPath, args[0], args[1], args[2], args[3], nightly)
```
lonng, line 109: why not check the `nightly` prefix?

doc/design/manifest.md
```diff
@@ -255,28 +255,36 @@ Where xxx is the id of the component, n is the version of the manifest, file is
...
     "aarch64-unknown-linux": { ... },
 },
 ```
 
 The platform id should be one of the supported TiUp target triples. Version ids must be valid semver. Dependencies are a map from component id to a semver version string.
 
+The "nightly" points to the version number of latest daily build, that version should be in the version list of all supported platforms. The version number of nightly build should (but not forced to) be in the following format:
+
+```
+vX.Y.Z-nightly+YYYYmmdd
```
nrc, line 284: This format is not valid semver (even ignoring the v prefix), I suggest using `-` instead of `+` (only `a-Z`, `0-9`, `-`, and `.` are allowed)

july2993, line 284:
```suggestion
vX.Y.Z-nightly-YYYY-mm-dd
```

july2993:
just for record:
we will have a version name: "nightly" as migrated from v0, then later will the ones like vX.Y.Z-nightly-YYYY-mm-dd ..

doc/design/manifest.md
AstroProfundis: I'm using the date as a meta part in the semver spec, but I'm also ok with other formats.

PR #264, opened 2020-05-25
Support nightly

july2993:
cc #261

### DIFF
--- a/pkg/meta/meta.go
+++ b/pkg/meta/meta.go
@@ -122,7 +122,7 @@ func (env *Environment) UpdateComponents(specs []string, nightly, force bool) er
 			component, v := ParseCompVersion(spec)
 			v1specs = append(v1specs, repository.ComponentSpec{ID: component, Version: v.String(), Force: force})
 		}
-		return env.v1Repo.UpdateComponents(v1specs)
+		return env.v1Repo.UpdateComponents(v1specs, nightly)
 	}
 
 	manifest, err := env.latestManifest()
@@ -178,7 +178,7 @@ func (env *Environment) downloadComponentv1(component string, version v0manifest
 		Force:   overwrite,
 	}
 
-	return env.v1Repo.UpdateComponents([]repository.ComponentSpec{spec})
+	return env.v1Repo.UpdateComponents([]repository.ComponentSpec{spec}, false)
 }
 
 // downloadComponent downloads the specific version of a component from repository
--- a/pkg/meta/meta.go
+++ b/pkg/meta/meta.go
@@ -120,9 +120,9 @@ func (env *Environment) UpdateComponents(specs []string, nightly, force bool) er
 		var v1specs []repository.ComponentSpec
 		for _, spec := range specs {
 			component, v := ParseCompVersion(spec)
-			v1specs = append(v1specs, repository.ComponentSpec{ID: component, Version: v.String(), Force: force})
+			v1specs = append(v1specs, repository.ComponentSpec{ID: component, Version: v.String(), Force: force, Nightly: nightly})
 		}
-		return env.v1Repo.UpdateComponents(v1specs, nightly)
+		return env.v1Repo.UpdateComponents(v1specs)
 	}
 
 	manifest, err := env.latestManifest()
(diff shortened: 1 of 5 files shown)


pkg/meta/meta.go
```diff
@@ -122,7 +122,7 @@ func (env *Environment) UpdateComponents(specs []string, nightly, force bool) er
 			component, v := ParseCompVersion(spec)
 			v1specs = append(v1specs, repository.ComponentSpec{ID: component, Version: v.String(), Force: force})
 		}
-		return env.v1Repo.UpdateComponents(v1specs)
+		return env.v1Repo.UpdateComponents(v1specs, nightly)
```
nrc, line 125: I think nightly should be added to the component specs, not passed as another argument.

pkg/repository/v0manifest/version.go
```diff
@@ -54,8 +54,7 @@ func (v Version) IsEmpty() bool {
 
 // IsNightly returns true if the version is nightly
 func (v Version) IsNightly() bool {
-	return string(v) == version.NightlyVersion
-
+	return strings.Contains(string(v), "nightly")
```
nrc, line 57: This is not true for v1 - although the nightly version should include `nightly` it doesn't have to.

july2993: seems the only way we can check this is checking weather it contains "nightly" according to #261 @AstroProfundis

nrc: I think if we put nightly in the component spec, then we don't need to check the version to know (from the manifests, whether a version is nightly is external to the version identifier).

july2993:
> I think using `version == "nightly"` in the component spec rather than passing a nightly bool will simplify things a lot (e.g., not needing the `IsNightly` check). Otherwise looks good.

I think maybe need to instal specified version as v1.2.2-nightly-yyyy-mm-dd, not just "nightly" ?

pkg/repository/v0manifest/version.go
AstroProfundis: The idea in #261 was that, "nightly" is only a pointer to *some* version, and we have multiple such versions from daily builds but they are all treated as normal versions, except the one pointed by "nightly".

pkg/repository/v1_repository.go
```diff
@@ -16,13 +16,14 @@ package repository
...
 	"fmt"
-	"golang.org/x/mod/semver"
 	"io"
 	"path/filepath"
 	"runtime"
 	"strconv"
 	"strings"
 
+	"golang.org/x/mod/semver"
+
```
AstroProfundis, line 26: remove this empty line

nrc:
> I think maybe need to instal specified version as v1.2.2-nightly-yyyy-mm-dd, not just "nightly" ?

Ah, in that case I think we can add a nightly field to the component spec.

july2993:
@nrc PTAL 
I have add a nightly field to the component spec

pkg/repository/v1_repository.go
```diff
@@ -90,7 +93,16 @@ func (r *V1Repository) UpdateComponents(specs []ComponentSpec) error {
...
+		if spec.Nightly {
+			spec.Version = manifest.Nightly
+			// The v0 "nightly" is not versioned, force update as v0...
+			// we will add daily ones like: "v3.0.0-nightly-yyyy-mm-dd"
+			if spec.Version == "nightly" {
+				spec.Force = true
+			}
+		}
+
+		if v0manifest.Version(spec.Version).IsNightly() && !manifest.HasNightly(r.PlatformString()) {
```
nrc, line 105:
```suggestion
		if spec.Nightly && !manifest.HasNightly(r.PlatformString()) {
```

pkg/repository/v1_repository.go
```diff
@@ -188,6 +200,10 @@ func (r *V1Repository) selectVersion(id string, versions map[string]v1manifest.V
 		var latest string
 		var latestItem v1manifest.VersionItem
 		for version, item := range versions {
+			if v0manifest.Version(version).IsNightly() {
```
nrc, line 203: Can we inline the 'is nightly' check here so we don't need to use v0manifest.Version?

nrc: I think that we should add to manifests.md that a nightly version must contain the substring `nightly`. We could also check that in `Component.IsValid`

pkg/repository/v0manifest/version.go
nrc: A problem is that we must skip nightly versions when getting the latest version of a component, so we need to know which component versions are nightly and which aren't

pkg/repository/v1_repository.go
nrc: Good catch by the way, I did not think about this issue with nightlies.

PR #265, opened 2020-05-26
Misc security fixes and tidying up for v1 manifests

nrc:
This PR is a bit of a bag of different things I found while checking we were properly doing signature and hash verification. It includes:

* Handling verification of expired root manifests by checking keys for expiry when they are used, not when loaded.
* Make `ReadJSON` in `Profile` private to reduce the surface area of json decoding. I have checked that every place where we decode JSON verifies the result if necessary. I created a `ReadMetaFile` function to accomplish this, I think it is a good abstraction in any case.
* Removed some mention of `sha512` hashes which we do not support.
* Removed the unused `DownloadComponent` from `v1Repository`
* Removed an unnecessary reload of the root manifest in `ensureManifests`, this was a mistake from a previous PR (whoops).

PTAL @lucklove @july2993

### DIFF
--- a/doc/design/manifest.md
+++ b/doc/design/manifest.md
@@ -108,10 +108,6 @@ A key definition is:
 
 ```json
 "KEYID": {
-    "keyid_hash_algorithms": [
-        "sha256",
-        "sha512",
-    ],
     "keytype": "ed25519",
     "keyval": {
         "public": "KEY",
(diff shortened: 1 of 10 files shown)


doc/design/manifest.md
```diff
@@ -108,10 +108,6 @@ A key definition is:
 
 ```json
 "KEYID": {
-    "keyid_hash_algorithms": [
```
lonng, line 111: I think both `sha256/sha512` and can be supported in the current version. see: https://github.com/pingcap-incubator/tiup/pull/263/files#diff-7e28c891f69ae43e5142e9e8c085c0deR28

nrc: I don't think we need `keyid_hash_algorithms` in any case because it is not role dependent. We currently only read the sha256 hash, but we can still have multiple hashes so support sha512 easily in the future.

nrc: I added back in the line of the spec about sha256/sha512, but left `keyid_hash_algorithms` out.

PR #270, opened 2020-05-26
tools/migrate: add function to rehash tarballs and update with existing keys

AstroProfundis:
Update an exist repo (generated by `migrate new`) without regenerating root.json and keys.

This is useful when some of the files in repository has changed, or when the v0 manifest has changed, but there is already a v1 repository generated and it's not convenient to resign the `root.json` with keys held by different individuals.

Add a `migrate rehash` command, it reads the `root.json` and `index.json` and then find keys for other manifests. Then it performs operations similar as `migrate new` but without generating new keys.

### Check List 

Tests 

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/tools/migrate/main.go
+++ b/tools/migrate/main.go
@@ -194,7 +194,18 @@ func migrate(srcDir, dstDir string, rehash bool) error {
 	}
 	signedManifests := make(map[string]*v1manifest.Manifest)
 
-	if rehash {
+	snapshot := v1manifest.NewSnapshot(initTime)
+
+	if rehash { // read exist manifests
+		s := &v1manifest.Manifest{
+			Signed: &v1manifest.Snapshot{},
+		}
+		err = readManifest1(dstDir, filepath.Join("manifests", snapshot.Filename()), s)
+		if err != nil {
+			return err
+		}
+		snapshot = s.Signed.(*v1manifest.Snapshot)
+
 		r := &v1manifest.Manifest{
 			Signed: &v1manifest.Root{},
 		}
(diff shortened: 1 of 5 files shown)


tools/migrate/main.go
```diff
@@ -136,57 +150,128 @@ func readVersions(srcDir, comp string) (*v0manifest.VersionManifest, error) {
 	return m, nil
 }
 
-func migrate(srcDir, dstDir string) error {
-	m, err := readManifest0(srcDir)
+func fnameWithVersion(fname string, version uint) string {
```
lonng, line 153: Please reuse https://github.com/pingcap-incubator/tiup/blob/1c5ebde430a7b57c305a370acd901acb7aca0e3a/pkg/repository/v1_repository.go#L256

tools/migrate/main.go
```diff
@@ -136,57 +150,128 @@ func readVersions(srcDir, comp string) (*v0manifest.VersionManifest, error) {
...
+	base := filepath.Base(fname)
+	dir := filepath.Dir(fname)
+
+	versionBase := strconv.Itoa(int(version)) + "." + base
+	return filepath.Join(dir, versionBase)
+}
+
+func writeManifest(fname string, m *v1manifest.Manifest) error {
```
lonng, line 161: Please reuse https://github.com/pingcap-incubator/tiup/blob/1c5ebde430a7b57c305a370acd901acb7aca0e3a/pkg/repository/v1manifest/repo.go#L547

tools/migrate/main.go
```diff
@@ -136,57 +150,128 @@ func readVersions(srcDir, comp string) (*v0manifest.VersionManifest, error) {
...
-	if signedRoot {
-		tys = append(tys, v1manifest.ManifestTypeRoot)
-	}
+	if rehash { // read keys
+		for _, ty := range tys { // read manifest keys
+			privKi := make([]*v1manifest.KeyInfo, 0)
+			for id := range root.Roles[ty].Keys {
+				pk := &v1manifest.KeyInfo{}
+				f, err := os.Open(filepath.Join(dstDir, "keys",
+					fmt.Sprintf("%s-%s.json", id[:16], ty)))
```
lonng, line 247: Prefer to define a constant for the magic number `16`.

PR #271, opened 2020-05-26
Fix mirror clone

lonng:

Fix some minor bugs.


pkg/repository/clone_mirror.go
```diff
@@ -299,33 +299,69 @@ func cloneComponents(repo *V1Repository,
 		compManifests[name] = newManifest
 	}
 
+	// Download TiUP binary
+	for _, goos := range options.OSs {
```
nrc, line 303: We should probably share this code with the client code in v1repository, but we can refactor later.

PR #273, opened 2020-05-27
Treat TiUP as a component when migrating

lonng:

Treat TiUP as a component when migrating

### DIFF
--- a/tools/migrate/main.go
+++ b/tools/migrate/main.go
@@ -317,10 +317,10 @@ func migrate(srcDir, dstDir string, rehash bool) error {
 		}
 	}
 
-	// Tread TiUP as a component
+	// Treat TiUP as a component
 	tiup := v0manifest.ComponentInfo{
 		Name: "tiup",
-		Desc: "Components manager for TiDB ecosystem",
+		Desc: "Component manager for TiDB ecosystem",
 	}
 	tiupVersions := &v0manifest.VersionManifest{
 		Description: tiup.Desc,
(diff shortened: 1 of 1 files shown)


tools/migrate/main.go
```diff
@@ -316,10 +317,41 @@ func migrate(srcDir, dstDir string, rehash bool) error {
 		}
 	}
 
+	// Tread TiUP as a component
```
nrc, line 320:
```suggestion
	// Treat TiUP as a component
```

tools/migrate/main.go
```diff
@@ -316,10 +317,41 @@ func migrate(srcDir, dstDir string, rehash bool) error {
...
+	tiup := v0manifest.ComponentInfo{
+		Name: "tiup",
+		Desc: "Components manager for TiDB ecosystem",
```
nrc, line 323:
```suggestion
		Desc: "Component manager for TiDB ecosystem",
```

tools/migrate/main.go
```diff
@@ -316,10 +317,41 @@ func migrate(srcDir, dstDir string, rehash bool) error {
...
+		Versions: []v0manifest.VersionInfo{
+			{
+				Version: v0manifest.Version(tiupver.NewTiUPVersion().SemVer()),
+				Date:    time.Now().Format(time.RFC3339),
+				Entry:   "tiup",
+			},
+		},
+	}
+	for _, goos := range []string{"linux", "darwin"} {
+		for _, goarch := range []string{"amd64", "arm64"} {
```
nrc, line 336: I think we don't need arm/OS X

lonng: Yes, the darwin/arm64 won't be added to manifest because there is no tarball.

PR #276, opened 2020-05-27
tools/migrate: fix version and expire time of updated manfiests

AstroProfundis:
The version and expire time of `index.json` is not correctly set on update. Also fixes wrong expire time of `snapshot.json`.

Update values for exist manifests

### Check List 

Tests 

Code changes

 - Has persistent data change

### DIFF
--- a/tools/migrate/main.go
+++ b/tools/migrate/main.go
@@ -22,7 +22,6 @@ import (
 	"strings"
 	"time"
 
-	cjson "github.com/gibson042/canonicaljson-go"
 	"github.com/pingcap-incubator/tiup/pkg/repository"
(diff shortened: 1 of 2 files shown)


tools/migrate/main.go
```diff
@@ -466,16 +468,6 @@ func migrate(srcDir, dstDir string, rehash bool) error {
 			Standalone: comp.Standalone,
 			Hidden:     comp.Hide,
 		}
-
-		bytes, err := cjson.Marshal(signedManifests[component.ID])
```
lonng, line 470: Why delete this part?

AstroProfundis: The versions are updated with `SetVersions()` on line 518, this block does not have any real affect to the final output.

PR #279, opened 2020-05-27
Fix fail to load component and not handle install only

july2993:
Fix fail to load component and not handle install only

### DIFF
--- a/cmd/list.go
+++ b/cmd/list.go
@@ -93,25 +93,11 @@ func (lr *listResult) print() {
 }
 
 func showComponentList(env *meta.Environment, opt listOptions) (*listResult, error) {
-	local, err := v1manifest.NewManifests(env.Profile())
+	index := new(v1manifest.Index)
+	var err error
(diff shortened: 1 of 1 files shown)


cmd/list.go
```diff
@@ -93,27 +93,13 @@ func (lr *listResult) print() {
...
-	local, err := v1manifest.NewManifests(env.Profile())
+	index := new(v1manifest.Index)
+	var err error
+	index, err = env.V1Repository().FetchIndexManifest()
 	if err != nil {
 		return nil, err
 	}
 
-	index := new(v1manifest.Index)
-	if opt.refresh {
```
lonng, line 102: If we don't need the `refresh` flag, it should be deleted from flags.

july2993: https://github.com/pingcap-incubator/tiup/pull/279/commits/06baca5a8a39f34670483c8af8b13b3a76aea0a0 PTAL

PR #289, opened 2020-05-27
fix manifest panic

lichunzhu:
panic when we use `tiup cluster list` after re-install

1. fix script
2. capture panic and returns error prompt user to run `tiup list` to re-generate manifest.

### Check List 

Tests 

image


pkg/repository/v1_repository.go
```diff
@@ -611,6 +617,9 @@ func (r *V1Repository) BinaryPath(installPath string, componentID string, versio
 	if err != nil {
 		return "", err
 	}
+	if component == nil {
+		return "", errors.Errorf("component %s manifest not found, you can try `tiup list` to generate", componentID)
```
lonng, line 621: It's better to fetch component manifest directly instead of returning a error.

PR #304, opened 2020-05-31
Move the repository tiup-cluster into components/cluster

lonng:

This PR tries to merge the repository `https://github.com/pingcap-incubator/tiup-cluster` into `https://github.com/pingcap-incubator/tiup-cluster` as a component.


lonng:
The tiup-dm integration tests failed because of there is no v4.0.0 version in the repository.

PR #306, opened 2020-05-31
Support save commits for later sync

lucklove:

The server should save every upload file changes to an independent directory so that we can trace the changes through the time (for example: upload them to CDN)

In the Commit function of FsTxn, we add a sync stage which sync the temp directory to the commit directory.

### Check List 

Tests 

Code changes

Side effects

Related changes


server/store/sync.go
```diff
@@ -0,0 +1,59 @@
...
+package store
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path"
+	"time"
+
+	"github.com/pingcap-incubator/tiup/pkg/utils"
```
lonng, line 23: Please fix the imports.

server/handler/handler.go
```diff
@@ -16,22 +16,27 @@ package handler
 import (
 	"context"
 
+	"github.com/pingcap-incubator/tiup/pkg/logger/log"
```
lonng, line 19: Please fix the imports

lonng:
Seems the CI doesn't work properly.

PR #389, opened 2020-06-01
Add mirror clone prompt

lichunzhu:
some users misuse `tiup mirror clone` to set `[global-version]` as `<target-dir>`.

add mirror clone prompt to avoid misuse.

### DIFF
--- a/pkg/repository/clone_mirror.go
+++ b/pkg/repository/clone_mirror.go
@@ -44,6 +44,9 @@ type CloneOptions struct {
 
 // CloneMirror clones a local mirror from the remote repository
 func CloneMirror(repo *V1Repository, components []string, targetDir string, selectedVersions []string, options CloneOptions) error {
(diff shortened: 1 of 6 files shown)


pkg/repository/clone_mirror.go
```diff
@@ -44,6 +44,9 @@ type CloneOptions struct {
 
 // CloneMirror clones a local mirror from the remote repository
 func CloneMirror(repo *V1Repository, components []string, targetDir string, selectedVersions []string, options CloneOptions) error {
+	fmt.Printf("start to clone mirror, targetDir is %s, selectedVersions are %s\n", targetDir, selectedVersions)
```
lonng, line 47: Please use the upper case at the beginning of the log message and the `selectedVersions` is an array, you cannot use `%s` to format it.

lichunzhu: addressed

lichunzhu: addressed

PR #392, opened 2020-06-01
Skip tiup component when tiup update/install

Win-Man:
Skip update or install tiup when execute `tiup install/update`.

### Check List 

Tests 

### DIFF
--- a/pkg/environment/env.go
+++ b/pkg/environment/env.go
@@ -132,7 +132,7 @@ func (env *Environment) UpdateComponents(specs []string, nightly, force bool) er
 		var v1specs []repository.ComponentSpec
 		for _, spec := range specs {
 			component, v := ParseCompVersion(spec)
-			if spec == TiUPName {
+			if component == TiUPName {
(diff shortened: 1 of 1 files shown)


pkg/environment/env.go
```diff
@@ -127,6 +132,9 @@ func (env *Environment) UpdateComponents(specs []string, nightly, force bool) er
 		var v1specs []repository.ComponentSpec
 		for _, spec := range specs {
 			component, v := ParseCompVersion(spec)
+			if spec == TiUPName {
```
lonng, line 135:
```
if component ==
```

pkg/environment/env.go
```diff
@@ -138,6 +146,9 @@ func (env *Environment) UpdateComponents(specs []string, nightly, force bool) er
 	}
 	for _, spec := range specs {
 		component, v := ParseCompVersion(spec)
+		if spec == TiUPName {
```
lonng, line 149: ditto

PR #393, opened 2020-06-01
cluster: Support display Dashboard address only

breeswish:
Signed-off-by: Breezewish <hi@breeswish.org>

Support `tiup cluster display --dashboard`

Sample:

image

tiup-cluster

### Check List 

Tests 

 - Manual test


components/cluster/command/display.go
```diff
@@ -44,11 +48,22 @@ func newDisplayCmd() *cobra.Command {
...
+				return errors.Errorf("Cluster %s not found", clusterName)
 			}
-			if err := displayClusterTopology(clusterName, &gOpt); err != nil {
-				return err
+
+			if showDashboardOnly {
+				if err := displayDashboardInfo(clusterName); err != nil {
+					return err
+				}
+			} else {
```
AstroProfundis, line 60: I think we can return in `showDashboardOnly` and avoid the `else` block

breeswish: There is a tear down step `destroyTombstoneIfNeed`, is it ok to skip?

lucklove: In this case, I think it's ok to skip. This step is used to destroy tombstone state tikv,  it may take some time. It will be confusing to wait such a time for `--dashboard`

breeswish: Resolved, PTAL again, thanks!

breeswish:
@lonng @lucklove PTAL, thanks!

PR #398, opened 2020-06-02
Fix dashboard address

baurine:

Show dashboard node status even that node is abnormal.

企业微信20200602034753

Get the dashboard address from all PD nodes instead of the node self, because when the node is down, it can't return the response.

### Check List 

Tests 

   - Run a cluster with 3 PD nodes
   - Kill the PD node which runs the dashboard
   - Run `tiup cluster display clusterName` to see the nodes status

Side effects

 - Increased code complexity

### DIFF
--- a/pkg/cluster/api/pdapi.go
+++ b/pkg/cluster/api/pdapi.go
@@ -20,7 +20,6 @@ import (
 	"fmt"
 	"io"
 	"net/url"
-	"strings"
 	"time"
 
 	"github.com/pingcap/errors"
@@ -275,8 +274,7 @@ func (pc *PDClient) GetDashboardAddress() (string, error) {
 		return "", errors.AddStack(err)
 	}
 
-	r := strings.NewReplacer("http://", "", "https://", "")
-	return r.Replace(pdConfig.PDServerCfg.DashboardAddress), nil
+	return pdConfig.PDServerCfg.DashboardAddress, nil
 }
 
 // EvictPDLeader evicts the PD leader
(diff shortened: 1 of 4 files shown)


baurine:
@breeswish @HunDunDM PTAL, thanks!

pkg/cluster/meta/topology.go
```diff
@@ -278,40 +278,41 @@ type PDSpec struct {
...
-			suffix += "|UI"
+			suffix = "|L" + suffix
 		}
 		if member.Health {
 			return "Healthy" + suffix
 		}
 		return "Unhealthy" + suffix
 	}
-	return "N/A"
+	return "N/A" + suffix
```
baurine, line 315: Don't know when this kind of situation happens, I not sure whether the suffix is needed for "N/A".

pkg/cluster/api/pdapi.go
```diff
@@ -274,7 +275,8 @@ func (pc *PDClient) GetDashboardAddress() (string, error) {
 		return "", errors.AddStack(err)
 	}
 
-	return pdConfig.PDServerCfg.DashboardAddress, nil
+	r := strings.NewReplacer("http://", "", "https://", "")
+	return r.Replace(pdConfig.PDServerCfg.DashboardAddress), nil
```
baurine, line 279: This affects PR #393 , not sure whether this is acceptable, if doesn't, we need to move this part logic out.

breeswish: Yes, it affects #393 as well as our documentations. We'd better leave it unchanged, since it will be used to construct a full Dashboard access URL later.

baurine: Done.

PR #405, opened 2020-06-03
Remove version/help dependency on Environment

lichunzhu:
Current tiup will fail to print `tiup --version` nor `tiup xxx --help` while local data is not correct.

Move env init to `PersistentPreRunE`. We can run `tiup --version`, `tiup --help`, `tiup xxx --help` even when there is no network nor `root.json`. However, other commands like `tiup install`, help commands like `tiup help mirror` still can't be used.

Example(without local root.json):
image
image
image

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -58,11 +58,11 @@ the latest stable version will be downloaded from the repository.`,
 			if err != nil {
 				return err
 			}
-			environment.SetTiupEnv(e)
+			environment.SetGlobalEnv(e)
 			return nil
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
-			env := environment.TiupEnv()
+			env := environment.GlobalEnv()
 			if binary != "" {
 				component, ver := environment.ParseCompVersion(binary)
 				selectedVer, err := env.SelectInstalledVersion(component, ver)
@@ -96,7 +96,7 @@ the latest stable version will be downloaded from the repository.`,
 			return cmd.Help()
 		},
 		PersistentPostRunE: func(cmd *cobra.Command, args []string) error {
-			return environment.TiupEnv().Close()
+			return environment.GlobalEnv().Close()
 		},
 		SilenceUsage: true,
 	}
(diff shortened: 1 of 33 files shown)


cmd/root.go
```diff
@@ -122,17 +124,25 @@ the latest stable version will be downloaded from the repository.`,
...
 		cmd, n, e := cmd.Root().Find(args)
-		if (cmd == rootCmd || e != nil) && len(n) > 0 {
+		if (cmd == rootCmd || e != nil) && len(n) > 0 && err != nil {
 			externalHelp(env, n[0], n[1:]...)
 		} else {
 			cmd.InitDefaultHelpFlag() // make possible 'help' flag to be shown
 			_ = cmd.Help()
 		}
+		if err != nil {
+			fmt.Printf("\ninit local env meet error: %s\n", err)
```
lichunzhu, line 136: BTW, I want to delete this line. Users can get this error when they try to execute some commands.

lonng: I think it's ok to remove it.

pkg/environment/tiup.go
```diff
@@ -11,20 +11,16 @@
...
-	tiupmeta "github.com/pingcap/tiup/pkg/environment"
-)
+var _env *Environment
 
-var _env *tiupmeta.Environment
-
-// SetTiupEnv the gloable env used.
-func SetTiupEnv(env *tiupmeta.Environment) {
+// SetTiupEnv the global env used.
+func SetTiupEnv(env *Environment) {
```
lonng, line 19: How about changing this to `SetGlobalEnv`?

pkg/environment/tiup.go
```diff
@@ -11,20 +11,16 @@
...
 	_env = env
 }
 
-// TiupEnv Get the gloable env used.
-func TiupEnv() *tiupmeta.Environment {
+// TiupEnv Get the global env used.
+func TiupEnv() *Environment {
```
lonng, line 24: ditto, `GlobalEnv()`.

cmd/root.go
```diff
@@ -122,17 +124,23 @@ the latest stable version will be downloaded from the repository.`,
...
+		env, err := environment.InitEnv(repoOpts)
 		cmd, n, e := cmd.Root().Find(args)
-		if (cmd == rootCmd || e != nil) && len(n) > 0 {
+		if (cmd == rootCmd || e != nil) && len(n) > 0 && err != nil {
 			externalHelp(env, n[0], n[1:]...)
 		} else {
 			cmd.InitDefaultHelpFlag() // make possible 'help' flag to be shown
 			_ = cmd.Help()
 		}
+		if err == nil {
```
lonng, line 135: Do we need to set the global env? I think there will not be used in the subsequent.

PR #411, opened 2020-06-03
TiFlash: Update max-replicas settings before enable-placement-rules

birdstorm:

Settings of max-replicas may be overwritten by enable-placement-rules if it was set after enable-placement-rules is set to true.

Manually tested.

Move the setting of max-replicas before enable-placement-rules.

Also, add low-space-ratio check.

### Check List 

Tests 

 - Manual test

Code changes

Side effects

Related changes


components/playground/instance/tiflash.go
```diff
@@ -26,7 +26,7 @@ import (
 	"time"
 
 	"github.com/pingcap/errors"
-	"github.com/pingcap/tiup/components/playground/api"
+	"github.com/pingcap/tiup/pkg/cluster/api"
```
JaySon-Huang, line 29: So "playground" is somehow relied on "cluster"?

components/playground/instance/tiflash.go
```diff
@@ -96,7 +104,28 @@ func (inst *TiFlashInstance) Start(ctx context.Context, version v0manifest.Versi
 
 	// Wait for PD
 	pdClient := api.NewPDClient(endpoints, 10*time.Second, nil)
-	enablePlacementRules, err := json.Marshal(replicateConfig{
+	// set low-space-ratio to 1 to avoid low disk space
+	lowSpaceRatio, err := json.Marshal(scheduleConfig{
+		LowSpaceRatio: 1,
```
JaySon-Huang, line 109: In PD's [document](Maximum value: less than 1), it is "less than 1", but not "less than or equal 1".
Maybe set it to 0.99?

components/playground/instance/tiflash.go
```diff
@@ -96,7 +104,28 @@ func (inst *TiFlashInstance) Start(ctx context.Context, version v0manifest.Versi
...
+	})
+	if err != nil {
+		return nil
```
JaySon-Huang, line 112: return err ?

components/playground/instance/tiflash.go
```diff
@@ -96,7 +104,28 @@ func (inst *TiFlashInstance) Start(ctx context.Context, version v0manifest.Versi
...
+	}
+	if err = pdClient.UpdateScheduleConfig(bytes.NewBuffer(lowSpaceRatio)); err != nil {
+		return err
+	}
+	// Update maxReplicas before placement rules so that it would not be overwritten
+	maxReplicas, err := json.Marshal(replicateMaxReplicaConfig{
+		MaxReplicas: 1,
+	})
+	if err != nil {
+		return nil
```
JaySon-Huang, line 122: ditto

pkg/cluster/api/pdapi.go
```diff
@@ -537,23 +538,20 @@ func (pc *PDClient) DelPD(name string, retryOpt *clusterutil.RetryOption) error
 	return nil
 }
 
-// IsTombStone check if the node is Tombstone.
-// The host parameter should be in format of IP:Port, that matches store's address
-func (pc *PDClient) IsTombStone(host string) (bool, error) {
+func (pc *PDClient) isState(host string, state metapb.StoreState) (bool, error) {
```
JaySon-Huang, line 541: Maybe `isSameState` is better

components/playground/instance/tiflash.go
birdstorm: sure

components/playground/instance/tiflash.go
birdstorm: yes. i will fix it.

pkg/cluster/api/pdapi.go
birdstorm: ok

components/playground/instance/tiflash.go
birdstorm: the api of PDClient is copied from cluster. I think I will move them to a common package afterwards.

birdstorm: @JaySon-Huang It actually relies on `pkg`, `cluster` is a sub-directory of `pkg`, different from module `cluster`

bearHunting:
very effecient! Actually I found this problem and report to authority. Glad to help!

PR #424, opened 2020-06-04
Publish: use mirror address as default endpoint

lucklove:

The command `tiup mirror publish` need a flag --repo to specific the remote tiup-server. However, in real use the remote server address will be the mirror address, so we make them same in flag default options

### DIFF
--- a/pkg/repository/remote/upload.go
+++ b/pkg/repository/remote/upload.go
@@ -58,7 +58,7 @@ type transporter struct {
 // New returns a Transporter
 func New(endpoint, component, version, entry string) Transporter {
 	return &transporter{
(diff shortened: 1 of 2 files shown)


pkg/repository/remote/upload.go
```diff
@@ -57,6 +57,10 @@ type transporter struct {
 
 // New returns a Transporter
 func New(endpoint, component, version, entry string) Transporter {
+	if strings.HasSuffix(endpoint, "/") {
```
AstroProfundis, line 60: why not `strings.TrimSuffix()`

lucklove: Ok, that's better

PR #431, opened 2020-06-05
Use file service discover for prometheus in playground

### LINKED ISSUE #427: use file_sd_config instead of static_configs of Prometheus in playground
This can avoid restart need when scaling instance.

ref:
https://prometheus.io/docs/prometheus/latest/configuration/configuration/#file_sd_config

july2993:
fix #427 

ref: https://prometheus.io/docs/prometheus/latest/configuration/configuration/#file_sd_config
use file service discovers to avoid restart Prometheus. 

after `tiup playground --monitor`
Screen Shot 2020-06-05 at 1 47 52 PM

after `tiup playground scale-out --db 1
Screen Shot 2020-06-05 at 1 48 13 PM

after `tiup playground scale-in --pid <pid of tidb>
Screen Shot 2020-06-05 at 1 49 26 PM


components/playground/monitor.go
```diff
@@ -0,0 +1,125 @@
+package main
```
lonng, line 1: The License header is missing. And I think it's better to move to `instance`.

july2993: add by https://github.com/pingcap/tiup/pull/431/commits/b8901633a3b61746d70c003b3e8ceeb51ad9b218

PR #446, opened 2020-06-08
utils: fix directory removing

AstroProfundis:
After importing, it raises a `directory not empty` error when trying to move the inventory dir to a new location.

### DIFF
--- a/pkg/utils/ioutil.go
+++ b/pkg/utils/ioutil.go
@@ -140,5 +140,5 @@ func Move(src, dst string) error {
 	if err := Copy(src, dst); err != nil {
 		return errors.Trace(err)
 	}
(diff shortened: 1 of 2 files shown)


pkg/utils/ioutil.go
```diff
@@ -140,5 +140,5 @@ func Move(src, dst string) error {
 	if err := Copy(src, dst); err != nil {
```
lonng, line 140: Can the `copy` apply to the directory?

AstroProfundis: Yes, that was introduced in #410

lucklove: It can apply to directories now

PR #447, opened 2020-06-08
display: update state displaying for offline tikv node

AstroProfundis:
The raw store state returned from TiKV is misleading, the "offline" is a state that the store is being transferred to other node and "tombstone" is when the offline is really complete.

There is risk that user misunderstand the meaning of "Offline" and use `--force` arg to scale in other instances in the cluster, which may lead to data loss.

Set the displayed state to "Pending Offline" for better understanding.

### DIFF
--- a/pkg/cluster/meta/topology.go
+++ b/pkg/cluster/meta/topology.go
@@ -226,7 +226,11 @@ func checkStoreStatus(storeAddr string, pdList ...string) string {
 		}
 	}
 	if latestStore != nil {
-		return latestStore.Store.StateName
+		state := latestStore.Store.StateName
+		if strings.ToLower(state) == "offline" {
+			state = "Pending Offline" // avoid misleading
+		}
+		return state
 	}
 	return "N/A"
 }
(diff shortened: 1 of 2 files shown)


pkg/cluster/meta/topology.go
```diff
@@ -226,7 +226,11 @@ func checkStoreStatus(storeAddr string, pdList ...string) string {
 		}
 	}
 	if latestStore != nil {
-		return latestStore.Store.StateName
+		state := latestStore.Store.StateName
+		if strings.ToLower(state) == "offline" {
+			state = "Pending Offline" // avoid misleading
```
lonng, line 231: I don't think this is the proper way. I think we add `Pending`  suffix just when both `spec.Offline` and `status==offline` are true. It's better to change the status display string at  https://github.com/pingcap/tiup/pull/447/files#diff-dc4c66f2c87a0397b0ad841b79378f17R217

pkg/cluster/meta/topology.go
```diff
@@ -234,7 +234,11 @@ func checkStoreStatus(storeAddr string, pdList ...string) string {
 // Status queries current status of the instance
 func (s TiKVSpec) Status(pdList ...string) string {
```
lonng, line 235: The same change should apply to the TiFlash.

PR #449, opened 2020-06-08
don't generate config file of prometheus if we don't start it

july2993:
fail to start without monitor cause monitor is nil.

introduced by https://github.com/pingcap/tiup/pull/431

but we have not released any version having this issue.


july2993:
@lonng the CI are expected to found this issue?

PR #450, opened 2020-06-08
executor/ssh: force using C locale for commands

AstroProfundis:
We use string matching to check for command outputs at several places, and that might fail if the user's deploy server is not using an English locale.

Force all SSH commands to be executed with `LANG=C` environment variable set.

### DIFF
--- a/pkg/cluster/executor/ssh.go
+++ b/pkg/cluster/executor/ssh.go
@@ -129,7 +129,7 @@ func (e *SSHExecutor) Execute(cmd string, sudo bool, timeout ...time.Duration) (
 	}
 
 	if e.Locale != "" {
-		cmd = fmt.Sprintf("LANG=%s %s", e.Locale, cmd)
+		cmd = fmt.Sprintf("export LANG=%s; %s", e.Locale, cmd)
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/cluster/executor/ssh.go
```diff
@@ -126,6 +128,10 @@ func (e *SSHExecutor) Execute(cmd string, sudo bool, timeout ...time.Duration) (
 		cmd = fmt.Sprintf("sudo -H -u root bash -c \"%s\"", cmd)
 	}
 
+	if e.Locale != "" {
+		cmd = fmt.Sprintf("export LANG=%s; %s", e.Locale, cmd)
```
lonng, line 132: Have you ever tested in a non-English environment?

AstroProfundis: Yes I tested with `LANG=zh_CN.UTF-8` with bash and sh, other shell may behaviour differently (if they don't support the `export` syntax, i.e., fish).

We need to export the env first, otherwise the env is only working for new process forked by the shell, but not the running shell itself, that means messages like "command not found" is not affected by the setting.

PR #457, opened 2020-06-09
Remove args in Command

july2993:
remove args in Command.
the default will be the command names join by space, and the specified command run can append more args if need.

currently, all cluster name is hash, and the version is appended raw.
example: "command":"cluster start cluster_0cc175b9c0f1b6a831c399e269772661"

### DIFF
--- a/components/cluster/command/root.go
+++ b/components/cluster/command/root.go
@@ -49,12 +49,21 @@ var (
 	skipConfirm bool
 )
 
+func scrubClusterName(n string) string {
+	return "cluster_" + telemetry.HashReport(n)
+}
+
 func getParentNames(cmd *cobra.Command) []string {
 	if cmd == nil {
 		return nil
 	}
 
 	p := cmd.Parent()
+	// always use 'cluster' as the root command name
+	if cmd.Parent() == nil {
+		return []string{"cluster"}
+	}
+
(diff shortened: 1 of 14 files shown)


components/cluster/command/root.go
```diff
@@ -49,12 +49,21 @@ var (
...
+
 func getParentNames(cmd *cobra.Command) []string {
 	if cmd == nil {
 		return nil
 	}
 
 	p := cmd.Parent()
+	// always use 'cluster' as the root command name
+	if cmd.Parent() == nil {
+		return []string{"cluster"}
```
lonng, line 64: It's better to use `cliutil.OsArgs0()` if the `cmd.Parent() == nil`.

july2993: https://github.com/pingcap/tiup/blob/cb5bcd4fadb1521fbad4df5d1032b0e35fd355f9/pkg/cliutil/cliutil.go#L45
then we will always get `tiup`  if run by tiup.  

I think what you want is os.Args[0], the program name of cluster. keep literal "cluster" here should be ok here.

july2993: correct： `cliutil.OsArgs0()` is "tiup cluster" but not just "tiup", but the name of root cmd is "tiup"

PR #458, opened 2020-06-09
Fix a dead link in README.md

sticnarf:

The link in `README.md` pointing to the dev doc is not correct.

Corrects the link.

### Check List 

Tests 

 - No code


sticnarf:
PTAL @lucklove

Issue #472, opened 2020-06-09
PD nodes status are not stable if some nodes are down


baurine:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?

I deploy a cluster with 3 PD nodes, 1 TiDB node, 1 TiKV node by vagrant, then I pause a PD node.

![企业微信20200609114349](https://user-images.githubusercontent.com/1284531/84126265-43df7b00-aa70-11ea-91cb-cad864fad602.png)

2. What did you expect to see?

I expect to see PD nodes status with 2 healthy and 1 unhealthy. 

3. What did you see instead?

The PD nodes statuses are not stable with different output each time after running `tiup cluster display` command, while they are normal in the tidb-dashboard and the response of API `/pd/health`.

![企业微信20200609114307](https://user-images.githubusercontent.com/1284531/84126321-55c11e00-aa70-11ea-96ab-8b309faffec3.png)

![企业微信20200609113905](https://user-images.githubusercontent.com/1284531/84126357-61ace000-aa70-11ea-816a-ea8de0858cf7.png)

![企业微信20200609113929](https://user-images.githubusercontent.com/1284531/84126381-696c8480-aa70-11ea-8985-527ae7fbfb3a.png)

4. What version of TiUP are you using (`tiup --version`)?

v0.0.5 tiup with v1.0.3 tiup cluster

baurine:
I will submit PR to resolve it.

PR #473, opened 2020-06-09
enlarge the pd api timeout time to fix the pd nodes statuses are not stable

baurine:

resolve a part of #472 

Enlarge the PD API query timeout duration from 2 seconds to 10 seconds, because when there are some PD nodes are down, the `/pd/health` API will usually cost more than 2 seconds.

Before:

企业微信20200609114307

After:

image

### Check List 

Tests 


baurine:
@lucklove PTAL, thanks!

PR #474, opened 2020-06-09
Simplify PD node status

baurine:

Resolve a part of #472 

Currently, the "Unhealthy" PD node is recognized as "Down" status, and the "Unhealthy" status never displays.

Differ the "Down" and "Unhealthy" status for PD node, when all nodes don't work, recognize them as "Down", else recognize the not work node as "Unhealthy".

### Check List 

Tests 

Deploy a cluster with 3 PD nodes, pause 1 node.

企业微信20200609114349

Before:

企业微信20200609114229

After:

image

Pause 2 PD nodes:

image

### DIFF
--- a/pkg/cluster/meta/topology.go
+++ b/pkg/cluster/meta/topology.go
@@ -279,6 +279,7 @@ type PDSpec struct {
 // Status queries current status of the instance
 func (s PDSpec) Status(pdList ...string) string {
 	curAddr := fmt.Sprintf("%s:%d", s.Host, s.ClientPort)
+	curPdAPI := api.NewPDClient([]string{curAddr}, statusQueryTimeout, nil)
 	allPdAPI := api.NewPDClient(pdList, statusQueryTimeout, nil)
 	suffix := ""
 
@@ -292,15 +293,13 @@ func (s PDSpec) Status(pdList ...string) string {
 		suffix = "|UI"
 	}
 
-	// "Down" means all PD nodes don't work,
-	// while "Unhealthy" means current PD node doesn't work
-	healths, err := allPdAPI.GetHealth()
+	healths, err := curPdAPI.GetHealth()
 	if err != nil {
 		return "Down" + suffix
 	}
 
 	// find leader node
-	leader, err := allPdAPI.GetLeader()
+	leader, err := curPdAPI.GetLeader()
 	if err != nil {
 		return "ERR" + suffix
 	}
--- a/pkg/cluster/meta/topology.go
+++ b/pkg/cluster/meta/topology.go
@@ -293,7 +293,8 @@ func (s PDSpec) Status(pdList ...string) string {
 		suffix = "|UI"
 	}
 
-	healths, err := curPdAPI.GetHealth()
+	// check health
+	err := curPdAPI.CheckHealth()
 	if err != nil {
 		return "Down" + suffix
 	}
@@ -303,20 +304,10 @@ func (s PDSpec) Status(pdList ...string) string {
 	if err != nil {
 		return "ERR" + suffix
 	}
-
-	for _, member := range healths.Healths {
-		if s.Name != member.Name {
-			continue
-		}
-		if s.Name == leader.Name {
-			suffix = "|L" + suffix
-		}
-		if member.Health {
-			return "Healthy" + suffix
-		}
-		return "Unhealthy" + suffix
+	if s.Name == leader.Name {
+		suffix = "|L" + suffix
 	}
-	return "N/A" + suffix
+	return "Up" + suffix
 }
 
 // Role returns the component role of the instance
--- a/pkg/cluster/meta/topology.go
+++ b/pkg/cluster/meta/topology.go
@@ -279,7 +279,6 @@ type PDSpec struct {
 // Status queries current status of the instance
 func (s PDSpec) Status(pdList ...string) string {
 	curAddr := fmt.Sprintf("%s:%d", s.Host, s.ClientPort)
-	curPdAPI := api.NewPDClient([]string{curAddr}, statusQueryTimeout, nil)
 	allPdAPI := api.NewPDClient(pdList, statusQueryTimeout, nil)
 	suffix := ""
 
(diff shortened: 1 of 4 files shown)


baurine:
@lucklove @HunDunDM please help take a look as well, thanks!

breeswish:
From the user perspective do we have differences about unhealthy and down? @HunDunDM  PTAL

pkg/cluster/meta/topology.go
```diff
@@ -293,13 +292,15 @@ func (s PDSpec) Status(pdList ...string) string {
 		suffix = "|UI"
 	}
 
-	healths, err := curPdAPI.GetHealth()
+	// "Down" means all PD nodes don't work,
+	// while "Unhealthy" means current PD node doesn't work
```
AstroProfundis, line 296: I'm confused by this, I thought "Down" means the specific PD node is not working, and "Unhealthy" means the PD cluster is not in proper state.

disksing:
I don't think we have made a clear distinction between these two different terms.
IMO it is a good idea to distinguish between two states: single node down / majority nodes down. However, the state names would be better if they are more self-explanatory.

pkg/cluster/meta/topology.go
baurine: But according to the old logic, they will be all recognized as "Down" status, we never can go to the `return "Unhealthy" + suffix` line logic.

baurine:
In fact, my key point is that, according to the code, we use the "Healthy" status to represent the work PD node, so it is natural to think that we should use "Unhealthy" to represent the not work PD node. It is true the "Unhealthy" is used in the code, but unfortunately, its logic never reach, it is always recognized "Down".

(I guess the "Healthy" and "Unhealthy" come from the `/pd/health` API response. In the reponse we use `health: true/false` to represent whether a node is working.)

So, the PD node has 3 kinds of status: "Healthy", "Unhealthy" and "Down", but only "Healthy" and "Down" happen.

Actually, for other components (TiDB / TiKV / TiFlash), we use "Up" to represent work node, and use "Down" to represent not work node.

So it seems here exists some inconsistency. Now I don't think this PR is a good solution. Or maybe we can use "Up" to replace "Healthy", remove the unreached "Unhealthy", and keep "Down" for PD node to keep consistency with other components, how about it?

disksing:
Sounds reasonable to me. Any thought? /cc @nolouch @HunDunDM

AstroProfundis:
It was because the **is** a state in PD's API response, it was never clearly documented what those states represent. It's ok to just remove the health/unhealthy term in response body and use up/down for only check response http code.

But I would wonder: what on earth does the `health` state in PD's response mean?

AstroProfundis:
What state would a PD node report, if another node in the cluster is down, while itself is still working, health = true or false ?

HunDunDM:
We determine that the health of an instance is based on that the http server of this node can respond normally.

HunDunDM:
> In fact, my key point is that, according to the code, we use the "Healthy" status to represent the work PD node, so it is natural to think that we should use "Unhealthy" to represent the not work PD node. It is true the "Unhealthy" is used in the code, but unfortunately, its logic never reach, it is always recognized as "Down".
> 
> (I guess the "Healthy" and "Unhealthy" come from the `/pd/health` API response. In the response we use `health: true/false` to represent whether a node is working.)
> 
> So, the PD node has 3 main kinds of status: "Healthy", "Unhealthy" and "Down", but only "Healthy" and "Down" happen.
> 
> Actually, for other components (TiDB / TiKV / TiFlash), we use "Up" to represent work node, and use "Down" to represent not work node.
> 
> So it seems here exists some inconsistency. Now I don't think this PR is a good solution. Or maybe we can use "Up" to replace "Healthy", remove the unreached "Unhealthy", and keep "Down" for PD node to keep consistency with other components, how about it?

Sounds reasonable to me too.

AstroProfundis:
> We determine that the health of an instance is based on that the http server can respond normally.

Thanks for the clarification, my curiosity is that when would a PD node be responding to API calls and return an object with `health: false`; and what would other PD node report in this filed, when a node is down in the cluster.

HunDunDM:
When you request `/health` for any pd:
* If it is a follower node, this node will internally proxy the request to the leader node.
* If it is a leader node, this node will try to request `/ping` of all pds (including itself), and mark the node as healthy if it can respond normally. Then return the result.

AstroProfundis:
>     * If it is a leader node, this node will try to request `/ping` of all pds (including itself), and mark the node as healthy if it can respond normally. Then return the result.

What if a follower is down, will the leader respond unhealthy state? And if I understand it correctly, in this circumstance, all alive nodes will report the same unhealthy state?

HunDunDM:
The leader will only mark the node as unhealthy. 

> all alive nodes will report the same unhealthy state

yes.

baurine:
@AstroProfundis , this is an example response for `/pd/health` API which requested from `http://10.0.1.11:2379`:

```json
[
  {
    "name": "pd-10.0.1.14-2379",
    "member_id": 4436600309396281139,
    "client_urls": [
      "http://10.0.1.14:2379"
    ],
    "health": true
  },
  {
    "name": "pd-10.0.1.11-2379",
    "member_id": 16310129336451408704,
    "client_urls": [
      "http://10.0.1.11:2379"
    ],
    "health": true
  },
  {
    "name": "pd-10.0.1.15-2379",
    "member_id": 16570216285418057773,
    "client_urls": [
      "http://10.0.1.15:2379"
    ],
    "health": false
  }
]
```

AstroProfundis:
OK I understand. So there are 2 possible solutions:

1. Keep the current healthy/unhealthy state in displaying, but request that from any alive PD instead of only the node itself (we have similar approach when finding PD leader in the code), if the node with problem is not found in the first success response, it's marked as down
2. Request the PD API of the specific node, ignore the state of PD node in response, use an up/down state based on the http response code of `/pd/health`, not parsing the content at all

I prefer 2.

baurine:
Prefer 2 as well.

baurine:
Hi @AstroProfundis , I have updated it, PTAL, thanks!

btw, I used the `/pd/ping` API replace `/pd/health` to check the single node health status according to @HunDunDM recommendation.

besides, I found there are some similar code in `/components/dms/command/display.go` with the code in `/components/cluster/command/display.go`, should I change that as well?

lonng:
ping @AstroProfundis

PR #482, opened 2020-06-10
Fix the bug that generate wrong snapshot meta when tiup mirror clone

lonng:

Fix: #483 

Fix the bug that generate wrong snapshot meta when tiup mirror clone.


pkg/repository/clone_mirror.go
```diff
@@ -171,7 +171,7 @@ func CloneMirror(repo *V1Repository, components []string, targetDir string, sele
 			return err
 		}
 		var _ = len(bytes) // this length is the not final length, since we still change the manifests before write it to disk.
-		snapshot.Meta["/"+name] = v1manifest.FileVersion{
+		snapshot.Meta["/"+fname] = v1manifest.FileVersion{
```
AstroProfundis, line 174: Why did it working if the former lengths in manifests are wrong?

lonng:
> One more issue, there is `v1manifest.RenewManifest()` that reset the expires time of a manifest, so no need to calculate expire time manually with `SetExpiresAt`

@AstroProfundis The expiration of `v1manifest.RenewManifest()` is too short for the offline mirror, so I set the expiration manually.

PR #489, opened 2020-06-11
move init mirror to preRun for `tiup mirror clone`

lichunzhu:
move init mirror to preRun for `tiup mirror clone`. tiup --help and --version commands (expect `tiup mirror clone --help`) don't need init environment and connect to network any more.

Don't parse args for `tiup mirror clone` at beginning. Do it in running progress.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/cmd/mirror.go
+++ b/cmd/mirror.go
@@ -471,11 +471,34 @@ func yankComp(repo, id, version string) error {
 // the `mirror clone` sub command
 func newMirrorCloneCmd() *cobra.Command {
 	var (
-		options    = repository.CloneOptions{Components: map[string]*[]string{}}
-		components []string
-		repo       *repository.V1Repository
+		options     = repository.CloneOptions{Components: map[string]*[]string{}}
+		components  []string
+		repo        *repository.V1Repository
+		initialized bool
 	)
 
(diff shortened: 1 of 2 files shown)


cmd/mirror.go
```diff
@@ -470,50 +470,60 @@ func yankComp(repo, id, version string) error {
...
+				options.Components[name] = new([]string)
+				cmd.Flags().StringSliceVar(options.Components[name], name, nil, "Specify the versions for component "+name)
+			}
+			return nil
+		},
 		RunE: func(cmd *cobra.Command, args []string) error {
-			if len(args) < 1 {
-				return cmd.Help()
+			cmd.DisableFlagParsing = false
+			err := cmd.ParseFlags(args)
```
lonng, line 511: It's better to use `SetHelpFunc` to overwrite the help function. The current implementation will cause inconsistent between `tiup help mirror clone` and `tiup mirror clone -h`.

lichunzhu: addressed, PTAL

PR #490, opened 2020-06-11
Rewrite dms specification

lichunzhu:
rewrite dms specification and block dms commands to pass `make build`.

### Check List 

Tests 

 - Unit test

Code changes


lichunzhu:
> Prefer to use `/**/` to comment code block to reduce the diff size.

Yes, I will refine this.

lichunzhu:
@lonng Revised, PTAL again

pkg/dms/meta/paths.go
```diff
@@ -0,0 +1,37 @@
...
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package meta
```
lonng, line 14: It's better to reuse the existing `DirPaths` structure instead of re-defining a new one. You can move the exists one into a common package.

pkg/dms/meta/server_config.go
```diff
@@ -0,0 +1,53 @@
...
+	"strings"
+
+	"github.com/pingcap/tiup/pkg/cluster/clusterutil"
+	"github.com/pingcap/tiup/pkg/cluster/executor"
+	"github.com/pingcap/tiup/pkg/cluster/meta"
+
+	"github.com/pingcap/errors"
+)
+
+func checkConfig(e executor.TiOpsExecutor, componentName, clusterVersion, nodeOS, arch, config string, paths DirPaths) error {
```
lonng, line 28: ditto

pkg/dms/meta/paths.go
lichunzhu: ok

pkg/dms/meta/topology.go
```diff
@@ -18,48 +18,209 @@ import (
...
+type JobSpec struct {
+	Action  string                 `yaml:"action"`
+	Type    string                 `yaml:"type"`
+	Name    string                 `yaml:"name,omitempty"`
+	Config  map[string]interface{} `yaml:"config,omitempty"`
+	Sources []SourceSpec           `yaml:"sources"`
+	Sink    SourceSpec             `yaml:"sink"`
+	Workers []WorkerSpec           `yaml:"workers"`
+	// job related specs, once job is stopped, these specs should stopped
+	DMMasters    []DMMasterSpec     `yaml:"dm-masters,omitempty"`
```
lonng, line 84: Do we need the `dm-master` field?

lonng:
@lichunzhu 
There is a lot of duplicated code in the tiup codebase, it's better to reuse the existing code instead of copy-paste.

pkg/dms/meta/topology.go
lichunzhu: I distinguish specs into two types:
1. started with source, ended with source:
    For example, Dumpling, Lightning, Importer, DMWorker. When their related source is removed from the job, these components should be stopped.
2. started with job, ended with job:
    For example, DMMaster, Monitors. When one source is removed from the job, these components shouldn't be affected. But when the job is stopped, these components should be stopped.

For `dm-master`, I think one or two sources shouldn't affect `dm-master`, so I put it here.

PR #495, opened 2020-06-11
add listen_host for TiDB/TiKV/PD to specified the listen host

crazycs520:
Signed-off-by: crazycs520 <crazycs520@gmail.com>

Before this PR, TiDB/TiKV will always listen to the `0.0.0.0` host, this PR adds `listen_host` flag for PD/TiDB/TiKV to specify the listening address.

If user doesn't specified the `listen_host`, PD/TiDB/TiKV will listen on `0.0.0.0` defaultly.

### Check List 

Tests 

 - Manual test


pkg/cluster/meta/logic.go
```diff
@@ -248,6 +249,14 @@ func (i *instance) GetHost() string {
 	return i.host
 }
 
+// GetListenAddress implements Instance interface
+func (i *instance) GetListenAddress() string {
+	if len(i.listenAddress) == 0 {
```
AstroProfundis, line 254: why not compare with `""`?

pkg/cluster/meta/topology.go
```diff
@@ -127,6 +127,7 @@ func AllComponentNames() (roles []string) {
 // TiDBSpec represents the TiDB topology specification in topology.yaml
 type TiDBSpec struct {
 	Host            string                 `yaml:"host"`
+	ListenAddress   string                 `yaml:"listen_address"`
```
AstroProfundis, line 130: I think it's better to set `omitempty` for this

pkg/cluster/template/scripts/tidb.go
```diff
@@ -24,23 +24,25 @@ import (
...
+	StatusPort    int
+	DeployDir     string
+	LogDir        string
+	NumaNode      string
+	Endpoints     []*PDScript
 }
 
 // NewTiDBScript returns a TiDBScript with given arguments
-func NewTiDBScript(ip, deployDir, logDir string) *TiDBScript {
+func NewTiDBScript(ip, listenAddress, deployDir, logDir string) *TiDBScript {
```
AstroProfundis, line 38: How about using a similar approach with `Port`, that could avoid long argument list of the function. It's not a must, I just happen to recognized that.

pkg/cluster/meta/topology.go
crazycs520: great, done.

pkg/cluster/template/scripts/tidb.go
crazycs520: great, done.

pkg/cluster/meta/topology.go
```diff
@@ -127,6 +127,7 @@ func AllComponentNames() (roles []string) {
 // TiDBSpec represents the TiDB topology specification in topology.yaml
 type TiDBSpec struct {
 	Host            string                 `yaml:"host"`
+	ListenAddress   string                 `yaml:"listen_address,omitempty"`
```
lonng, line 130: I think `listen_address` is a bad name.

crazycs520: em... Any suggestion?

breeswish: Maybe call it `listen_host`? (since Address should contain the port part)

crazycs520: great,

crazycs520: done.

lonng:
@crazycs520 Please update the documentation/examples for `listen_host` field, thanks.

templates/scripts/run_tidb.sh.tpl
```diff
@@ -24,6 +24,7 @@ exec env GODEBUG=madvdontneed=1 bin/tidb-server \
 {{- end}}
     -P {{.Port}} \
     --status="{{.StatusPort}}" \
+    --host="{{.ListenAddress}}" \
```
buptzhoutian: Please add `--status-host="{{.ListenAddress}}"` , too

lonng: @crazycs520

PR #496, opened 2020-06-12
Support pump&drainer in playground

### LINKED ISSUE #422: support pump & drainer in playground

july2993:
fix #422 

this pr also fix the issue when --host is set as "0.0.0.0".
according to the doc:https://pingcap.com/docs-cn/dev/tiup/tiup-playground/
```
      --host string              设置每个组件的监听地址（默认为 127.0.0.1），如果要提供给别的电脑访问，可设置为 0.0.0.0
```

this actually can't work cause many advertise addr also use "0.0.0.0:<port>"

Note the default config of drainer is replicate to 127.0.0.1:3306, user must set up the downstream itself or provide the custom config for drainer, or drainer will fail to run.

usage example:
run MySQL local at port 3306.

start a playground with:
- tp --pump 3 --drainer 1

check the replication works

in another terminal
- scale-in one pump
- scale-in one drainer
- scale-out one pump
- scale-out one drainer
Screen Shot 2020-06-12 at 4 04 24 PM

### DIFF
--- a/components/playground/instance/drainer.go
+++ b/components/playground/instance/drainer.go
@@ -1,3 +1,16 @@
+// Copyright 2020 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package instance
 
 import (
--- a/components/playground/instance/drainer.go
+++ b/components/playground/instance/drainer.go
@@ -1,0 +1,93 @@
+package instance
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+
+	"github.com/pingcap/tiup/pkg/localdata"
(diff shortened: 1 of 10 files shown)


components/playground/instance/drainer.go
```diff
@@ -0,0 +1,93 @@
+package instance
```
lonng, line 1: The license header is missing.

components/playground/instance/drainer.go
```diff
@@ -0,0 +1,93 @@
...
+	if d.ConfigPath != "" {
+		args = append(args, fmt.Sprintf("--config=%s", d.ConfigPath))
+	}
+
+	d.cmd = exec.CommandContext(ctx, args[0], args[1:]...)
+	d.cmd.Env = append(
+		os.Environ(),
+		fmt.Sprintf("%s=%s", localdata.EnvNameInstanceDataDir, d.Dir),
+	)
+	d.cmd.Stderr = os.Stderr
```
lonng, line 79: Should we redirect the stdout/stderr into log file?

components/playground/instance/pump.go
```diff
@@ -0,0 +1,114 @@
+package instance
```
lonng, line 1: ditto

components/playground/instance/drainer.go
july2993: I am thinking of how about make all instance's log into the log file in the instance' directory.
because it's hard to tell what will the instance print out in the console.

then the console will only print out what we print int the playground.

PR #497, opened 2020-06-12
Implement --yank --standalone and --hide in tiup mirror publish|modify

lucklove:

This PR implement `--standalone` `--hide` flag on `tiup mirror publish` and add `tiup mirror modify` subcommand.

The `tiup mirror modify` is used to modify component attrs (hidden, yanked, standalone)


cmd/mirror.go
```diff
@@ -255,6 +337,8 @@ func newMirrorPublishCmd() *cobra.Command {
 	cmd.Flags().StringVarP(&goarch, "arch", "", goarch, "the target system architecture")
 	cmd.Flags().StringVarP(&desc, "desc", "", desc, "description of the component")
 	cmd.Flags().StringVarP(&endpoint, "endpoint", "", endpoint, "endpoint of the server")
+	cmd.Flags().BoolVarP(&standalone, "standalone", "", standalone, "can this component run directly")
+	cmd.Flags().BoolVarP(&hidden, "hide", "", hidden, "is this component visible in list")
```
AstroProfundis, line 341: the help should better be "is this component **invisible** on listing"

cmd/list.go
```diff
@@ -117,11 +117,23 @@ func showComponentList(env *environment.Environment, opt listOptions) (*listResu
...
 
 	localComponents := set.NewStringSet(installed...)
-	for id, comp := range index.Components {
+	compIDs := []string{}
+	for id := range index.Components {
+		compIDs = append(compIDs, id)
+	}
+	sort.Slice(compIDs, func(i, j int) bool {
+		return compIDs[i] < compIDs[j]
+	})
```
lonng, line 126: `sort.String()`

cmd/mirror.go
```diff
@@ -190,13 +193,81 @@ func delComp(repo, id, version string) error {
...
+			if err != nil {
+				return err
+			}
+
+			e := remote.NewEditor(endpoint, args[0]).WithDesc(desc)
+			flagSet := set.NewStringSet()
+			cmd.Flags().Visit(func(f *pflag.Flag) {
+				flagSet.Insert(f.Name)
+			})
+			if flagSet.Exist("standalone") {
```
lonng, line 239: Why not check the bool variable?

pkg/repository/remote/common.go
```diff
@@ -0,0 +1,73 @@
+package remote
```
lonng, line 1: License header is missing.

pkg/repository/remote/modify.go
```diff
@@ -0,0 +1,72 @@
+package remote
```
lonng, line 1: ditto

cmd/mirror.go
lucklove: If a bool value (eg. the --hide flag), is false, we don't know if it's `tiup mirror modify xxx --hide=false` or `tiup mirror modify xxx`. These are two very different semantics, the first one is that we should unhide the component, the second one is that we should do nothing on the hidden field

cmd/mirror.go
```diff
@@ -223,7 +294,26 @@ func newMirrorPublishCmd() *cobra.Command {
 				return err
 			}
 
-			t := remote.New(endpoint, args[0], args[1], args[3]).WithDesc(desc).WithOS(goos).WithArch(goarch)
+			flagSet := set.NewStringSet()
+			cmd.Flags().Visit(func(f *pflag.Flag) {
+				flagSet.Insert(f.Name)
+			})
+			m, err := env.V1Repository().FetchComponentManifest(args[0])
+			if err != nil {
```
lonng, line 302: Do we need to check the error type, e.g: NotFound?

server/handler/component.go
```diff
@@ -46,6 +46,17 @@ func (h *componentSigner) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 	sid := mux.Vars(r)["sid"]
 	name := mux.Vars(r)["name"]
+	options := make(map[string]bool)
+
+	for _, opt := range []string{"yanked", "standalone", "hidden"} {
+		if query(r, opt) == "true" {
+			options[opt] = true
+		} else if query(r, opt) == "false" {
+			options[opt] = false
+		}
```
lonng, line 56: The +52 to +56 can be simplified to `options[opt] = query(r, opt) == "true"`

server/handler/component.go
```diff
@@ -46,6 +46,17 @@ func (h *componentSigner) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	fmt.Println(options)
```
lonng, line 59: All output of tiup-server should print to the logger.

server/handler/component.go
```diff
@@ -80,20 +97,58 @@ func (h *componentSigner) sign(r *http.Request, m *model.ComponentManifest) (sr
...
+					URL:        fmt.Sprintf("/%s.json", name),
+					Yanked:     options["yanked"],
+					Standalone: options["standalone"],
+					Hidden:     options["hidden"],
+				}
 			}
-			keys = om.Signed.Owners["pingcap"].Keys
+
+			om.Signed.Components[name] = compItem
 			indexVersion = om.Signed.Version + 1
```
lonng, line 145: The `indexVersion = om.Signed.Version + 1` only be applied if `compExist = false`. It's better to move the L145 to  above L141.

lucklove: We must update component.json when user try to modify that component. Otherwise, people can tamper others' components because they don't need to sign to modify the attributes(hidden, standalone, yanked)

cmd/mirror.go
lucklove: The best way is do so. But it's a little hard to do this in this stage(we didn't define error well yet). So this is a tradeoff.

PR #500, opened 2020-06-15
Reorder imported config priority

lucklove:
The tiup config should always have higher priority
than tiup-ansible ones.

This PR gives tiup config higher priority than imported tidb-ansible config to reduce
the confusion.

I changed the merge methods and let it accept more than one overwrites, it will overwrite the orig with the overwrites one by one, so the latter overwrites have higher priority.

### Check List 

Tests 

 - Unit test


lonng:
It's better to add more tests and attach the manual test result in the PR description.

PR #501, opened 2020-06-15
Fix timeout when target directory exists and there is lots of content

lucklove:

When the user specifies a directory which contains lots of content (200GB for example) as log/data directory, the chown -R command will wait a long time to timeout.

`chown -R` -> `chown`

### DIFF
--- a/pkg/cluster/clusterutil/clusterutil.go
+++ b/pkg/cluster/clusterutil/clusterutil.go
@@ -30,6 +30,10 @@ func Abs(user, path string) string {
 func MultiDirAbs(user, paths string) []string {
 	var dirs []string
 	for _, path := range strings.Split(paths, ",") {
+		path = strings.TrimSpace(path)
+		if path == "" {
+			continue
+		}
 		dirs = append(dirs, Abs(user, path))
 	}
 	return dirs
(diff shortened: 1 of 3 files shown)


AstroProfundis:
If there is already a `*.log` file (e.g., `tidb.log`) owned by other user, the `tidb-server` might be fail to start. I think there are risks doing so.

lucklove:
> If there is already a `*.log` file (e.g., `tidb.log`) owned by other user, the `tidb-server` might be fail to start. I think there are risks doing so.

It should fail in this case. Think about that it's not just a log, but the data owned by other user. If we chown with -R, the data will be overwrite

pkg/cluster/clusterutil/clusterutil.go
```diff
@@ -30,6 +30,10 @@ func Abs(user, path string) string {
 func MultiDirAbs(user, paths string) []string {
 	var dirs []string
 	for _, path := range strings.Split(paths, ",") {
+		path = strings.Trim(path, " ")
```
lonng, line 33: Why not `strings.TrimSpace`？

lucklove: Fixed

Issue #499, opened 2020-06-15
Ansible imported into tiup without directory conflict detection


zhenjiaogao:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?

Ansible imported into tiup without directory conflict detection. When 
tidb and tikv or pd are deployed together and use the same directory, tiup is used for scale in, this operation will delete the data of other components than the non-target component. 

2. What did you expect to see?

3. What did you see instead?

4. What version of TiUP are you using (`tiup --version`)?
tiup --version
tiup version v1.0.0 tiup
Go Version: go1.13
Git Branch: master
GitHash: 5796405
Build Time: 2020-05-28 18:15:15 +0800

lucklove:
This will be fixed in v1.0.6

PR #502, opened 2020-06-15
cluster: optimize dir conflicts check & the destroying process 

AstroProfundis:
If user `scale-out` a node with `data_dir` or `deploy_dir` manually set to the same as imported node, the dir conflict check is not working for it. Also, when `scale-in` this node, the dir is deleted and the imported node is also effected. 

1. Add dir paths of imported node to the conflict check list, but not reporting error for imported nodes, thus other (non-imported) nodes will fail the check if they have same dir as imported ones.
2. Optimize the component destroying process, only delete a dir path when there is no other nodes share this path

### Check List 

Tests 

 - Unit test

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

### DIFF
--- a/components/cluster/command/scale_in.go
+++ b/components/cluster/command/scale_in.go
@@ -82,7 +82,7 @@ func scaleIn(clusterName string, options operator.Options) error {
 	}
 
 	metadata, err := meta.ClusterMetadata(clusterName)
-	if err != nil && !errors.Is(err, &meta.ValidateErr{}) {
+	if err != nil && !errors.Is(perrs.Cause(err), &meta.ValidateErr{}) {
 		// ignore conflict check error, node may be deployed by former version
 		// that lack of some certain conflict checks
 		return err
--- a/components/cluster/command/scale_in.go
+++ b/components/cluster/command/scale_in.go
@@ -81,7 +81,9 @@ func scaleIn(clusterName string, options operator.Options) error {
 	}
 
 	metadata, err := meta.ClusterMetadata(clusterName)
-	if err != nil {
+	if err != nil && !strings.Contains(err.Error(), "conflicts between") {
+		// ignore conflict check error, node may be deployed by former version
+		// that lack of some certain conflict checks
 		return err
 	}
 
--- a/components/cluster/command/scale_in.go
+++ b/components/cluster/command/scale_in.go
@@ -14,11 +14,12 @@
 package command
 
 import (
+	"errors"
 	"strings"
 
 	"github.com/fatih/color"
 	"github.com/joomcode/errorx"
-	"github.com/pingcap/errors"
+	perrs "github.com/pingcap/errors"
 	"github.com/pingcap/tiup/pkg/cliutil"
 	"github.com/pingcap/tiup/pkg/cluster/clusterutil"
 	"github.com/pingcap/tiup/pkg/cluster/meta"
@@ -77,11 +78,11 @@ func newScaleInCmd() *cobra.Command {
 
 func scaleIn(clusterName string, options operator.Options) error {
 	if tiuputils.IsNotExist(meta.ClusterPath(clusterName, meta.MetaFileName)) {
-		return errors.Errorf("cannot scale-in non-exists cluster %s", clusterName)
+		return perrs.Errorf("cannot scale-in non-exists cluster %s", clusterName)
 	}
 
 	metadata, err := meta.ClusterMetadata(clusterName)
-	if err != nil && !strings.Contains(err.Error(), "conflicts between") {
+	if err != nil && !errors.Is(err, &meta.ValidateErr{}) {
 		// ignore conflict check error, node may be deployed by former version
 		// that lack of some certain conflict checks
 		return err
@@ -159,7 +160,7 @@ func scaleIn(clusterName string, options operator.Options) error {
 			// FIXME: Map possible task errors and give suggestions.
 			return err
 		}
-		return errors.Trace(err)
+		return perrs.Trace(err)
 	}
 
 	log.Infof("Scaled cluster `%s` in successfully", clusterName)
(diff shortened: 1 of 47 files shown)


lonng:
Fix #499

components/cluster/command/scale_in.go
```diff
@@ -81,7 +81,9 @@ func scaleIn(clusterName string, options operator.Options) error {
 	}
 
 	metadata, err := meta.ClusterMetadata(clusterName)
-	if err != nil {
+	if err != nil && !strings.Contains(err.Error(), "conflicts between") {
```
lonng, line 84: It's better to define a error code for `conflicts` instead of string compares.

pkg/cluster/meta/topology.go
```diff
@@ -889,9 +887,78 @@ func (topo *TopologySpecification) dirConflictsDetect() error {
...
+					host := compSpec.FieldByName("Host").String()
+
+					if dir == "" {
+						continue
+					}
+					if !strings.HasPrefix(dir, "/") {
+						switch dirType {
+						case "DataDir":
+							if !strings.HasPrefix(dir, topo.GlobalOptions.DataDir+"/") {
+								dir = filepath.Join(topo.GlobalOptions.DataDir, dir)
```
lonng, line 925: Our data directory is relative to `DeployDir`.

pkg/cluster/meta/topology.go
```diff
@@ -889,9 +887,78 @@ func (topo *TopologySpecification) dirConflictsDetect() error {
...
+
+						}
+					}
+
+					_, exist := dirStats[host+dir]
+					if exist {
+						dirStats[host+dir] += 1
+					} else {
+						dirStats[host+dir] = 1
+					}
```
lonng, line 944: L939-L944 can be simplified to `dirStats[host+dir] += 1`

pkg/cluster/meta/topology.go
```diff
@@ -889,9 +887,78 @@ func (topo *TopologySpecification) dirConflictsDetect() error {
...
+				}
+			}
+		}
+	}
+
+	for k, v := range dirStats {
+		if strings.HasPrefix(k, targetHost+dirPrefix) {
```
lonng, line 951: You haven't converted the relative path of `dirPrefix` to the absolute path, some potential risks exist.
And add some tests for it.

pkg/cluster/meta/topology.go
AstroProfundis: This only works if the data_dir in global options is an absolute one, the data_dir in instance scope itself is already filled with custom default values at unmarshaling

lucklove:
LGTM

pkg/cluster/meta/meta.go
```diff
@@ -13,8 +13,59 @@
...
+// Error implements the error interface
+func (e *ValidateErr) Error() string {
+	return fmt.Sprintf("%s %s for '%v' between '%s' and '%s'", e.target, e.ty, e.value, e.one, e.two)
+}
+
+// Unwrap implements the error interface
+func (e *ValidateErr) Unwrap() error { return nil }
+
+// Is implements the error interface
+func (e *ValidateErr) Is(target error) bool {
```
lonng, line 51: Can you add some tests for this interface implementation?

lonng: Does type assertion not enough?

pkg/cluster/meta/topology.go
```diff
@@ -892,9 +909,78 @@ func (topo *TopologySpecification) dirConflictsDetect() error {
...
+			compSpec := compSpecs.Index(index)
+			// Directory conflicts
+			for _, dirType := range dirTypes {
+				if j, found := findField(compSpec, dirType); found {
+					dir := compSpec.Field(j).String()
+					host := compSpec.FieldByName("Host").String()
+
+					switch dirType { // the same as in logic.go for (*instance)
+					case "DeployDir":
+						dir = clusterutil.Abs(topo.GlobalOptions.User, dir)
```
lonng, line 943: We can extract this line and put it above `dirStats[host+dir] += 1`.

pkg/cluster/meta/topology_dm_test.go
```diff
@@ -205,6 +205,29 @@ dm-worker_servers:
...
+  ssh_port: 220
+  deploy_dir: "test-deploy"
+dm-master_servers:
+  - host: 172.16.5.138
+    deploy_dir: "master-deploy"
+    data_dir: "/test-data/data-1"
+dm-worker_servers:
+  - host: 172.16.5.53
+    data_dir: "test-1"
+`), &topo)
```
lonng, line 227: Test cases not enough, the following cases should be covered:

1. absolute global data path + relative instance  data directory
2. relative global data directory + absolute instance  data directory
3. absolute global data directory + absolute instance data directory
4. relative global data directory + relative instance data directory

pkg/cluster/operation/destroy.go
```diff
@@ -183,10 +186,26 @@ func DestroyComponent(getter ExecutorGetter, instances []meta.Instance, timeout
...
+		case meta.ComponentTiKV,
+			meta.ComponentPD,
+			meta.ComponentPump,
+			meta.ComponentDrainer,
+			meta.ComponentPrometheus,
+			meta.ComponentAlertManager,
+			meta.ComponentDMMaster,
+			meta.ComponentDMWorker,
+			meta.ComponentDMPortal:
+			if cls.CountDir(ins.GetHost(), ins.DataDir()) == 1 {
```
lonng, line 198:
```suggestion
			if cls.CountDir(ins.GetHost(), ins.DataDir()) <= 1 {
```

AstroProfundis: No it should not process when count is 0, it doesn't make any sense to return 0 and there must be something wrong in the counting function.

pkg/cluster/meta/meta.go
AstroProfundis: No, if two error objects has different values, they should be considered different errors

Issue #433, opened 2020-06-05
feature-request: add a new command `tiup mirror set` to switch the mirror address


lonng:
## Feature Request

Provide a new command `tiup mirror set` to switch the mirror address.

Currently, the different mirrors cannot be mixed because they have different root.json (delete the exist manifests also need), we should provide a command to accomplish the process.

The semantics of `tiup mirror set MIRROR_ADDRESS`

Flags
  - `--root` is used to specify the path of root.json
  - `--auto-fetch` will fetch the root.json from remote if `--root` does not be specified.

It's better to save exists root.json in the local system, e.g:

```
$ ls ~/.tiup/bin
hash(mirror-address).root.json hash(mirror-address).root.json
```

When we switch the mirror, we must check if there exists trusted root.json in the local file system.

PR #503, opened 2020-06-15
Support tiup mirror set command

lucklove:

We support use TIUP_MIRRORS to change the mirror, however, after the mirror changed, root.json and manifests will mismatch. 

Introduce the `tiup mirror set` command which will reset root.json change manifests while changing the mirror.

### Check List 

Tests 


lonng:
@lucklove I think some flags should be provided according to https://github.com/pingcap/tiup/issues/433

cmd/mirror.go
```diff
@@ -190,10 +191,33 @@ func delComp(repo, id, version string) error {
 	return nil
 }
 
-// the `repo publish` sub command
+// the `mirror set` sub command
+func newMirrorSetCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "set <mirror-addr>",
+		Short: "set mirror address",
+		Long:  "set mirror address, will replace the root certificate, which is risky",
```
lonng, line 199: I think there is no risk in most scenarios, we shouldn't fetch the `root.json` every time.
Please refer the https://github.com/pingcap/tiup/issues/433

cmd/mirror.go
```diff
@@ -190,10 +191,38 @@ func delComp(repo, id, version string) error {
...
+		Use:   "set <mirror-addr>",
+		Short: "set mirror address",
+		Long:  "set mirror address, will replace the root certificate",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) != 1 {
+				return cmd.Help()
+			}
+
+			addr := args[0]
+			profile := localdata.InitProfile()
```
lonng, line 208: I don't think we need to re-initialize a new `Profile`. There is an another way to retrieve the `Profile` via `environment.GlobalEnv().Profile()`.

cmd/mirror.go
```diff
@@ -190,10 +191,38 @@ func delComp(repo, id, version string) error {
...
+			if err := profile.ResetMirror(addr, root, force); err != nil {
+				fmt.Printf("Failed to set mirror: %s\n", err.Error())
+				return err
+			}
+			fmt.Printf("Set mirror to %s success\n", addr)
+			return nil
+		},
+	}
+	cmd.Flags().StringVarP(&root, "root", "r", root, "root.json path")
```
lonng, line 217:
```suggestion
	cmd.Flags().StringVarP(&root, "root", "r", root, "Specify the path of `root.json`")
```

cmd/mirror.go
```diff
@@ -190,10 +191,38 @@ func delComp(repo, id, version string) error {
...
+	cmd.Flags().BoolVarP(&force, "force", "f", force, "force replace root.json, may be risky")
```
lonng, line 218:
```suggestion
	cmd.Flags().BoolVarP(&force, "force", "f", force, "Auto fetch `root.json` from mirror, may be risky if the mirror is not offline repository")
```

pkg/cluster/template/install/local_install.sh.go
```diff
@@ -68,6 +68,8 @@ fi
 
 chmod 755 "$bin_dir/tiup"
 
+exec "$bin_dir/tiup" mirror set ${script_dir} --force
```
lonng, line 71: Please delete the https://github.com/pingcap/tiup/pull/503/files#diff-ac4e113128741529401e4349350cd6fdR96

pkg/localdata/profile.go
```diff
@@ -321,6 +328,60 @@ func (p *Profile) SelectInstalledVersion(component string, version v0manifest.Ve
...
+
+	var wc io.ReadCloser
+	if strings.HasPrefix(root, "http") {
+		if resp, err := http.Get(root); err == nil {
+			wc = resp.Body
+		} else {
+			return err
+		}
+	} else {
+		if file, err := os.Open(root); err == nil {
```
lonng, line 357: It's better to move +337 to +347 here.

pkg/localdata/profile.go
```diff
@@ -321,6 +328,60 @@ func (p *Profile) SelectInstalledVersion(component string, version v0manifest.Ve
 	return version, nil
 }
 
+// ResetMirror reset root.json and cleanup manifests directory
+func (p *Profile) ResetMirror(addr, root string, forceRefresh bool) error {
```
lonng, line 332: Should we convert the `addr` to absolute path if the current addr is an offline mirror?

lonng: We can early return if the mirror address is not changed.

pkg/localdata/profile.go
```diff
@@ -321,6 +328,60 @@ func (p *Profile) SelectInstalledVersion(component string, version v0manifest.Ve
...
+		} else if strings.HasSuffix(addr, "/") {
+			root = addr + "root.json"
+		} else {
+			root = addr + "/root.json"
+		}
+	}
+
+	var wc io.ReadCloser
+	if strings.HasPrefix(root, "http") {
+		if resp, err := http.Get(root); err == nil {
```
lonng, line 351: Return an error if the mirror starts with `http` and `root == "" && forceRefresh == false`.

pkg/environment/env.go
```diff
@@ -41,8 +41,18 @@ const (
 
 // Mirror return mirror of tiup.
 // If it's not defined, it will use "https://tiup-mirrors.pingcap.com/".
-func Mirror() string {
-	if m := os.Getenv(repository.EnvMirrors); m != "" {
+func Mirror(cfg *localdata.TiUPConfig) string {
+	m := os.Getenv(repository.EnvMirrors)
+	if cfg != nil && cfg.Mirror != "" {
+		if m != "" && m != cfg.Mirror {
+			fmt.Printf(`WARNING: both mirror config and TIUP_MIRRORS have been set.
```
AstroProfundis, line 48: I think environment variable should have higher priority, because in that way it's possible to set different only for one command.

install.sh
```diff
@@ -47,6 +47,8 @@ fi
 
 chmod 755 "$bin_dir/tiup"
 
+exec "$bin_dir/tiup" mirror set $repo --force
```
lonng, line 50: The `--force` flag has been removed, please change the shell script.

pkg/cluster/template/install/local_install.sh.go
lonng, line 71: ditto

pkg/environment/env.go
lonng, line 48: It's better to call `ResetMirror` to switch the mirror to the latest environment variable and output some message to make the user realize that the mirror has been switched.

lonng:
@lucklove Please don't use the `force-push` feature.

pkg/environment/env.go
```diff
@@ -42,7 +42,20 @@ const (
...
-	if m := os.Getenv(repository.EnvMirrors); m != "" {
+	profile := localdata.InitProfile()
+	cfg := profile.Config
+	m := os.Getenv(repository.EnvMirrors)
+	if cfg != nil && cfg.Mirror != "" {
+		if m != "" && m != cfg.Mirror {
+			fmt.Printf(`WARNING: both mirror config and TIUP_MIRRORS have been set.
+Setting mirror to TIUP_MIRRORS(%s)
+`, m)
+			_ = profile.ResetMirror(m, "")
```
lonng, line 53: It's better to print the error message.

pkg/localdata/config.go
```diff
@@ -0,0 +1,50 @@
...
+
+// TiUPConfig represent the config file of TiUP
+type TiUPConfig struct {
+	configBase
+	Mirror string `toml:"mirror"`
+}
+
+// InitConfig returns a TiUPConfig struct which can flush config back to disk
+func InitConfig(root string) *TiUPConfig {
+	config := TiUPConfig{configBase{path.Join(root, "tiup.toml")}, ""}
```
lonng, line 34: I think it's better to reset the mirror when the `tiup.toml` doesn't exist, which means it's the first time to call `Init'Config`.

pkg/environment/env.go
```diff
@@ -42,7 +42,20 @@ const (
 // Mirror return mirror of tiup.
 // If it's not defined, it will use "https://tiup-mirrors.pingcap.com/".
 func Mirror() string {
-	if m := os.Getenv(repository.EnvMirrors); m != "" {
+	profile := localdata.InitProfile()
+	cfg := profile.Config
+	m := os.Getenv(repository.EnvMirrors)
+	if cfg != nil && cfg.Mirror != "" {
```
lonng, line 48: It seems that the `cfg` always not nil.

server/handler/component.go
```diff
@@ -150,15 +149,17 @@ func (h *componentSigner) sign(r *http.Request, m *model.ComponentManifest) (sr
...
+				return err
+			}
 		}
 
 		if err := md.UpdateSnapshotManifest(initTime, func(om *model.SnapshotManifest) *model.SnapshotManifest {
-			om.Signed.Meta["/index.json"] = v1manifest.FileVersion{
-				Version: indexVersion,
-				Length:  uint(indexFi.Size()),
+			if indexFileVersion != nil {
+				om.Signed.Meta["/index.json"] = *indexFileVersion
```
lonng, line 162: There is a constant for `/index.json`.

PR #505, opened 2020-06-16
fix tiup server transaction bug

lobshunter:
Previous tiup server implementation updates timestamp of file before writing file to storage which may cause data corruption.
Because the other transaction with greater read timestamp can read file after timestamp of the file is updated, but before the file is written. 
When the latter transaction commits, the former commit is lost.

Update timestamp after writing file to storage.

### Check List 

Tests 

 - Unit test


lobshunter:
@lucklove PTAL

PR #514, opened 2020-06-18
Support ignore config check result

lucklove:

This PR ignore config check result in unnecessary case and support `--ignore-config-check` flag in necessary case.


components/cluster/command/reload.go
```diff
@@ -73,6 +74,7 @@ func newReloadCmd() *cobra.Command {
 	cmd.Flags().StringSliceVarP(&gOpt.Roles, "role", "R", nil, "Only start specified roles")
 	cmd.Flags().StringSliceVarP(&gOpt.Nodes, "node", "N", nil, "Only start specified nodes")
 	cmd.Flags().Int64Var(&gOpt.APITimeout, "transfer-timeout", 300, "Timeout in seconds when transferring PD and TiKV store leaders")
+	cmd.Flags().BoolVarP(&ignoreConfigCheck, "ignore-config-check", "", ignoreConfigCheck, "Ignore the config check result")
```
lonng, line 77: Can we add the `ignoreConfigCheck` field to `gOpt` directly? I think `skip-config-check` more properly.

pkg/cluster/task/init_config.go
```diff
@@ -45,6 +46,9 @@ func (c *InitConfig) Execute(ctx *Context) error {
 
 	err := c.instance.InitConfig(exec, c.clusterName, c.clusterVersion, c.deployUser, c.paths)
 	if err != nil {
+		if _, ok := err.(meta.ConfigCheckError); ok && c.ignoreCheck {
```
lonng, line 49: It's better to use `errors.Cause` to extract the root cause of the returned error.

components/cluster/command/reload.go
lucklove: I think `ignore` is better than `skip`, `ignore` means we still check the config, but ignore the result (just put it in the audit log). The `skip` should really do nothing at that step.

pkg/cluster/meta/server_config.go
```diff
@@ -37,10 +37,8 @@ const (
 	migrateLockName = "tiup-migrate.lck"
 )
 
-// ConfigCheckError represent error occured in config check stage
-type ConfigCheckError struct {
-	error
-}
+// ErrorCheckConfig represent error occured in config check stage
+var ErrorCheckConfig = errors.New("check config failed")
```
lonng, line 41: Use standard library `errors` package instead of `github.com/pingcap/errors` to prevent add stack information to the raw message.

components/cluster/command/reload.go
lonng: OK, prefer to put the `ignoreConfigCheck` into `operator.Options`.

pkg/cluster/task/init_config.go
```diff
@@ -45,6 +46,9 @@ func (c *InitConfig) Execute(ctx *Context) error {
 
 	err := c.instance.InitConfig(exec, c.clusterName, c.clusterVersion, c.deployUser, c.paths)
 	if err != nil {
+		if errors.Cause(err) == meta.ErrorCheckConfig {
```
lonng, line 49: `if c.ignoreCheck && errors.Cause(err) == meta.ErrorCheckConfig`.
Please add some unit tests for this. It's better to mock a `meta.Instance` to test the `InitConfig` task.

PR #525, opened 2020-06-22
Refactor the DMS and Cluster components topology

lonng:

1. This PR Refactors the DMS and Cluster components topology.

2. The original `topology.go` and `logic.go` have been separated into some individual files according to the component name:
    1. cdc.go
    2. grafana.go
    3. ...

3. Remove some DM logic.


components/cluster/command/deploy.go
```diff
@@ -353,14 +354,14 @@ func deploy(clusterName, clusterVersion, topoFile string, opt deployOptions) err
 func buildMonitoredDeployTask(
 	clusterName string,
 	uniqueHosts map[string]hostInfo, // host -> ssh-port, os, arch
-	globalOptions meta.GlobalOptions,
-	monitoredOptions meta.MonitoredOptions,
+	globalOptions spec.GlobalOptions,
+	monitoredOptions spec.MonitoredOptions,
```
lichunzhu, line 358: If dms use the similar GlobalOptions, should dms rewrite another one?

lonng: I think eventually `GlobalOptions` will be different with DMS (MonitoredOpions/Grafana/Prometheus/AlertManager will be the same, but we can leave them to the future PR).

PR #526, opened 2020-06-23
Support platform "any/any"

lucklove:
This PR introduces the "any" platform where the packages can satisfy all platforms.

In the package lookup stage, I add another step to lookup "any/any" if the previous lookup failed.

### Check List 

Tests 

 - Unit test


cmd/mirror.go
```diff
@@ -374,6 +379,20 @@ func newMirrorPublishCmd() *cobra.Command {
 	return cmd
 }
 
+func validatePlatform(goos, goarch string) error {
+	// Only support any/any, don't support linux/any, any/amd64 .etc.
+	if goos == "any" && goarch == "any" {
```
AstroProfundis, line 384: I don't think this bool expr is correct

pkg/repository/v1manifest/manifest.go
```diff
@@ -47,6 +47,9 @@ const (
 	// SpecVersion of current, maybe we could expand it later
 	CurrentSpecVersion = "0.1.0"
 
+	// AnyPlatform can match any os and any arch
```
AstroProfundis, line 50: AnyPlatform is the ID for platform independent components

pkg/repository/v1manifest/manifest_test.go
```diff
@@ -13,5 +13,38 @@
...
 // TODO test that invalid manifests trigger errors
 // TODO test SignAndWrite
+
+func TestVersionItem(t *testing.T) {
+	manifest := &Component{
+		Platforms: map[string]map[string]VersionItem{
+			"linux/amd64": map[string]VersionItem{
+				"v1.0.0": VersionItem{Entry: "test"},
+			},
+			"any/any": map[string]VersionItem{
```
AstroProfundis, line 31: should also add sth like `darwin/any`, `any/arm64`

cmd/mirror.go
lucklove: It's correct. any/any is supported, any/amd64 not

PR #527, opened 2020-06-23
Enable playground TiFlash component on Mac

birdstorm:
After all tiflash mac binary are updated to tiup mirror, this PR will be safe to test & merge.

enable playground to use tiflash component on mac by default.

### Check List 

Tests 

Related changes


lonng:
@birdstorm I think there should be a minimal version check because not all version support `darwin_amd64`.

PR #531, opened 2020-06-24
cluster: integrate tispark

### LINKED ISSUE #348: Add TiSpark support
- [ ] deploy
 - [ ] scale in/out
 - [ ] upgrade
 - [ ] import
 - [ ] monitoring configs
 - [ ] destroy

AstroProfundis:

Add TiSpark support to `cluster` component, being able to deploy and manage a cluster with TiSpark as one of the components.

Close #348 

 - [x] Add topology and component specifications for TiSpark
 - [x] Download packages as needed
 - [x] Render configs and scripts for TiSpark from template and specs
 - [x] Deploy necessary files to server
 - [x] Start/Stop tispark nodes
 - [x] Scale out of TiSpark nodes
 - [x] Handling upgrade process of TiSpark nodes
 - [x] Destroy / scale in of TiSpark nodes
 - [x] UI adjustment and testing 🚧
 - [x] Add documentation 🚧
 - Import TiSpark nodes from TiDB-Ansible (will be done in another PR)

### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes


Makefile
```diff
@@ -1,4 +1,5 @@
-.PHONY: server
+.PHONY: cmd components server
```
lonng, line 1: Is this change expected? There is no target named `cmd`.

components/cluster/command/deploy.go
```diff
@@ -146,6 +146,12 @@ func confirmTopology(clusterName, version string, topo *spec.Specification, patc
 		log.Errorf("    3. The component marked as `patched` has been replaced by previours patch command.")
 	}
 
+	if (topo.TiSparkMasters != nil && len(topo.TiSparkMasters) > 0) ||
```
lonng, line 149: This part can be simplified to `if len(topo.TiSparkMasters) > 0) || len(topo. TiSparkWorkers) > 0)` because `len(nil)` always returning 0.

Makefile
AstroProfundis: There is a directory `cmd`, so if run `make cmd`, it will try to make the directory. We don't have any Makefile in `cmd` so it's not a problem for now.

components/cluster/command/deploy.go
```diff
@@ -270,37 +276,80 @@ func deploy(clusterName, clusterVersion, topoFile string, opt deployOptions) err
...
-					Log:    logDir,
-					Cache:  spec.ClusterPath(clusterName, spec.TempConfigPath),
-				},
-			).
-			BuildAsStep(fmt.Sprintf("  - Copy %s -> %s", inst.ComponentName(), inst.GetHost()))
-		deployCompTasks = append(deployCompTasks, t)
+			).Shell( // spark is under a subdir, move it to deploy dir
+				inst.GetHost(),
+				fmt.Sprintf(
+					"cp -rf %[1]s %[2]s/ && cp -rf %[3]s/* %[2]s/ && rm -rf %[1]s %[3]s",
```
lonng, line 304: I think it's no matter that we retain the `filepath.Join(deployDir, "bin", sparkSubPath)` and `filepath.Join(deployDir, sparkSubPath)` directories.

components/cluster/command/deploy.go
```diff
@@ -373,6 +422,11 @@ func buildMonitoredDeployTask(
 		version := spec.ComponentVersion(comp, version)
 
 		for host, info := range uniqueHosts {
+			// FIXME: as the uniqueHosts list is built with os-arch as part of the key,
+			// for platform independent packages, it will be downloaded multiple times
+			// and be saved with different file names in the packages dir, the tarballs
+			// are identical and the only difference is platform in filename.
```
lonng, line 428: Do you mind creating an issue for it and link the URL here?

components/cluster/command/destroy.go
```diff
@@ -69,7 +69,8 @@ You can retain some nodes and roles data when destroy cluster, eg:
 
 			logger.EnableAuditLog()
 			metadata, err := spec.ClusterMetadata(clusterName)
-			if err != nil && !errors.Is(perrs.Cause(err), meta.ErrValidate) {
+			if err != nil && !errors.Is(perrs.Cause(err), meta.ErrValidate) &&
```
lonng, line 72: It's better to unify the way of check error type.

components/cluster/command/scale_out.go
```diff
@@ -247,11 +248,50 @@ func buildScaleOutTask(
 				filepath.Join(deployDir, "conf"),
 				filepath.Join(deployDir, "scripts")).
 			Mkdir(metadata.User, inst.GetHost(), dataDirs...)
+
+		// copy dependency component if needed
+		switch inst.ComponentName() {
```
lonng, line 253: IMO, the L253-L294 can be extracted as a function(don't repeat yourself).

examples/topology.example.yaml
```diff
@@ -211,6 +211,54 @@ tiflash_servers:
...
+    #spark_env:
+    #  SPARK_EXECUTOR_CORES: 5
+    #  SPARK_EXECUTOR_MEMORY: "10g"
+    #  SPARK_WORKER_CORES: 5
+    #  SPARK_WORKER_MEMORY: "10g"
+
+# NOTE: multiple worker nodes on the same host is not supported
+tispark_workers:
+  - host: 10.0.1.22
+    #spark_config:
```
lonng, line 234: Can we use `config` instead of `spark_config` to keep consistent with the others?

examples/topology.example.yaml
```diff
@@ -211,6 +211,54 @@ tiflash_servers:
...
+    #  spark.driver.memory: "2g"
+    #  spark.eventLog.enabled: "False"
+    #  spark.tispark.grpc.framesize: 268435456
+    #  spark.tispark.grpc.timeout_in_sec: 100
+    #  spark.tispark.meta.reload_period_in_sec: 60
+    #  spark.tispark.request.command.priority: "Low"
+    #  spark.tispark.table.scan_concurrency: 256
+    #spark_env:
```
lonng, line 242: Prefer to `env`.

pkg/cluster/embed/autogen_pkger.go
```diff
@@ -16,29 +16,33 @@ package embed
...
 
 func init() {
-	autogenFiles["/templates/scripts/run_pump.sh.tpl"] = "IyEvYmluL2Jhc2gKc2V0IC1lCgojIFdBUk5JTkc6IFRoaXMgZmlsZSB3YXMgYXV0by1nZW5lcmF0ZWQuIERvIG5vdCBlZGl0IQojICAgICAgICAgIEFsbCB5b3VyIGVkaXQgbWlnaHQgYmUgb3ZlcndyaXR0ZW4hCkRFUExPWV9ESVI9e3suRGVwbG95RGlyfX0KCmNkICIke0RFUExPWV9ESVJ9IiB8fCBleGl0IDEKCnt7LSBkZWZpbmUgIlBETGlzdCJ9fQogIHt7LSByYW5nZSAkaWR4LCAkcGQgOj0gLn19CiAgICB7ey0gaWYgZXEgJGlkeCAwfX0KICAgICAge3stICRwZC5TY2hlbWV9fTovL3t7JHBkLklQfX06e3skcGQuQ2xpZW50UG9ydH19CiAgICB7ey0gZWxzZSAtfX0KICAgICAgLHt7LSAkcGQuU2NoZW1lfX06Ly97eyRwZC5JUH19Ont7JHBkLkNsaWVudFBvcnR9fQogICAge3stIGVuZH19CiAge3stIGVuZH19Cnt7LSBlbmR9fQoKe3stIGlmIC5OdW1hTm9kZX19CmV4ZWMgbnVtYWN0bCAtLWNwdW5vZGViaW5kPXt7Lk51bWFOb2RlfX0gLS1tZW1iaW5kPXt7Lk51bWFOb2RlfX0gYmluL3B1bXAgXAp7ey0gZWxzZX19CmV4ZWMgYmluL3B1bXAgXAp7ey0gZW5kfX0KICAgIC0tbm9kZS1pZD0ie3suTm9kZUlEfX0iIFwKICAgIC0tYWRkcj0iMC4wLjAuMDp7ey5Qb3J0fX0iIFwKICAgIC0tYWR2ZXJ0aXNlLWFkZHI9Int7Lkhvc3R9fTp7ey5Qb3J0fX0iIFwKICAgIC0tcGQtdXJscz0ie3t0ZW1wbGF0ZSAiUERMaXN0IiAuRW5kcG9pbnRzfX0iIFwKICAgIC0tZGF0YS1kaXI9Int7LkRhdGFEaXJ9fSIgXAogICAgLS1sb2ctZmlsZT0ie3suTG9nRGlyfX0vcHVtcC5sb2ciIFwKICAgIC0tY29uZmlnPWNvbmYvcHVtcC50b21sIDI+PiAie3suTG9nRGlyfX0vcHVtcF9zdGRlcnIubG9nIgo="
```
lonng, line 19: FIXME: we should refactor the `pkger` tool and sorts all filenames.

pkg/cluster/spec/instance.go
```diff
@@ -272,11 +281,19 @@ func (i *instance) LogDir() string {
 }
 
 func (i *instance) OS() string {
-	return reflect.ValueOf(i.InstanceSpec).FieldByName("OS").Interface().(string)
+	v := reflect.ValueOf(i.InstanceSpec).FieldByName("OS")
```
lonng, line 284: I don't think there should be a component that hasn't `OS/Arch` (`any` is an acceptable value).

pkg/cluster/spec/spec.go
```diff
@@ -36,7 +36,11 @@ const (
 
 // general role names
 var (
-	RoleMonitor = "monitor"
+	RoleMonitor              = "monitor"
+	RoleTiSparkMaster        = "tispark-master"
+	RoleTiSparkWorker        = "tispark-worker"
+	ErrNoTiSparkMaster       = errors.New("There must be a Spark master node if you want to use the TiSpark component")
```
lonng, line 42: Error string should start with the lower case according to Golang idiomatical

pkg/cluster/spec/spec.go
```diff
@@ -36,7 +36,11 @@ const (
...
+	ErrMultipleTiSparkMaster = errors.New("TiSpark enabled cluster with more than 1 Spark master node is not supported")
```
lonng, line 43: ditto

components/cluster/command/deploy.go
AstroProfundis: They're mixed with decompressed files and may cause confuses to users.

components/cluster/command/scale_out.go
AstroProfundis: They should not be in `deploy.go` from the very begaining, they should be handled in manifest's dependency logic. But we don't have that implemented now. It's ok to put these to a dedicated task like `DeploySpark` or so, but I do this in `deploy.go` for I don't expect these lines to be here for long.

pkg/cluster/spec/spec.go
```diff
@@ -495,9 +501,32 @@ func (s *Specification) CountDir(targetHost, dirPrefix string) int {
...
+		}
+		return ErrNoTiSparkMaster
+	}
+
+	// We only support 1 spark master at present
+	if s.TiSparkMasters != nil && len(s.TiSparkMasters) > 1 {
+		return ErrMultipleTiSparkMaster
+	}
+
+	return nil
```
lonng, line 520: This function can be simplified if we use `len` function instead of check the `nil` value.

examples/topology.example.yaml
AstroProfundis: Then what about the `spark_env`, also rename to `env`?

pkg/cluster/spec/tispark.go
```diff
@@ -0,0 +1,394 @@
...
+
+				usedPorts: []int{
+					s.Port,
+					s.WebPort,
+				},
+				usedDirs: []string{
+					s.DeployDir,
+				},
+				statusFn: func(_ ...string) string {
+					return "N/A"
```
lonng, line 129: I think we should implement the `statusFn` for the tispark.

pkg/cluster/spec/instance.go
AstroProfundis: If it has these fields, the global one will be used, and then it requires user to specify `any` in topology file

pkg/cluster/spec/util.go
```diff
@@ -57,7 +57,7 @@ func ClusterMetadata(clusterName string) (*ClusterMeta, error) {
 	var cm ClusterMeta
 	err := GetSpecManager().Metadata(clusterName, &cm)
 	if err != nil {
-		return nil, errors.AddStack(err)
+		return &cm, errors.AddStack(err)
```
lonng, line 60: Should we return the `cm` even though error encountered?

AstroProfundis: Yes, it is possible the error could be ignored and the `cm` continue be processed. A typical scenario is when the topology file is broken (from `edit-config`, manual modification, or a broken legacy release), we could keep a max possibility that operations like `display` and `destroy` are still working.

examples/topology.example.yaml
lucklove: > Then what about the spark_env, also rename to env?
 I think we should add env support to ever components except spark

pkg/cluster/embed/autogen_pkger.go
lonng: https://github.com/pingcap/tiup/pull/599

pkg/cluster/spec/instance.go
lonng: You can use `OS string default:any` to avoid inherit global value.

components/cluster/command/scale_out.go
```diff
@@ -252,44 +257,13 @@ func buildScaleOutTask(
...
-			tb = tb.Shell( // move tispark jar to correct path
-				inst.GetHost(),
-				fmt.Sprintf(
-					"cp -f %[1]s/*.jar %[2]s/jars/ && rm -f %[1]s/*.jar",
-					filepath.Join(deployDir, "bin"),
-					deployDir,
-				),
-				false, // (not) sudo
-			)
+			tb = tb.DeploySpark(inst, version, deployDir)
```
lonng, line 260: Seems the patched tiflash will not be deployed when scale-out.

components/cluster/command/check.go
```diff
@@ -167,6 +167,7 @@ func checkSystemInfo(s *cliutil.SSHConnectionProps, topo *spec.Specification, op
 					inst.OS(),
 					inst.Arch(),
 					insightVer,
+					"", // use default srcPath
```
lonng, line 170: I think the better to define an interface for the component source file. e.g:
```
type ComponentSource interface {
    FilePath() string
}

type OSArchComponent struct {
    OS string
    Arch string
    Version string
}

func (...) FilePath() string {
    return fmt.sprintf()
}

type PatchedComponent string
func (p PatchedComponent) FilePath() string {
    return string(p)
}
```

lucklove: Maybe we should go further:
- Define a `Component` interface which contains `Deploy` or other needed methods
- Implement the `Component` interface in `TiDBComponent`, `TiSparkComponent` etc. So that we don't need to check what the component it is.

AstroProfundis: That would be a huge refactor and is already out of the scope of this PR.

components/dm/command/scale_out.go
```diff
@@ -234,22 +234,31 @@ func buildScaleOutTask(
...
-				Log:    logDir,
-			},
-		).Build()
+
+		t := tb.CopyComponent(
+				inst.ComponentName(),
+				inst.OS(),
+				inst.Arch(),
+				version,
+				"", // use default srcPath
```
lucklove, line 248: It seems should use srcPath here?

pkg/cluster/spec/instance.go
```diff
@@ -212,6 +214,13 @@ func (i *instance) InstanceName() string {
 
 // ServiceName implements Instance interface
 func (i *instance) ServiceName() string {
+	switch i.ComponentName() {
+	case ComponentSpark, ComponentTiSpark:
+		if i.port > 0 {
+			return fmt.Sprintf("%s-%d.service", i.Role(), i.port)
+		}
+		return fmt.Sprintf("%s.service", i.Role())
```
lucklove, line 222: I think we should specify a default port in the Unmarshal stage. So `i.port > 0` should always be true.

pkg/cluster/spec/spec.go
```diff
@@ -495,9 +501,30 @@ func (s *Specification) CountDir(targetHost, dirPrefix string) int {
...
+}
+
 // Validate validates the topology specification and produce error if the
 // specification invalid (e.g: port conflicts or directory conflicts)
 func (s *Specification) Validate() error {
+	if err := s.validateTiSparkSpec(); err != nil {
```
lucklove, line 524: I think this should be checked after other checks (eg. platformConflictsDetect)
For example, I have this scale-out.yaml:
```
tidb_servers:
  - host: 172.16.5.139
  - host: 172.16.5.139
```
This is clearly in conflict. However, if there is a `ErrNoTiSparkMaster` error, it may be ignored in https://github.com/pingcap/tiup/pull/531/files#diff-19d3b6c1bf7701c04716d697df146cc1R100

pkg/cluster/template/config/tispark.go
```diff
@@ -0,0 +1,82 @@
...
+		return nil, err
+	}
+	return c.ConfigWithTemplate(string(tpl))
+}
+
+// ConfigToFile write config content to specific path
+func (c *TiSparkConfig) ConfigToFile(file string) error {
+	config, err := c.Config()
+	if err != nil {
+		panic(err)
```
lucklove, line 63: should return error instead of panic

pkg/cluster/spec/instance.go
AstroProfundis: It does, if we defined a default value for port in the spec tag.

pkg/utils/ioutil.go
```diff
@@ -148,3 +151,54 @@ func Move(src, dst string) error {
...
+
+// CopyFile copies a file from src to dst
+func CopyFile(src, dst string) error {
+	in, err := os.Open(src)
+	if err != nil {
+		return err
+	}
+	defer in.Close()
+
+	if _, err := os.Stat(dst); !os.IsNotExist(err) {
```
lucklove, line 174:
```suggestion
	if IsExist(dst) {
```

pkg/utils/ioutil.go
```diff
@@ -148,3 +151,54 @@ func Move(src, dst string) error {
...
+		if os.IsNotExist(err) {
+			return os.MkdirAll(path, 0755)
+		}
+		return err
+	}
+	return nil
+}
+
+// CopyFile copies a file from src to dst
+func CopyFile(src, dst string) error {
```
lucklove, line 167: Can we use `utils.Copy`, which can copy both directories and files.

AstroProfundis: It's not related to the tispark integration, we could do that in a dedicated PR.

examples/topology.example.yaml
birdstorm: I think spark_env should not be set for workers.

examples/topology.example.yaml
```diff
@@ -211,6 +211,54 @@ tiflash_servers:
 #   - host: 10.0.1.21
 #   - host: 10.0.1.22
 
+# NOTE: Only 1 master node is supported for now
+tispark_masters:
+  - host: 10.0.1.21
```
birdstorm, line 216: is it possible to have multiple spark masters?

examples/topology.example.yaml
birdstorm: I think config should be set to tispark_master only. workers should be unaware of these configs.

birdstorm:
I think that Spark masters and Spark workers should share the same env/conf. We can just set them in master's config.

examples/topology.example.yaml
AstroProfundis: Is there any way to setup multiple masters without setting up a zookeeper? I didn't see such configs in `tidb-ansible`.

birdstorm: I think we can remain the logic to support only one master node as it is.

PR #537, opened 2020-06-27
Update install.sh

tohghua:
exec在cent os上面造成它后面的语句不会执行。

PR #542, opened 2020-06-28
Fix install script

lucklove:
The install prcess didn't print out the latter action user should
take after install tiup


lucklove:
dup with https://github.com/pingcap/tiup/pull/537

Issue #423, opened 2020-06-04
TiUP Cluster Scale-in process may be unexpected as designed.


birdstorm:
## Bug Report

When enabling Force option in scale-in process, the logic goes as follows:

https://github.com/pingcap/tiup/blob/840ce6857fa24dde3901879da3111f4f0e93ca4a/pkg/cluster/operation/scale_in.go#L119-L137

The logic of scale-in forces to stop and destroy the current instance, and it returns **BEFORE** stores are removed from PD, which is shown as following lines:

https://github.com/pingcap/tiup/blob/840ce6857fa24dde3901879da3111f4f0e93ca4a/pkg/cluster/operation/scale_in.go#L201-L248

I think it does not work as expected.

4. What version of TiUP are you using (`tiup --version`)?

master

PR #546, opened 2020-06-28
always try to update state when force scale in instance

july2993:
fix https://github.com/pingcap/tiup/issues/423
1.  for pd/tikv/tiflash Always try to delete the store by calling the API of pd.
2. for pump&drainer update state as offline after stop and destroy.

also, tiflash should scale-in async like tikv.

tests:
manual force scale in tiflash/pump/drainer and check the state.


pkg/cluster/operation/action.go
```diff
@@ -200,7 +200,7 @@ func DestroyClusterTombstone(
 			continue
 		}
 
-		id := s.Host + ":" + strconv.Itoa(s.FlashProxyPort)
+		id := s.Host + ":" + strconv.Itoa(s.FlashServicePort)
```
lonng, line 203: @birdstorm PTAL

lonng:
Please add proper labels for this PR.

pkg/cluster/operation/action.go
birdstorm: lgtm

Issue #101, opened 2020-04-04
Disable space_placeholder_file for playgrounds


wwar:
## Feature Request

I think what you are doing with tiup is great! I have a small nit:

**Is your feature request related to a problem? Please describe:**

Each of the playgrounds create TiKV's `space_placeholder_file` (2GB). I think this makes sense for production environments, but less so on a desktop where you might have multiple playgrounds.

**Describe the feature you'd like:**

Disable space place holders.

**Describe alternatives you've considered:**

It's not that harmful, alternative would be to keep as is. That will limit having multiple playgrounds.

**Teachability, Documentation, Adoption, Migration Strategy:**

N/A

lonng:
@BusyJay @stn PTAL

BusyJay:
@lonng playground can set `storage.reserve-space` to zero by default.

lonng:
@BusyJay Thanks, I will fix it.

PR #548, opened 2020-06-29
Disable space holder of tikv in playground

july2993:
Reserve disk space to make tikv would have enough space to compact when
the disk is full.
the default value is 2G, we can disable it in playground.

Fix https://github.com/pingcap/tiup/issues/101


lonng:
@july2993 It's better to change the `tiflash_proxy_config` too.

july2993:
> @july2993 It's better to change the `tiflash_proxy_config` too.

I don't see tiflash creating such a file, so I think it has no such feature.

Issue #335, opened 2020-05-25
Display some message for impossible auto-fixing


hhkbp2:
Hi there,

When we use `tiup check --apply`, it's a bit confusing if the output doesn't show any message for those checks impossible to fix. Usually a program succeeds silently but it outputs some message when it fails or doesn't support. I suggest to add some message for these checks that could not be fixed, in:

https://github.com/pingcap-incubator/tiup-cluster/blob/master/cmd/cluster/command/check.go#L399

lucklove:
Could you provide your tiup version?
```
tiup --version
```

Is this issue still exists in the newest version?

hhkbp2:
PR #550 is created to fix this issue. PTAL.

AstroProfundis:
At present, if a failed check is impossible to be fixed by us, it will show a message/suggestion in the check result table.

The problem is, there will **always** be something we can not fix (i.e., cpu threads, memory size, os version, etc.) and I think it's not a good experience to show error message for user for them.

hhkbp2:
> At present, if a failed check is impossible to be fixed by us, it will show a message/suggestion in the check result table.
> 
> The problem is, there will **always** be something we can not fix (i.e., cpu threads, memory size, os version, etc.) and I think it's not a good experience to show error message for user for them.

So if there is something that can not be fixed, it just does nothing and keep silent about this fact that "this check failure can not be fixed" and let the user to guess what happens after running "check --apply"? Are you sure it's a good experience?

AstroProfundis:
Ah, my fault, this was about the message in check result table, not the command itself, then it looks fine for me.

PR #550, opened 2020-06-29
cluster: show message for impossible fix

hhkbp2:

This PR fixes issue #335 

Add message for failed checking result that's impossible to fix by `tiup cluster check --apply`, for better user experience.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes


components/cluster/command/check.go
```diff
@@ -414,8 +414,7 @@ func fixFailedChecks(ctx *task.Context, host string, res *operator.CheckResult,
 		operator.CheckNameMem,
 		operator.CheckNameCommand,
 		operator.CheckNameFio:
-		// don't show unsupported message for checks that are impossible to fix by us
-		return "", nil
+		msg = fmt.Sprintf("%s, auto fixing not possible", res)
```
AstroProfundis, line 417: This case section can be just deleted in whole, use the `default` is enough.

hhkbp2: Comment addressed. PTAL.

Issue #556, opened 2020-06-30
tiup cluster patch can't overwrite twice


lucklove:
<img width="991" alt="屏幕快照 2020-06-30 下午6 28 25" src="https://user-images.githubusercontent.com/8718109/86116152-8a6c4680-baff-11ea-9bb4-504208a23941.png">
Can't overwrite existed symbol

PR #558, opened 2020-07-01
Bugfix: patch can't overwrite twice

lucklove:
Fix https://github.com/pingcap/tiup/issues/556


components/cluster/command/patch.go
```diff
@@ -186,13 +187,23 @@ func overwritePatch(clusterName, comp, packagePath string) error {
...
 		return err
 	}
+
 	checksum, err := tiuputils.Checksum(packagePath)
 	if err != nil {
 		return err
 	}
+
 	tg := spec.ClusterPath(clusterName, spec.PatchDirName, comp+"-"+checksum[:7]+".tar.gz")
+	if utils.IsExist(tg) {
```
lonng, line 197: I think we can skip the `CopyFile` is the target file exists.

lucklove: Not a good idea, for example:
```sh
# In tidb code directory
make tidb
tar -czf tidb.tar.gz tidb-server
tiup cluster patch test-cluster $(pwd)/tidb.tar.gz -R tidb

# Test tidb in the cluster found a bug, changed the code
make tidb
tar -czf tidb.tar.gz tidb-server
tiup cluster patch test-cluster $(pwd)/tidb.tar.gz -R tidb
```

This will skip the second patch.

lonng: The checksum will be different.

Issue #557, opened 2020-07-01
support deploy and manage dm 2.0 


july2993:
### Description

At first, we had supported deploy dm but later was changed to work on the `dms` now, we abandon the `dms`
and need to support deploy dm directly again.

Thinks we need to do mainly:
- try revert some code make it works, mainly ref: https://github.com/pingcap/tiup/pull/490
- also, refactor the code because the codes are mainly copied from the cluster containing too much duplicate and useless code.

### Category
new component/new feature

### Value
better easy to use compare to DM-Ansible

### TODO list

relate issues:

- [x] Should not use the same directory of tiup-cluster to save meta, will conflict if using the save cluster name, we can encapsulate the logic about metafiles. #567 
~~- [ ] refactor codes in topology_dm.go to avoid much duplicate code with the one that cluster use.~~
- [x] support monitor(p8s and grafana) #656 
~~- [ ] avoid use `bin/bin/` in the deploy dir~~
- [x] support import from ansible with 1.* and upgrade to 2.0 #672 #724 
- [x] add relate integration tests #629 
- [x] add docs https://github.com/pingcap/docs-dm/pull/225
- [x] reject deploy < 2.0 version #645 
- [x] remove dm-portal support since it's deprecated in plan by DM team #629 
- [x] update `examples/topology.dm.example.yaml` #679
...

support commands(#587):
- [x] deploy
- [x] start
- [x] stop
- [x] restart
- [x] destroy
- [x] exec
- [x] list 
- [x] upgrade
- [x] patch
- [x] scale-in/scale-out
- [x] edit config
- [x] reload
- [x] display

### Documentations
- TODO

### Schedule
GanttStart: 2020-07-01
GanttDue: 2020-08-31
GanttProgress: 100%

PR #567, opened 2020-07-02
encapsulate operation about meta file

july2993:
ref #557 

encapsulate operation about meta file in pkg/meta/spec*


pkg/dm/spec/cluster.go
```diff
@@ -34,42 +30,7 @@ type DMMeta struct {
...
-		return nil, errors.Trace(err)
-	}
-
-	if err = yaml.Unmarshal(yamlFile, &cm); err != nil {
-		return nil, errors.Trace(err)
-	}
-	return &cm, nil
+// NewDMSpec create dm Spec.
+func NewDMSpec() *meta.Spec {
+	return meta.NewSpec(filepath.Join(cspec.ProfileDir(), "dm"))
```
lonng, line 35: Prefer to define a constant for `dm` string literal.

pkg/meta/spec.go
```diff
@@ -0,0 +1,152 @@
+package meta
```
lonng, line 1: The license header is missing.

pkg/meta/spec_test.go
```diff
@@ -0,0 +1,75 @@
+package meta
```
lonng, line 1: ditto

pkg/meta/spec.go
```diff
@@ -0,0 +1,152 @@
...
+	// MetaFileName is the file name of the meta file.
+	MetaFileName = "meta.yaml"
+	// PatchDirName is the directory to store patch file eg. {PatchDirName}/tidb-hotfix.tar.gz
+	PatchDirName = "patch"
+	// BackupDirName is the directory to save backup files.
+	BackupDirName = "backup"
+)
+
+// Spec control management of spec meta data.
+type Spec struct {
```
lonng, line 35: Prefer to choose another name, e.g: `SpecManager`.

pkg/meta/spec.go
```diff
@@ -0,0 +1,152 @@
...
+func NewSpec(base string) *Spec {
+	return &Spec{
+		base: base,
+	}
+}
+
+// ClusterPath returns the full path to a subpath (file or directory) of a
+// cluster, it is a subdir in the profile dir of the user, with the cluster name
+// as its name.
+func (s *Spec) ClusterPath(cluster string, subpath ...string) string {
```
lonng, line 49: It's a good idea to abstract a new type to organize metadata, I think all exported methods name should refactor and shouldn't contain the `Cluster` .

components/cluster/command/root.go
```diff
@@ -49,6 +50,8 @@ var (
 	skipConfirm bool
 )
 
+var tidbSpec *meta.SpecManager
```
lonng, line 53: I think the `tidbSpec` should the global variable of `pkg/cluster/spec` and make it reusable in https://github.com/pingcap/tiup/pull/567/files#diff-fdca3c2d2bcb7afb276475bee9c63c47R51

components/dm/command/root.go
```diff
@@ -42,6 +44,8 @@ var (
 	skipConfirm bool
 )
 
+var dmspec *meta.SpecManager
```
lonng, line 47: ditto

components/cluster/command/root.go
july2993: resolve in https://github.com/pingcap/tiup/pull/567/commits/a82ee6c3f48fd751629f13db687915dc646e3c91

PR #576, opened 2020-07-06
cluster/scale-out: check for imported field earlier

AstroProfundis:
`imported: true` is not allowed for scale out instances, currently we check that when building scale out tasks, this might be too late for some circumstances, this PR move that check to an early stage of the process, to let it fail ASAP.

Move the check to an early stage of the process.

### Check List 

Tests 

Code changes

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/components/cluster/command/scale_out.go
+++ b/components/cluster/command/scale_out.go
@@ -95,6 +95,21 @@ func scaleOut(clusterName, topoFile string, opt scaleOutOptions) error {
 		return err
 	}
 
+	var iterErr error
+	iterErr = nil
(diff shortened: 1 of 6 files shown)


components/cluster/command/scale_out.go
```diff
@@ -95,6 +95,21 @@ func scaleOut(clusterName, topoFile string, opt scaleOutOptions) error {
 		return err
 	}
 
+	var iterErr error
+	iterErr = nil
+	newPart.IterInstance(func(instance spec.Instance) {
```
lonng, line 100: It's better to add a method for `spec` and add some tests for it.

Issue #104, opened 2020-04-04
Write to .bashrc if .bash_profile doesn't exist?


wwar:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?

Ran the tiup installer.

2. What did you expect to see?

I am not a bash expert, but after running the install the `tiup` binary was not in my `PATH`. I debugged it, and it looks like it created+wrote to `.bash_profile` (which Ubuntu doesn't use).

I thought that it would use `.bash_profile` if it existed.. but it seems not. So I moved the contents to my `.bashrc` file and it works reliably. I also tried `.profile`, but that didn't work.

I am using a pretty much stock Ubuntu 20.04 beta with bash 5.0.16(1)-release.

3. What did you see instead?

```
$ tiup

Command 'tiup' not found, did you mean:

  command 'tilp' from deb tilp2 (1.18-3)
  command 'tup' from deb tup (0.7.8-3)

Try: sudo apt install <deb name>
```

4. What version of TiUP are you using (`tiup --version`)?

```
tiup version v0.0.3 tiup
```

lonng:
@wwar Thanks for your feedback.
@lucklove PTAL

PR #578, opened 2020-07-06
Introduce a more general way to config profile

lucklove:
Fix https://github.com/pingcap/tiup/issues/104


install.sh
```diff
@@ -52,13 +52,18 @@ chmod 755 "$bin_dir/tiup"
...
 
-echo "Detected shell: ${bold}$SHELL${sgr0}"
-case $SHELL in
-    *bash*) PROFILE=$HOME/.bash_profile;;
-     *zsh*) PROFILE=$HOME/.zshrc;;
-         *) PROFILE=$HOME/.profile;;
-esac
-
+# Refrence: https://stackoverflow.com/questions/14637979/how-to-permanently-set-path-on-linux-unix
+shell=$(echo $SHELL | awk 'BEGIN {FS="/";} { print $NF }')
```
lonng, line 56: We can add an integration test, which runs the `install.sh` scripts and uses `which tiup` to check the result.

.github/workflows/install.yaml
```diff
@@ -0,0 +1,26 @@
...
+              uses: actions/checkout@v2
+              with:
+                ref: ${{ github.event.pull_request.head.sha  }}
+                path: go/src/github.com/${{ github.repository  }}
+            - name: Install And Check
+              working-directory: ${{env.working-directory}}
+              run: |
+                sh install.sh
+                source ~/.profile
+                which tiup || (echo "no tiup found" && exit 1)
```
lonng, line 26: Should add `tiup mirror clone && local_install.sh` also.

PR #579, opened 2020-07-07
Always match the newest store when matching by address

july2993:
logically, only the store id identify a store.
if normally scale-in a store and scale-out using the same ip:port,
then there's will not only one store with the same address.

before this pr, only some API will match the newest store, this pr make sure all API match the newest store when matching by address.

also, this pr fix the tidbSpec initialize which we must make sure
initiaize the profile dir first.


pkg/cluster/spec/util.go
```diff
@@ -48,11 +47,12 @@ type ClusterMeta struct {
...
-// newTiDBSpec create a Spec for tidb cluster.
-func newTiDBSpec() *meta.SpecManager {
+func initTidbSpec() {
+	if !initialized {
+		panic("must Initialize profile first")
+	}
 	clusterBaseDir := filepath.Join(profileDir, TiOpsClusterDir)
-	clusterSpec := meta.NewSpec(clusterBaseDir)
-	return clusterSpec
+	tidbSpec = meta.NewSpec(clusterBaseDir)
```
lonng, line 55: How about putting `tidbSpec = meta.NewSpec(clusterBaseDir)` into `Initialize` function directly?

july2993: https://github.com/pingcap/tiup/pull/579/commits/fb62996f2a542e5f63b9a9a029e46d0f5a84efe7

PR #585, opened 2020-07-08
cluster: properly pass --wait-timeout to systemd operations

AstroProfundis:
The systemd module uses a hard-coded 100s timeout and ignores the `--wait-timeout` global argument.

Properly pass the value of `--wait-timeout` to systemd module, so that the SSH execute timeout of `systemctl` operations has the timeout specified with `--wait-timeout`.

### Check List 

Tests 

 - Integration test

Code changes

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/components/cluster/command/root.go
+++ b/components/cluster/command/root.go
@@ -119,7 +119,9 @@ func init() {
 	cliutil.BeautifyCobraUsageAndHelp(rootCmd)
 
 	rootCmd.PersistentFlags().Int64Var(&gOpt.SSHTimeout, "ssh-timeout", 5, "Timeout in seconds to connect host via SSH, ignored for operations that don't need an SSH connection.")
-	rootCmd.PersistentFlags().Int64Var(&gOpt.OptTimeout, "wait-timeout", 60, "Timeout in seconds to wait for an operation to complete, ignored for operations that don't fit.")
+	// the value of wait-timeout is also used for `systemctl` commands, as the default timeout of systemd for
+	// start/stop operations is 90s, the default value of this argument is better be longer than that
+	rootCmd.PersistentFlags().Int64Var(&gOpt.OptTimeout, "wait-timeout", 120, "Timeout in seconds to wait for an operation to complete, ignored for operations that don't fit.")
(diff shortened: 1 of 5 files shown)


AstroProfundis:
The default value of `wait-timeout` is 300s in the command, it's far more larger than systemd's default 90s, and is also used in other SSH operations.

The point of this PR is to make `systemctl` operations follow that value as other SSH operations do, in this way user can set any value as they want. I think 300s is a long enough default timeout.

july2993:
> The default value of `wait-timeout` is 300s in the command, it's far more larger than systemd's default 90s, and is also used in other SSH operations.
> 
> The point of this PR is to make `systemctl` operations follow that value as other SSH operations do, in this way user can set any value as they want. I think 300s is a long enough default timeout.

i

july2993:
> The default value of `wait-timeout` is 300s in the command, it's far more larger than systemd's default 90s, and is also used in other SSH operations.
> 
> The point of this PR is to make `systemctl` operations follow that value as other SSH operations do, in this way user can set any value as they want. I think 300s is a long enough default timeout.

I think it's 60s now?
```
components/cluster/command/root.go
122:    rootCmd.PersistentFlags().Int64Var(&gOpt.OptTimeout, "wait-timeout", 60, "Timeout in seconds to wait for an operation to complete, ignored for operations that don't fit.")
```

AstroProfundis:
I've updated it to 120s

components/cluster/command/root.go
```diff
@@ -119,7 +119,7 @@ func init() {
 	cliutil.BeautifyCobraUsageAndHelp(rootCmd)
 
 	rootCmd.PersistentFlags().Int64Var(&gOpt.SSHTimeout, "ssh-timeout", 5, "Timeout in seconds to connect host via SSH, ignored for operations that don't need an SSH connection.")
-	rootCmd.PersistentFlags().Int64Var(&gOpt.OptTimeout, "wait-timeout", 60, "Timeout in seconds to wait for an operation to complete, ignored for operations that don't fit.")
+	rootCmd.PersistentFlags().Int64Var(&gOpt.OptTimeout, "wait-timeout", 120, "Timeout in seconds to wait for an operation to complete, ignored for operations that don't fit.")
```
july2993, line 122: pls add a comment here for why should better > 90s ...

PR #587, opened 2020-07-08
Support deploy and manage dm 2.0

### LINKED ISSUE #461: Refactor & decouple cluster/dms components
## Feature Request

Refactor & decouple cluster/dms components

**Workload assessment:**
3d

july2993:
ref #557 

close #461 

This PR refactor some code and make tiup-dm support all commands like tiup-cluster(except for `check` and `import`)
- code ` components/cluster/command/*` are move into `pkg/cluster/deploy/` and change it to deploy a general cluster instead of tidb cluster, both tiup-cluster and tiup-dm use it to avoid many duplicate codes, and maybe later for some tiup-** to use it too.
- move `pkg/dm/` into `components/dm`
- many code in pkg/cluster/* are change to support work with a general cluster, I prefer to move code `tiup-cluser` specified code into component/cluster later.
- note tiup-dm does not support monitor yet in this pr.

all the existing command are supported:
```
Deploy a DM cluster for production

Usage:
  tiup-dm [flags]
  tiup-dm [command]

Available Commands:
  deploy      Deploy a DM cluster for production
  start       Start a DM cluster
  stop        Stop a DM cluster
  restart     Restart a DM cluster
  list        List all clusters
  destroy     Destroy a specified DM cluster
  audit       Show audit log of cluster operation
  exec        Run shell command on host in the dm cluster
  edit-config Edit DM cluster config
  display     Display information of a DM cluster
  reload      Reload a DM cluster's config and restart if needed
  upgrade     Upgrade a specified DM cluster
  patch       Replace the remote package with a specified package and restart the service
  scale-out   Scale out a DM cluster
  scale-in    Scale in a DM cluster
  help        Help about any command

Flags:
  -h, --help               help for tiup-dm
  -v, --version            version for tiup-dm
      --wait-timeout int   Timeout in seconds to wait for an operation to complete, ignored for operations that don't fit. (default 60)

Use "tiup-dm help [command]" for more information about a command.
```

a cluster may look like:
```
root@control:/tiup-cluster# tdm display test
dm Cluster: test
dm Version: nightly
ID                 Role       Host          Ports      OS/Arch       Status     Data Dir                   Deploy Dir
--                 ----       ----          -----      -------       ------     --------                   ----------
172.19.0.101:8261  dm-master  172.19.0.101  8261/8291  linux/x86_64  Healthy    data                       deploy/dm-master-8261
172.19.0.101:8361  dm-master  172.19.0.101  8361/8292  linux/x86_64  Unhealthy  data                       deploy/dm-master-8361
172.19.0.102:8261  dm-master  172.19.0.102  8261/8291  linux/x86_64  Healthy    data                       deploy/dm-master-8261
172.19.0.103:8261  dm-master  172.19.0.103  8261/8291  linux/x86_64  Healthy    /home/tidb/my_master_data  deploy/dm-master-8261
172.19.0.104:8261  dm-master  172.19.0.104  8261/8291  linux/x86_64  Healthy|L  data                       deploy/dm-master-8261
172.19.0.105:8261  dm-master  172.19.0.105  8261/8291  linux/x86_64  Healthy    data                       deploy/dm-master-8261
172.19.0.101:8280  dm-portal  172.19.0.101  8280       linux/x86_64  Up         data                       deploy/dm-portal-8280
172.19.0.102:8280  dm-portal  172.19.0.102  8280       linux/x86_64  Up         data                       deploy/dm-portal-8280
172.19.0.101:8262  dm-worker  172.19.0.101  8262       linux/x86_64  Free       data                       deploy/dm-worker-8262
172.19.0.101:8263  dm-worker  172.19.0.101  8263       linux/x86_64  Free       data                       deploy/dm-worker-8263
172.19.0.102:8262  dm-worker  172.19.0.102  8262       linux/x86_64  Free       data                       deploy/dm-worker-8262
172.19.0.103:8262  dm-worker  172.19.0.103  8262       linux/x86_64  Free       data                       deploy/dm-worker-8262
172.19.0.104:8262  dm-worker  172.19.0.104  8262       linux/x86_64  Free       data                       deploy/dm-worker-8262
```

tests:
- manual run every command (except patch) will add integration test in follow up pr.


components/cluster/command/check.go
```diff
@@ -63,8 +62,6 @@ conflict checks with other clusters`,
 				return cmd.Help()
 			}
 
-			logger.EnableAuditLog()
```
lonng, line 66: Why delete this line? We should record the audit log if user use command `tiup cluster check --apply` because `--apply` will change the system configuration.

components/cluster/command/deploy.go
```diff
@@ -55,30 +43,9 @@ var (
...
-		os   string // operating system
-		arch string // cpu architecture
-		// vendor string
-	}
-)
-
 func newDeploy() *cobra.Command {
-	opt := deployOptions{
-		identityFile: path.Join(tiuputils.UserHome(), ".ssh", "id_rsa"),
+	opt := deploy.DeployOptions{
```
lonng, line 47: Prefer to rename the `DeployOptions` to `Options` according to Golang idiomatic.

components/cluster/command/check.go
july2993: set in root.go `PersistentPreRunE` instead of every command setting this.

components/cluster/command/deploy.go
```diff
@@ -94,360 +61,46 @@ func newDeploy() *cobra.Command {
...
-				inst.Arch(),
+			return deployer.Deploy(
+				clusterName,
 				version,
-				"", // use default srcPath
-				inst.GetHost(),
-				deployDir,
+				topoFile,
+				opt,
+				afterDeployForDeployCluster,
```
AstroProfundis, line 79: It's better use "post" instead of "after", and as this is an unexported function, I think it can be with some simpler name like `postDeployActions` or `postDeployHooks` or similar thing.

components/cluster/command/deploy.go
```diff
@@ -94,360 +61,46 @@ func newDeploy() *cobra.Command {
 				return nil
 			}
 
-			logger.EnableAuditLog()
```
lonng, line 97: Is it by design?

components/cluster/command/exec.go
```diff
@@ -14,27 +14,12 @@
...
-	"github.com/fatih/color"
-	"github.com/joomcode/errorx"
-	perrs "github.com/pingcap/errors"
-	"github.com/pingcap/tiup/pkg/cluster/spec"
-	"github.com/pingcap/tiup/pkg/cluster/task"
-	"github.com/pingcap/tiup/pkg/logger"
-	"github.com/pingcap/tiup/pkg/logger/log"
-	"github.com/pingcap/tiup/pkg/meta"
-	"github.com/pingcap/tiup/pkg/set"
+	"github.com/pingcap/tiup/pkg/cluster/deploy"
```
AstroProfundis, line 17: I think `deploy` is not a proper package name, how about using something like `cmd` or `subcmd` or just leave them under `pkg/cluster`?

components/cluster/command/exec.go
```diff
@@ -46,87 +31,12 @@ func newExecCmd() *cobra.Command {
...
-				if len(stdout) > 0 {
-					log.Infof("%s:\n%s", color.GreenString("stdout"), stdout)
-				}
-				if len(stderr) > 0 {
-					log.Infof("%s:\n%s", color.RedString("stderr"), stderr)
-				}
-			}
-
-			return nil
+			return deployer.Exec(clusterName, opt, gOpt)
```
AstroProfundis, line 34: Also, `deployer` is not a proper name for operations other than `deploy` and `scale-out`.

components/cluster/command/deploy.go
lonng: Ignore this comment.

components/cluster/command/root.go
```diff
@@ -50,12 +50,12 @@ var (
 	skipConfirm bool
 )
 
-var tidbSpec *meta.SpecManager
+var tidbSpec *spec.SpecManager
+var deployer *deploy.Deployer
```
AstroProfundis, line 54: ditto

components/cluster/command/root.go
```diff
@@ -50,12 +50,12 @@ var (
...
 
 func scrubClusterName(n string) string {
 	return "cluster_" + telemetry.HashReport(n)
 }
-
```
AstroProfundis, line 58: keep the empty line for better formatting... BTW, I think `go fmt` should add the empty line automatically?

components/cluster/command/root.go
```diff
@@ -93,6 +93,8 @@ func init() {
 			}
 
 			tidbSpec = spec.GetSpecManager()
+			deployer = deploy.NewDeployer("tidb", tidbSpec)
```
AstroProfundis, line 96: ditto, something like `NewCommandSet` or `NewSubcommands` or so would be better IMO.

components/cluster/command/root.go
```diff
@@ -204,6 +206,7 @@ func extractSuggestionFromErrorX(err *errorx.Error) string {
 
 // Execute executes the root command
 func Execute() {
+
```
AstroProfundis, line 209: I believe this empty line is unnecessary.

components/cluster/command/scale_out.go
```diff
@@ -169,178 +77,11 @@ func convertStepDisplaysToTasks(t []*task.StepDisplay) []task.Task {
...
-	if hasImported {
-		if err := spec.HandleImportPathMigration(clusterName); err != nil {
-			return task.NewBuilder().Build(), err
-		}
-	}
+func final(builder *task.Builder, name string, meta spec.Metadata) {
+	builder.UpdateTopology(name, meta.(*spec.ClusterMeta), nil)
+}
 
+func afterDeploy(builder *task.Builder, newPart spec.Topology) {
```
AstroProfundis, line 84: ditto

components/cluster/command/deploy.go
july2993: set in root.go PersistentPreRunE instead of every command setting this.

pkg/cluster/deploy/deploy.go
```diff
@@ -0,0 +1,1803 @@
...
+	deployCompTasks = append(deployCompTasks, dpTasks...)
+
+	builder := task.NewBuilder().
+		Step("+ Generate SSH keys",
+			task.NewBuilder().SSHKeyGen(d.specManager.Path(clusterName, "ssh", "id_rsa")).Build()).
+		ParallelStep("+ Download TiDB components", downloadCompTasks...).
+		ParallelStep("+ Initialize target host environments", envInitTasks...).
+		ParallelStep("+ Copy files", deployCompTasks...)
+
+	if afterDeploy != nil {
```
AstroProfundis, line 957: ditto

pkg/cluster/deploy/deploy.go
```diff
@@ -0,0 +1,1803 @@
...
+		return nil, perrs.AddStack(err)
+	}
+
+	if !exist {
+		return nil, perrs.Errorf("cluster `%s` not exists", name)
+	}
+
+	metadata = d.specManager.NewMetadata()
+	err = d.specManager.Metadata(name, metadata)
+	if err != nil && !errors.Is(perrs.Cause(err), meta.ErrValidate) &&
```
AstroProfundis, line 1206: This error should be checked by the caller, sub commands like `deploy`, `scale-out`, `reload`, `upgrade` and `restart` should not allow any topology validation errors, however, sub commands like `scale-in`, `destroy`, `display` and `edit-config` should ignore topology validation errors to make it possible to fix a broken topology.

pkg/cluster/spec/instance.go
```diff
@@ -56,14 +56,21 @@ type Component interface {
 	Instances() []Instance
 }
 
+// RollingUpdateInstance represent a instance need to transfer state when restart.
+// e.g transfer leader.
+type RollingUpdateInstance interface {
+	BeforeRestart(topo Topology, apiTimeoutSeconds int) error
```
AstroProfundis, line 62: Better use `PreRestart` and `PostRestart`

components/cluster/command/start.go
```diff
@@ -63,31 +47,8 @@ func newStartCmd() *cobra.Command {
...
-			// FIXME: Map possible task errors and give suggestions.
-			return err
-		}
-		return perrs.Trace(err)
-	}
-
-	log.Infof("Started cluster `%s` successfully", clusterName)
-
-	return nil
+	return deployer.StartCluster(clusterName, options, func(b *task.Builder, metadata spec.Metadata) {
```
lonng, line 50: Most of the logics have been removed from `startCluster`, I think we can inline this statement into `RunE` directly.

components/dm/command/deploy.go
```diff
@@ -14,54 +14,18 @@
...
-	"fmt"
-	"os"
 	"path"
-	"path/filepath"
-	"strings"
 
-	cspec "github.com/pingcap/tiup/pkg/cluster/spec"
-	"github.com/pingcap/tiup/pkg/dm/spec"
-	"github.com/pingcap/tiup/pkg/meta"
+	"github.com/pingcap/tiup/pkg/cluster/deploy"
```
lonng, line 20: Delete the blank line.

components/dm/command/destroy.go
```diff
@@ -13,24 +13,12 @@
...
-	"github.com/pingcap/tiup/pkg/cluster/meta"
 	operator "github.com/pingcap/tiup/pkg/cluster/operation"
-	"github.com/pingcap/tiup/pkg/cluster/task"
-	"github.com/pingcap/tiup/pkg/logger"
-	"github.com/pingcap/tiup/pkg/logger/log"
-	tiuputils "github.com/pingcap/tiup/pkg/utils"
 	"github.com/spf13/cobra"
 )
 
+// TODO support retain data like cluster?
```
lonng, line 21: I am not sure is there the same requirement.

components/dm/spec/logic.go
```diff
@@ -326,37 +327,36 @@ func (i *DMMasterInstance) InitConfig(e executor.Executor, clusterName, clusterV
...
-			paths.Data[0],
-			paths.Log,
-		).WithPort(spec.Port).WithNumaNode(spec.NumaNode).WithPeerPort(spec.PeerPort).AppendEndpoints(c.Endpoints(deployUser)...)
-
-		fp := filepath.Join(paths.Cache, fmt.Sprintf("run_dm-master_%s_%d.sh", i.GetHost(), i.GetPort()))
-		log.Infof("script path: %s", fp)
-		if err := cfg.ConfigToFile(fp); err != nil {
-			return err
-		}
+	c := topo.(*DMTopologySpecification)
```
lonng, line 335: Can we simplify the name `DMTopologySpecification`? Or just `Specification` or `Topology` is enough? The two specification definitions are placed in a different package, I think it's ok to use the same type name.

components/dm/spec/topology_dm.go
```diff
@@ -86,9 +86,9 @@ type (
 
 	// DMTopologySpecification represents the specification of topology.yaml
 	DMTopologySpecification struct {
-		GlobalOptions GlobalOptions `yaml:"global,omitempty"`
-		// MonitoredOptions MonitoredOptions   `yaml:"monitored,omitempty"`
-		ServerConfigs DMServerConfigs    `yaml:"server_configs,omitempty"`
+		GlobalOptions GlobalOptions `yaml:"global,omitempty" validate:"global:editable"`
+		// MonitoredOptions MonitoredOptions   `yaml:"monitored,omitempty" validate:"monitored:editable"`
```
lonng, line 90: Don't we need the monitor options?

pkg/cliutil/prepare/prepare.go
```diff
@@ -36,36 +35,48 @@ var (
...
-		{dirKind: "log directory", accessor: func(instance spec.Instance, topo *spec.Specification) string { return instance.LogDir() }},
+		{dirKind: "deploy directory", accessor: func(instance spec.Instance, topo spec.Topology) string { return instance.DeployDir() }},
+		{dirKind: "data directory", accessor: func(instance spec.Instance, topo spec.Topology) string { return instance.DataDir() }},
+		{dirKind: "log directory", accessor: func(instance spec.Instance, topo spec.Topology) string { return instance.LogDir() }},
 	}
 	hostDirAccessor := []DirAccessor{
-		{dirKind: "monitor deploy directory", accessor: func(instance spec.Instance, topo *spec.Specification) string {
-			return topo.MonitoredOptions.DeployDir
+		{dirKind: "monitor deploy directory", accessor: func(instance spec.Instance, topo spec.Topology) string {
+			m := topo.BaseTopo().MonitoredOptions
```
lonng, line 61: Monitor options shouldn't be missing.

pkg/cliutil/prepare/prepare.go
```diff
@@ -89,16 +100,18 @@ func CheckClusterDirConflict(clusterSpec *meta.SpecManager, clusterName string,
...
 		err := clusterSpec.Metadata(name, metadata)
 		if err != nil {
 			return errors.Trace(err)
 		}
 
-		f := fixDir(metadata.Topology)
-		metadata.Topology.IterInstance(func(inst spec.Instance) {
+		topo := metadata.GetTopology()
+
+		f := fixDir(metadata.GetTopology())
```
lonng, line 111: Can we reuse the result of L109.

pkg/cliutil/prepare/prepare.go
```diff
@@ -210,6 +224,7 @@ func CheckClusterPortConflict(clusterSpec *meta.SpecManager, clusterName string,
 			continue
 		}
 
+		// TODO
```
lonng, line 227: Todo what?

pkg/cluster/audit/audit.go
```diff
@@ -0,0 +1,99 @@
+package audit
```
lonng, line 1: The license header is missing.

pkg/logger/audit.go
```diff
@@ -15,25 +15,24 @@ package logger
...
-	"io/ioutil"
 	"os"
 	"strings"
-	"time"
 
 	utils2 "github.com/pingcap/tiup/pkg/utils"
 
-	"github.com/pingcap/tiup/pkg/base52"
-	"github.com/pingcap/tiup/pkg/cluster/spec"
+	"github.com/pingcap/tiup/pkg/cluster/audit"
```
lonng, line 23: Delete the blank above.

pkg/cluster/task/update_dm_meta.go
```diff
@@ -13,65 +13,63 @@
...
+	dmspec "github.com/pingcap/tiup/components/dm/spec"
 
 	"github.com/pingcap/tiup/pkg/set"
 )
 
 // UpdateDMMeta is used to maintain the DM meta information
 type UpdateDMMeta struct {
 	cluster        string
-	metadata       *meta.DMMeta
+	metadata       *dmspec.DMMeta
```
lonng, line 28: Can we rename the `DMMeta` to `Meta` or `Metadata`?

pkg/cluster/task/scale_config.go
```diff
@@ -23,10 +23,11 @@ import (
 
 // ScaleConfig is used to copy all configurations to the target directory of path
 type ScaleConfig struct {
+	specManager    *spec.SpecManager
```
lonng, line 26: I don't think the `SpecManager` should belong to the cluster spec package. It's better to move it into an independent package.

pkg/cluster/operation/upgrade.go
```diff
@@ -28,91 +25,48 @@ import (
...
 
-					if err := stopInstance(getter, instance, options.OptTimeout); err != nil {
-						return errors.Annotatef(err, "failed to stop %s", instance.GetHost())
-					}
-					if err := startInstance(getter, instance, options.OptTimeout); err != nil {
-						return errors.Annotatef(err, "failed to start %s", instance.GetHost())
-					}
-				}
+			if !options.Force {
+				rollingInstance, _ = instance.(spec.RollingUpdateInstance)
```
lonng, line 50: It's better to check the assertion result instead of check the nil value of `rollingInstance`.

pkg/cluster/deploy/deploy.go
```diff
@@ -0,0 +1,1803 @@
...
+
+	topo := metadata.GetTopology()
+	base := metadata.GetBaseMeta()
+
+	b := task.NewBuilder().
+		SSHKeySet(
+			d.specManager.Path(name, "ssh", "id_rsa"),
+			d.specManager.Path(name, "ssh", "id_rsa.pub")).
+		ClusterSSH(topo, base.User, options.SSHTimeout).
+		Serial(task.NewFunc("StartCluster", func(ctx *task.Context) error {
```
lonng, line 73: Why not use `Builder.Func()` directly?

pkg/cluster/deploy/deploy.go
```diff
@@ -0,0 +1,1803 @@
...
+			return operator.Start(ctx, topo, options)
+		}))
+
+	if len(fn) > 0 {
+		if len(fn) != 1 {
```
lonng, line 78: Can we support multiple callback functions?

pkg/cluster/deploy/deploy.go
```diff
@@ -0,0 +1,1803 @@
...
+
+	topo := metadata.GetTopology()
+	base := metadata.GetBaseMeta()
+
+	t := task.NewBuilder().
+		SSHKeySet(
+			d.specManager.Path(clusterName, "ssh", "id_rsa"),
+			d.specManager.Path(clusterName, "ssh", "id_rsa.pub")).
+		ClusterSSH(metadata.GetTopology(), base.User, options.SSHTimeout).
+		Serial(task.NewFunc("StopCluster", func(ctx *task.Context) error {
```
lonng, line 113: ditto

pkg/cluster/deploy/deploy.go
```diff
@@ -0,0 +1,1803 @@
...
+
+	topo := metadata.GetTopology()
+	base := metadata.GetBaseMeta()
+
+	t := task.NewBuilder().
+		SSHKeySet(
+			d.specManager.Path(clusterName, "ssh", "id_rsa"),
+			d.specManager.Path(clusterName, "ssh", "id_rsa.pub")).
+		ClusterSSH(topo, base.User, options.SSHTimeout).
+		Serial(task.NewFunc("RestartCluster", func(ctx *task.Context) error {
```
lonng, line 145: ditto

pkg/cluster/deploy/deploy.go
```diff
@@ -0,0 +1,1803 @@
...
+		}
+		log.Infof("Destroying cluster...")
+	}
+
+	t := task.NewBuilder().
+		SSHKeySet(
+			d.specManager.Path(clusterName, "ssh", "id_rsa"),
+			d.specManager.Path(clusterName, "ssh", "id_rsa.pub")).
+		ClusterSSH(topo, base.User, gOpt.SSHTimeout).
+		Serial(task.NewFunc("StopCluster", func(ctx *task.Context) error {
```
lonng, line 221: ditto

pkg/cluster/deploy/deploy.go
```diff
@@ -0,0 +1,1803 @@
...
+			return operator.Stop(ctx, topo, operator.Options{})
+		})).
+		Serial(task.NewFunc("DestroyCluster", func(ctx *task.Context) error {
```
lonng, line 224: ditto

pkg/cluster/deploy/deploy.go
```diff
@@ -0,0 +1,1803 @@
...
+	}
+
+	t := task.NewBuilder().
+		SSHKeySet(
+			d.specManager.Path(clusterName, "ssh", "id_rsa"),
+			d.specManager.Path(clusterName, "ssh", "id_rsa.pub")).
+		ClusterSSH(topo, base.User, opt.SSHTimeout).
+		Parallel(refreshConfigTasks...).
+		Serial(task.NewFunc("UpgradeCluster", func(ctx *task.Context) error {
+			return operator.Upgrade(ctx, topo, opt)
```
lonng, line 521: https://github.com/pingcap/tiup/pull/625

july2993: https://github.com/pingcap/tiup/pull/587/commits/5432d87deb4517edf332bf305c43144289df840e

pkg/cliutil/prepare/prepare.go
july2993: can not use the tidb cluster specified meta data
https://github.com/pingcap/tiup/pull/587/commits/f09b8b9491cebd0d73a25f0ecee68e857fff5b3d

pkg/cliutil/prepare/prepare.go
july2993: This `MonitoredOptions` control thinks about node/backbox export, I don't think this is a *must* for deploying a cluster although tiup-cluster always having this now.

components/dm/spec/topology_dm.go
july2993: dm-ansible will not deploy exporter, I will add it if we do want this.
Also, I think it's not a good idea to forcing deploy it.

pkg/cluster/deploy/deploy.go
july2993: https://github.com/pingcap/tiup/pull/587/commits/51854efcf95ca02cc359d9ce4f19846edbc904a9

pkg/cluster/task/update_dm_meta.go
july2993: https://github.com/pingcap/tiup/pull/587/commits/9c6aae46dd607af798ffbf42e2e465a7b6c114b3

components/cluster/command/root.go
july2993: what about put it in pkg/cluster as `Manager`?

july2993: https://github.com/pingcap/tiup/pull/587/commits/db9083bee92acb9fac22975076962e62208e006b

components/cluster/command/scale_out.go
```diff
@@ -169,178 +77,11 @@ func convertStepDisplaysToTasks(t []*task.StepDisplay) []task.Task {
...
-	if hasImported {
-		if err := spec.HandleImportPathMigration(clusterName); err != nil {
-			return task.NewBuilder().Build(), err
-		}
-	}
+func final(builder *task.Builder, name string, meta spec.Metadata) {
+	builder.UpdateTopology(name, meta.(*spec.ClusterMeta), nil)
+}
 
+func scaleOutPostDeployHook(builder *task.Builder, newPart spec.Topology) {
```
AstroProfundis, line 84: `postScaleOutHook`

pkg/cluster/manager.go
```diff
@@ -0,0 +1,1875 @@
...
+		return nil, perrs.AddStack(err)
+	}
+
+	if !exist {
+		return nil, perrs.Errorf("cluster `%s` not exists", name)
+	}
+
+	metadata = d.specManager.NewMetadata()
+	err = d.specManager.Metadata(name, metadata)
+	if err != nil && !errors.Is(perrs.Cause(err), meta.ErrValidate) &&
```
AstroProfundis, line 1229: This error should be returned to the caller to decide whither ignore validation error or not, as my previous comment said...

components/dm/command/destroy.go
july2993: that's why I  left "?", but it will easy to support it, just add the flash like cluster.

pkg/cluster/task/scale_config.go
july2993: will do this in later PR

components/cluster/command/scale_out.go
july2993: resolve by https://github.com/pingcap/tiup/pull/587/commits/3e2f36aeec87da31c2b8949542ab9eb283e1c6db

pkg/cluster/manager.go
july2993: Reading the metadata should not failed by the Validate, the user still can call Validate explicitly if need like L1152 in cluster/manager.go.

It's a bad idea to do the check like this in #531 at many places
```
			if err != nil && !errors.Is(perrs.Cause(err), meta.ErrValidate) &&
				!errors.Is(perrs.Cause(err), spec.ErrNoTiSparkMaster) {
				return perrs.AddStack(err)
			}
```

I think it will better to let the user validate explicitly if they need instead of returning the error and you may need to ignore it.

july2993:
@lonng @AstroProfundis PTAL

components/dm/command/scale_in.go
```diff
@@ -43,18 +33,24 @@ func newScaleInCmd() *cobra.Command {
...
-					"This operation will delete the %s nodes in `%s` and all their data.\nDo you want to continue? [y/N]:",
-					strings.Join(gOpt.Nodes, ","),
-					color.HiYellowString(clusterName)); err != nil {
-					return err
-				}
-				log.Infof("Scale-in nodes...")
+
+			scale := func(b *task.Builder, imetadata spec.Metadata) {
+				metadata := imetadata.(*dm.Metadata)
+				b.Serial(
```
lonng, line 39: Use `Builder.Func`

pkg/cluster/deploy/deploy.go
july2993: dup of https://github.com/pingcap/tiup/pull/587#discussion_r460887981

components/dm/command/scale_in.go
july2993: https://github.com/pingcap/tiup/pull/587/commits/5234b953149ff9b0421e866f3693e48c2957817e

pkg/cluster/manager.go
```diff
@@ -0,0 +1,1886 @@
+package cluster
```
lonng, line 1: The license header is missing.

pkg/cluster/manager.go
```diff
@@ -0,0 +1,1886 @@
...
+}
+
+func (d *Manager) meta(name string) (metadata spec.Metadata, err error) {
+	exist, err := d.specManager.Exist(name)
+	if err != nil {
+		return nil, perrs.AddStack(err)
+	}
+
+	if !exist {
+		return nil, perrs.Errorf("cluster `%s` not exists", name)
```
lonng, line 1230:
```suggestion
		return nil, perrs.Errorf("%s cluster `%s` not exists", d.sysName, name)
```

pkg/cluster/manager.go
```diff
@@ -0,0 +1,1886 @@
...
+		}
+		return perrs.Trace(err)
+	}
+
+	log.Infof("Scaled cluster `%s` out successfully", clusterName)
+
+	return nil
+}
+
+func (d *Manager) meta(name string) (metadata spec.Metadata, err error) {
```
lonng, line 1223: Prefer to rename all method's receiver of `*Manager`.

july2993: resolve by https://github.com/pingcap/tiup/pull/587/commits/06db827e3b1d4e01a2bc61cafa09a15561a0db58

pkg/cluster/manager.go
```diff
@@ -0,0 +1,1886 @@
...
+		return perrs.AddStack(err)
+	}
+
+	topo := metadata.GetTopology()
+	base := metadata.GetBaseMeta()
+
+	// display cluster meta
+	cyan := color.New(color.FgCyan, color.Bold)
+	fmt.Printf("%s Cluster: %s\n", d.sysName, cyan.Sprint(clusterName))
+	fmt.Printf("%s Version: %s\n", d.sysName, cyan.Sprint(base.Version))
```
lonng: The previous implementation will display the dashboard address, I think we should keep it.

july2993: it remain the origin logic in ~/code/tiup/components/cluster/command/display.go

pkg/cluster/manager.go
july2993: https://github.com/pingcap/tiup/pull/587/commits/5d1ad886de4c930d5cd279b770e3bf3796403ff7

pkg/cluster/manager.go
```diff
@@ -0,0 +1,1886 @@
...
+func (d *Manager) Deploy(
+	clusterName string,
+	clusterVersion string,
+	topoFile string,
+	opt DeployOptions,
+	afterDeploy func(b *task.Builder, newPart spec.Topology),
+	skipConfirm bool,
+	optTimeout int64,
+	sshTimeout int64,
+	nativeSSH bool,
```
lonng, line 807: All the last four options are contained in `operator.Options`, why not pass the `options` directly?

pkg/cluster/manager.go
AstroProfundis: There is no "user" of this function.

My point is that the meta.yaml may be broken or with illegal content, due to no matter what reason, and we need to keep there's still some possibility for user to correct that with commands like `display`, `scale-in`, `edit-config` and `destroy`.

july2993: > There is no "user" of this function.

I mean the caller of `Manager.meta` here.
So no need to check for the commands you said here.
and no need to add ignoring one more error type fore the commands when calling this function when you add one more
error type in the validate process during unmarshaling the file.

AstroProfundis: There will always be some checks, if not for those ignoring some errors, then must be for those not ignoring errors.

Our sub commands treat the validation error differently, I'm not feeling good with the idea of ignoring some errors and recheck outside, I'm more comfortable with throwing errors to upper level and let the upper level decide what to do.

july2993: You can treat `meta()` here as an upper level of Unmarshaling here.
the caller will not need to care about what error type to ignore.

pkg/cluster/manager.go
july2993: `skipConfirm` is not.
the good thing for keeping the left 3 params is making it clear what this function needs since `operator.Options` contains many unrelate params.

I prefer to just keep it now.

pkg/cluster/manager.go
AstroProfundis: In `Upgrade()` the `m.meta()` is called to get the metadata of cluster, this is for example a case that validation error should not be ignored. If the topology is incorrect, upgrading will very likely make things worse.

july2993: then you can Validate it in `Upgrade()`...

Note in master https://github.com/pingcap/tiup/blob/f1142b1c787065f3d4f2754a9b76ad26a0670f27/components/cluster/command/upgrade.go#L86, also v1.0.8

it always ignores the ErrValidate for the upgrade command(only ignore ErrValidate but not ErrNoTiSparkMaster?)

Your example just shows it's good to do it like now IMO.

AstroProfundis: `Upgrade()` is just an example and not necessarily to be the case. So my point is always that:
 - We need to treat validation error differently in different sub commands
 - The error is raised when unmarshaling the yaml file
 - There are two ways of doing so:
    1. Ignore all validation errors when unmarshaling the file, and do validation again for those sub commands where the validation errors should not be ignored
    2. Raise any errors occurs during unmarshaling the file, and ignore validation errors for those sub commands where the validation errors shoud be ignored

I think 2 is better because ignoring errors in `m.meta()` makes the upper level caller getting less information, this raises limitation of the caller to process against errors.

If your concerns is about repeating the `errors.Is()` boolean sentence too many times, It's also okay to use a dedicated function like `IsValidationError()` to wrap them, this would be no more lines than calling `Validate()` again as it is now in `ScaleOut()`.

july2993: `Unmarshal` should be `Unmarshal`, I think it will better to remove the `Validate` hack in `Unmarshal` but it will need more change.  It's hard to know you doing this int the `Unmarshal`.

What if you need just the truly `Unmarshal`?  (similar to the currently meta())

AstroProfundis: Unmarshal itself has formatting validation. I'm okay to split spec validation from unmarshal, but that will lead to a problem that the validation is not atomic anymore, we'll need some wrapping to read the topology, unmarshal it and then validate it.

Actually, I don't see any situation that we need to read the topology file but not validating it's content. The "validation errors" been ignored in some sub commands are not all errors the validation process may raise, but only some topology logic error we defined in our own code.

pkg/cliutil/prepare/prepare.go
```diff
@@ -36,36 +35,48 @@ var (
...
+			return clusterutil.Abs(topo.BaseTopo().GlobalOptions.User, dir)
 		}
 		return dir
 	}
 }
 
 // CheckClusterDirConflict checks cluster dir conflict
-func CheckClusterDirConflict(clusterSpec *meta.SpecManager, clusterName string, topo *spec.Specification) error {
+func CheckClusterDirConflict(clusterSpec *spec.SpecManager, clusterName string, topo spec.Topology) error {
 	type DirAccessor struct {
```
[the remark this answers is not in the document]
july2993: It's used only in this function, I think it's good to keep the limit scope.

components/dm/spec/topology_dm.go
```diff
@@ -263,8 +263,8 @@ func (s PortalSpec) IsImported() bool {
 }
 
 // UnmarshalYAML sets default values when unmarshaling the topology file
-func (topo *DMTopologySpecification) UnmarshalYAML(unmarshal func(interface{}) error) error {
-	type topology DMTopologySpecification
+func (topo *Topology) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	type topology Topology
```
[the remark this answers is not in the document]
july2993: Calling unmarshal(topo) directly will cause endless loop because it will Call Topology.UnmarshalYAML.
this is the same as the cluster one doing.

pkg/cluster/manager.go
```diff
@@ -0,0 +1,1899 @@
+// Copyright 2020 PingCAP, Inc.
```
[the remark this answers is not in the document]
july2993: I think the file is too long is not a problem if it contains the same `Struct` for doing one thing.
like https://github.com/golang/go/blob/master/src/net/http/server.go

components/dm/command/root.go
```diff
@@ -91,25 +109,26 @@ func init() {
...
+		newExecCmd(),
 		newEditConfigCmd(),
+		newDisplayCmd(),
 		newReloadCmd(),
+		newUpgradeCmd(),
 		newPatchCmd(),
+		newScaleOutCmd(),
+		newScaleInCmd(),
+	/*
+		newCheckCmd(),
```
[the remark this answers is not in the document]
july2993: No now, the origin code just copy from cluster.

pkg/cluster/spec/util.go
```diff
@@ -45,7 +43,49 @@ type ClusterMeta struct {
...
+	}
+}
+
+// AuditDir return the directory for saving audit log.
+func AuditDir() string {
+	return filepath.Join(profileDir, TiOpsAuditDir)
+}
+
 // SaveClusterMeta saves the cluster meta information to profile directory
+// TODO remove this
```
lucklove, line 88: It seems this is still referenced by other code. Why we need remove it?

july2993: https://github.com/pingcap/tiup/pull/587/commits/17ce8e0c3c3953522ddc03d32fda98199da3a93d
It will better to remote all ClusterMeta into components/cluster
at least not utils.go

pkg/cluster/manager.go
july2993: https://github.com/pingcap/tiup/pull/587/commits/8d8d3c8275aff976017302143e03d54e6c86204e
align with the master

july2993:
@lonng @AstroProfundis PTAL

Issue #555, opened 2020-06-30
It's better for playground to run component binary directly


lucklove:
## Feature Request

**Is your feature request related to a problem? Please describe:**

At this time, the command `tiup playground` has a long call `tiup -> playground -> tiup -> tidb/tikv/pd`. It's more graceful for playground to call tidb/tikv/pd directly.

For code reuse, we can import tiup as a library in the playground component and reuse the runComponent logic

PR #588, opened 2020-07-08
mini refactor playground to run component directly

lucklove:
Fix https://github.com/pingcap/tiup/issues/555

### DIFF
--- a/components/playground/instance/process.go
+++ b/components/playground/instance/process.go
@@ -9,8 +9,7 @@ import (
 
 	"github.com/pingcap/errors"
 	"github.com/pingcap/tiup/pkg/environment"
-	e "github.com/pingcap/tiup/pkg/exec"
-	"github.com/pingcap/tiup/pkg/localdata"
+	tiupexec "github.com/pingcap/tiup/pkg/exec"
 	"github.com/pingcap/tiup/pkg/repository/v0manifest"
 )
 
@@ -83,11 +82,7 @@ func NewComponentProcess(ctx context.Context, dir, binPath, component string, ve
 	}
 
 	env := environment.GlobalEnv()
-	oldDir := os.Getenv(localdata.EnvNameInstanceDataDir)
-	defer os.Setenv(localdata.EnvNameInstanceDataDir, oldDir)
-	os.Setenv(localdata.EnvNameInstanceDataDir, dir)
-
-	cmd, err := e.PrepareCommand(ctx, component, version, binPath, "", arg, env)
+	cmd, err := tiupexec.PrepareCommand(ctx, component, version, binPath, "", dir, arg, env)
 	if err != nil {
 		return nil, err
 	}
--- a/components/playground/instance/process.go
+++ b/components/playground/instance/process.go
@@ -13,10 +13,18 @@ import (
 	"github.com/pingcap/tiup/pkg/repository/v0manifest"
 )
 
+// ErrorWaitTimeout is used to represent timeout of a command
+// Example:
+//		_ = syscall.Kill(cmd.Process.Pid, syscall.SIGKILL)
+//		if err := WaitContext(context.WithTimeout(context.Background(), 3), cmd); err == ErrorWaitTimeout {
+//			// Do something
+//		}
+var ErrorWaitTimeout = errors.New("wait command timeout")
+
 // Process represent process to be run by playground
 type Process interface {
 	Start() error
-	Wait() error
+	Wait(ctx context.Context) error
 	Pid() int
 	Uptime() string
 	SetOutputFile(fname string) error
@@ -37,8 +45,8 @@ func (p *process) Start() error {
 }
 
 // Wait implements Instance interface.
-func (p *process) Wait() error {
-	return p.cmd.Wait()
+func (p *process) Wait(ctx context.Context) error {
+	return WaitContext(ctx, p.cmd)
 }
 
 // Pid implements Instance interface.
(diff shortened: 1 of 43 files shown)


lonng:
@lucklove Please set the proper labels and projects for this PR.

pkg/telemetry/meta.go
```diff
@@ -68,3 +73,16 @@ func (m *Meta) SaveTo(fname string) error {
 
 	return ioutil.WriteFile(fname, data, 0644)
 }
+
+// GetTelemetryMeta read the telemeta from disk
+func GetTelemetryMeta(env *environment.Environment) (meta *Meta, fname string, err error) {
```
lonng, line 78: Prefer to rename `GetTelemetryMeta` to `Meta` because the function is placed in the `telemetry` package.

pkg/telemetry/meta.go
```diff
@@ -3,12 +3,17 @@ package telemetry
 import (
 	"io/ioutil"
 	"os"
+	"path/filepath"
```
lonng, line 6: Please add license header for this file.

components/playground/instance/process.go
```diff
@@ -2,41 +2,54 @@ package instance
...
 	"context"
-	"fmt"
 	"io"
 	"os"
 	"os/exec"
 	"time"
 
 	"github.com/pingcap/errors"
+	"github.com/pingcap/tiup/pkg/environment"
+	e "github.com/pingcap/tiup/pkg/exec"
```
lonng, line 12: It's better to choose another better alias.

components/playground/instance/process.go
```diff
@@ -55,27 +68,34 @@ func (p *Process) setOutputFile(fname string) error {
...
-	)
+	env, err := environment.InitEnv(repository.Options{})
+	if err != nil {
+		return nil, err
+	}
 
-	cmd.Stdout = os.Stdout
-	cmd.Stderr = os.Stderr
+	oldDir := os.Getenv(localdata.EnvNameInstanceDataDir)
+	defer os.Setenv(localdata.EnvNameInstanceDataDir, oldDir)
```
lonng, line 92: I think the better way to pass the instance data directory is to use a string slice instead of set the os environment. It's a potential bug if we call `NewComponentProcess` parallelly in the future.

components/playground/playground.go
```diff
@@ -986,6 +917,103 @@ func (p *Playground) renderSDFile() error {
 	return nil
 }
 
+func (p *Playground) bootMonitor(ctx context.Context, monitorInfo *MonitorInfo) (*exec.Cmd, *grafana, error) {
+	options := p.bootOptions
+
+	// set up prometheus
+	if err := installIfMissing(p.profile, "prometheus", options.version); err != nil {
```
lonng, line 924: It's better to use `PreCommand` too.

lonng:
@lucklove 👍 Awesome works.

I think we can change the default version to `nightly` if the user doesn't specify any version. We use the installed latest stable version in the previous implementation, it's proved that there are some potential risks(e.g: different versions will be used between different components).

july2993:
> @lucklove 👍 Awesome works.
> 
> I think we can change the default version to `nightly` if the user doesn't specify any version. We use the installed latest stable version in the previous implementation, it's proved that there are some potential risks(e.g: different versions will be used between different components).

I don't agree with this, nightly is easier to be broken and it's not the same versions of different components actually.
we can fetch the newest stable version of tidb then run all components with the specified version.

lucklove:
> > @lucklove 👍 Awesome works.
> > I think we can change the default version to `nightly` if the user doesn't specify any version. We use the installed latest stable version in the previous implementation, it's proved that there are some potential risks(e.g: different versions will be used between different components).
> 
> I don't agree with this, nightly is easier to be broken and it's not the same versions of different components actually.
> we can fetch the newest stable version of tidb then run all components with the specified version.

Agree

components/playground/playground.go
lucklove: Prometheus and other components can use `PreCommand`, but grafana can't. It's not a standalone component (we can't run it with `tiup grafana`). Before running it, we need to copy and modify the *.json

components/playground/playground.go
```diff
@@ -603,15 +608,28 @@ func (p *Playground) addInstance(componentID string, cfg instance.Config) (ins i
...
 
-	if options.version != "nightly" && options.version != "" {
+	if options.version == "" {
+		version, _, err := env.V1Repository().LatestStableVersion("tidb")
+		if err != nil {
+			return err
+		}
+		options.version = version.String()
+
+		fmt.Println(color.YellowString(`Warning: no version specified for cluster. Will use the last stable one: %s
```
july2993, line 626: s/Warning: no version specified for cluster. Will use the last stable one/Use latest stable version/

I think this is not a warning, just a info to tell the user which version is run, we can use briefer description.

components/playground/playground.go
```diff
@@ -924,6 +859,17 @@ func (p *Playground) bootCluster(options *bootOptions) error {
 			if grafana != nil {
 				_ = syscall.Kill(grafana.cmd.Process.Pid, sig)
 			}
+		} else {
```
july2993, line 862: why SIGKILL? that means always SIGKILL when user press ctrl-c ?

can we gracefully stop it first, and use SIGKILL if user press again ctrl-c again or some reach some timeout.

components/playground/playground.go
lonng: Nice catch.

lucklove:
> I can not catch the point to run components directly actually.
> and do we need to change grafan and p8s to run directly too? it's still running by `tiup run...`
> 
> rest LGTM

If we don't run them directly, the signal handing will be complex and error-prone.
The grafana and prometheus should be run directly too.

PR #593, opened 2020-07-09
cluster: ignore not found error when deleting something via PD API

AstroProfundis:
When deleting some identity from the cluster via PD API, it's ok to ignore not found error, as that's what we want. This would make it easier to keep `scale-in` operation idempotent.

Check for PD API response and ignore not found errors.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Related changes

### DIFF
--- a/pkg/cluster/api/pdapi.go
+++ b/pkg/cluster/api/pdapi.go
@@ -19,6 +19,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"net/http"
 	"net/url"
 	"sort"
 	"time"
@@ -432,7 +433,7 @@ func (pc *PDClient) RemoveStoreEvict(host string) error {
 	_, err = tryURLs(endpoints, func(endpoint string) ([]byte, error) {
 		body, statusCode, err := pc.httpClient.Delete(endpoint, nil)
 		if err != nil {
-			if statusCode == 404 || bytes.Contains(body, []byte("scheduler not found")) {
+			if statusCode == http.StatusNotFound || bytes.Contains(body, []byte("scheduler not found")) {
 				log.Debugf("Store leader evicting scheduler does not exist, ignore.")
 				return body, nil
 			}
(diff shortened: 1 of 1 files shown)


pkg/cluster/api/pdapi.go
```diff
@@ -465,8 +465,16 @@ func (pc *PDClient) DelPD(name string, retryOpt *clusterutil.RetryOption) error
 	endpoints := pc.getEndpoints(cmd)
 
 	_, err = tryURLs(endpoints, func(endpoint string) ([]byte, error) {
-		body, _, err := pc.httpClient.Delete(endpoint, nil)
-		return body, err
+		body, statusCode, err := pc.httpClient.Delete(endpoint, nil)
+		if err != nil {
+			if statusCode == 404 || bytes.Contains(body, []byte("not found, pd")) {
```
lonng, line 470:
```suggestion
			if statusCode == http.NotFound || bytes.Contains(body, []byte("not found, pd")) {
```

pkg/cluster/api/pdapi.go
```diff
@@ -465,8 +465,16 @@ func (pc *PDClient) DelPD(name string, retryOpt *clusterutil.RetryOption) error
...
+				log.Debugf("PD node does not exist, ignore: %s", body)
```
lonng, line 471: Better to print the error too.

pkg/cluster/api/pdapi.go
```diff
@@ -562,8 +570,16 @@ func (pc *PDClient) DelStore(host string, retryOpt *clusterutil.RetryOption) err
 	endpoints := pc.getEndpoints(cmd)
 
 	_, err = tryURLs(endpoints, func(endpoint string) ([]byte, error) {
-		body, _, err := pc.httpClient.Delete(endpoint, nil)
-		return body, err
+		body, statusCode, err := pc.httpClient.Delete(endpoint, nil)
+		if err != nil {
+			if statusCode == 404 || bytes.Contains(body, []byte("not found")) {
```
lonng, line 575: ditto

pkg/cluster/api/pdapi.go
AstroProfundis: The error is just a wrap of body in this circumstance.

PR #602, opened 2020-07-14
Clone with yanked version

lucklove:
At this time, `tiup mirror clone <dir> <version>` will clone component manifest only contains <version>. 
If we deployed a cluster with an early version, some commands may fail since we can't find the entry in the manifest.

This PR will clone the component manifest with all versions, the versions besides the target version will be set as `yanked`


pkg/repository/v1manifest/types.go
```diff
@@ -210,6 +227,42 @@ func (manifest *Component) HasNightly(platform string) bool {
...
+	v, ok := manifest.Platforms[platform][manifest.Nightly]
+	if !ok {
+		return false
+	}
+
+	return !v.Yanked
+}
+
+// VersionList return all versions exclude yanked versions
+func (manifest *Component) VersionList(platform string) map[string]VersionItem {
```
lonng, line 239: I think public methods should well-tested.

lonng:
This PR lack of testing.

PR #605, opened 2020-07-15
Support yank a single version on client side

lucklove:
At this time tiup only supports yank a whole component (include all versions). 
In some scenes, we should yank a single version (for example, we push a wrong version).
This PR makes it possible to yank a single version.

### DIFF
--- a/cmd/mirror.go
+++ b/cmd/mirror.go
@@ -20,6 +20,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"sort"
+	"strings"
 	"time"
 
 	"github.com/pingcap/errors"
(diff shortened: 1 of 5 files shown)


lucklove:
The yanked version handle has been complete in https://github.com/pingcap/tiup/pull/602

cmd/mirror.go
```diff
@@ -253,15 +254,22 @@ func newMirrorModifyCmd() *cobra.Command {
 				return err
 			}
 
-			m, err := env.V1Repository().FetchComponentManifest(args[0])
+			spec := strings.Split(args[0], ":")
```
lonng, line 257: Please use `ParseComponentVersion` function.

PR #609, opened 2020-07-17
cluster: validate topo changes after edit-config

AstroProfundis:
add validation for topology diff to forbid changes of immutable fields

### Check List 

Tests 

 - Unit test

Code changes

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/pkg/cluster/edit/diff.go
+++ b/pkg/cluster/edit/diff.go
@@ -24,8 +24,9 @@ import (
 )
 
 const (
-	validateTagName  = "validate"
-	validateTagValue = "editable"
+	validateTagName     = "validate"
+	validateTagEditable = "editable"
+	validateTagIgnore   = "ignore"
 )
 
 // ShowDiff write diff result into the Writer.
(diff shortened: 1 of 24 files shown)


pkg/cluster/edit/diff.go
```diff
@@ -16,10 +16,19 @@ package edit
...
 	"fmt"
 	"io"
+	"strconv"
+	"strings"
 
+	"github.com/r3labs/diff"
 	"github.com/sergi/go-diff/diffmatchpatch"
 )
 
+const (
```
lonng, line 26: Prefer to move this into a common package, which can be used by `dm/cluster`.

pkg/cluster/edit/diff.go
```diff
@@ -29,3 +38,62 @@ func ShowDiff(t1 string, t2 string, w io.Writer) {
...
+			if c.Type == diff.UPDATE && c.Path[len(c.Path)-1] == validateTagEditable {
+				// If the field is marked as editable, it is allowed to be modified no matter
+				// its parent level element is marked as editable or not
+				continue
+			}
+
+			pathEditable := true
+			pathIgnore := false
+			for _, p := range c.Path {
+				if _, err := strconv.Atoi(p); err == nil {
```
lonng, line 73: I think we cannot ignore the slice, for example, user add a new host into `tidb_servers`.

pkg/cluster/edit/diff.go
```diff
@@ -29,3 +38,62 @@ func ShowDiff(t1 string, t2 string, w io.Writer) {
...
+				}
+			}
+			// if the path has any ignorable item, just ignore it
+			if pathIgnore || pathEditable && (c.Type == diff.CREATE || c.Type == diff.DELETE) {
+				// If *every* parent elements on the path are all marked as editable,
+				// AND the field itself is marked as editable, it is allowed to add or delete
+				continue
+			}
+		}
+		msg = append(msg, fmt.Sprintf("(%s) %s '%v' -> '%v'", c.Type, strings.Join(c.Path, "."), c.From, c.To))
```
lonng, line 92: I think we should generate more readable error message. e.g:

1. update: `cannot change the xxx field from 1 to 2`
2. delete: `cannot delete the field xxx`
3. create: `cannot add the field xxx into yyy`

pkg/cluster/edit/diff.go
AstroProfundis: `tidb_servers` is not ignored, `server_configs.tidb` is.

PR #612, opened 2020-07-20
Fix data_dir of tiflash when import cluster

### LINKED ISSUE #611: wrong data-dir of tiflash after import from ansible
## Bug Report

1. What did you do?

- deploy a cluster with tiflash using ansible
- import from ansible using tiup cluster import

after reload or upgrade, the config file of *.toml is same as the origin one.

3. What did you see instead?
`path` and `tmp_path` is different from the origin ones.
(the data_dir of tiflash is data/tiflash-{port} as default value of tiup cluster)

import command will only get data_dir from inventory.ini and start script of tiflash and the start script contains the only config file flag, so it can not get the truly use path and use the default value.

1.0.8

july2993:
fix #611
import command will only get data_dir from inventory.ini and start
script of tiflash and the start script contains the only config file
flag, so it can not get the truly use path and use the default value.

parse the config file of tiflash to get `path` and `tmp_path` and set it in `DataDir` and `TmpDir` in spec.

### test
- unit-test
- manual-test
  * test using import command and check the spec.

### DIFF
--- a/pkg/cluster/ansible/dirs.go
+++ b/pkg/cluster/ansible/dirs.go
@@ -15,9 +15,11 @@ package ansible
 
 import (
 	"fmt"
+	"path/filepath"
 	"strings"
 	"time"
 
+	"github.com/BurntSushi/toml"
 	"github.com/pingcap/errors"
 	"github.com/pingcap/tiup/pkg/cluster/executor"
 	"github.com/pingcap/tiup/pkg/cluster/spec"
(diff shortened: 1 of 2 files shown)


pkg/cluster/ansible/dirs.go
```diff
@@ -123,6 +125,21 @@ func parseDirs(user string, ins spec.InstanceSpec, sshTimeout int64) (spec.Insta
 				newIns.DeployDir = strings.Trim(strings.Split(line, " ")[1], "\"")
 				continue
 			}
+
```
lonng, line 128: It's better to add some tests.

july2993: forgot to commit the file: https://github.com/pingcap/tiup/pull/612/commits/6dd6f95c1dd479b206e26e8982d2ad8476cd73c7
but it only covers about parsing the file and get info

pkg/cluster/ansible/dirs_test.go
```diff
@@ -0,0 +1,61 @@
...
+var _ = check.Suite(&parseSuite{})
+
+func (s *parseSuite) TestParseTiflashConfigFromFileData(c *check.C) {
+	spec := new(spec.TiFlashSpec)
+	data := []byte(tiflashConfig)
+
+	err := parseTiflashConfigFromFileData(spec, data)
+	c.Assert(err, check.IsNil)
+
+	c.Assert(spec.DataDir, check.Equals, "/data1/test-cluster/leiysky-ansible-test-deploy/tiflash/data/db")
```
AstroProfundis, line 59: TiFlash supports multiple comma seperated paths like `/path/1,/path/2`, we should also add cases for that.

july2993: here we just need to make sure DataDir is the same as `path` in tiflash.toml, so I don't think we need to add suck case here.

PR #614, opened 2020-07-20
Add help msg of EDITOR

july2993:
Add help msg of EDITOR

```
➜  tiup git:(editor) ✗ ./bin/tiup-cluster edit-config -h
Edit TiDB cluster config.
Will use editor from environment variable `EDITOR`, default use vi

Usage:
  tiup-cluster edit-config <cluster-name> [flags]

Flags:
  -h, --help   help for edit-config

Global Flags:
      --wait-timeout int   Timeout in seconds to wait for an operation to complete, ignored for operations that don't fit. (default 120)
```


lonng:
https://github.com/pingcap/docs-cn/pull/4000

PR #615, opened 2020-07-20
Support native ssh client

lucklove:
At current we use easyssh as the ssh tunnel. However, sometimes users will use some ssh plugin such as LDAP. In this case the easyssh client can't work.

This PR introduces an option for users to choose the native ssh client in his host as the ssh client to connect the cluster. This can enable any ssh plugins once the user configs them correctly.

TODOLIST:
- [x] Add native ssh Executor struct
- [x] Implement Execute method for native ssh executor
  - [x] Support connect host with default key file
  - [x] Support connect with password (via sshpass)
  - [x] Support connect with user specified key file
- [x] Implement Transfer method for native ssh executor
- [x] Enable `--native-ssh` global flag
- [x] Adopt Linux
  - [x] OpenSSH_8.0p1
  - [x] OpenSSH_7.4p1
  - [x] OpenSSH_5.3p1
- [x] Adopt macos
- [x] Adopt different ssh client version (optional)
  - [x] Adopt different password prompt
- [x] Add test


pkg/cluster/executor/ssh.go
```diff
@@ -207,3 +226,111 @@ func (e *SSHExecutor) Transfer(src string, dst string, download bool) error {
...
+func (e *NativeSSHExecutor) Transfer(src string, dst string, download bool) error {
+	args := []string{"scp", "-r", "-o", "StrictHostKeyChecking=no"}
+	if e.Config.Password != "" {
+		args = append([]string{"sshpass", "-p", e.Config.Password, "-P", "password"}, args...)
+	}
+	if e.Config.KeyFile != "" {
+		args = append(args, "-i", e.Config.KeyFile)
+	}
+	if e.Config.Passphrase != "" {
+		args = append([]string{"sshpass", "-p", e.Config.Passphrase, "-P", "passphrase"}, args...)
```
lonng, line 317: Are you sure the `sshpass` always exists in the user environment?

lucklove: No, It's a tradeoff. I think it's not a hard task to install it in the bastion host.

components/cluster/command/root.go
```diff
@@ -78,6 +78,7 @@ func init() {
 	// Initialize the global variables
 	flags.ShowBacktrace = len(os.Getenv("TIUP_BACKTRACE")) > 0
 	cobra.EnableCommandSorting = false
+	gOpt.NativeSSH = len(os.Getenv(localdata.EnvNameNativeSSHClient)) > 0
```
lonng, line 81: I don't think it's better to use `len(os.Getenv(localdata.EnvNameNativeSSHClient)) > 0` to indicate the native ssh has been used. e.g:
```
TIUP_NATIVE_SSH=false
or
TIUP_NATIVE_SSH=0
or
TIUP_NATIVE_SSH=disable
```

Prefer to `len(os.Getenv(localdata.EnvNameNativeSSHClient)) > 0 && ( os.Getenv(localdata.EnvNameNativeSSHClient)) == false || os.Getenv(localdata.EnvNameNativeSSHClient)) == 1 || os.Getenv(localdata.EnvNameNativeSSHClient)) == enable)`

docker/secret/control.env
```diff
@@ -0,0 +1,4 @@
+# generated by tiup-cluster/docker/up.sh, parsed by tiup-cluster/docker/control/bashrc
```
lonng, line 1: I think we shouldn't add this file into git.

pkg/cluster/executor/ssh.go
```diff
@@ -81,28 +98,41 @@ type (
...
 	}
 
+	if native {
+		e := &NativeSSHExecutor{
+			Config: &c,
+			Locale: "C",
+			Sudo:   sudo,
+		}
+		if c.Password != "" || (c.KeyFile != "" && c.Passphrase != "") {
+			_, _, e.ConnectionTestResult = e.Execute(connectionTestCommand, false, executeDefaultTimeout)
```
lonng, line 122: I think we shouldn't try to connect remote when we create the instance of `SSHExecutor`.

pkg/localdata/constant.go
```diff
@@ -64,6 +64,12 @@ const (
 	// EnvTag is the tag of the running component
 	EnvTag = "TIUP_TAG"
 
+	// EnvNameSSHPassPrompt is the variable name by which user specific the password prompt for sshpass
+	EnvNameSSHPassPrompt = "TIUP_SSHPASS_PROMPT"
+
+	// EnvNameNativeSSHClient is the variable name by which user can specific use natiive ssh client or not
+	EnvNameNativeSSHClient = "TIUP_USE_NATIVE_SSH"
```
lonng, line 71:
```suggestion
	EnvNameNativeSSHClient = "TIUP_NATIVE_SSH"
```

lonng: I think we it's time to add a new subcommand `tiup env` to show all environment variables that TiUP aware. I have submit an issue https://github.com/pingcap/tiup/issues/632.

pkg/utils/exec.go
```diff
@@ -0,0 +1,35 @@
+package utils
```
lonng, line 1: License header is missing.

tests/tiup-cluster/test_cmd.sh
```diff
@@ -4,5 +4,5 @@ set -eu
 
 source script/cmd_subtest.sh
 
-echo "test cluster for verision v4.0.2 with CDC"
-cmd_subtest v4.0.2 true
+echo "test cluster for verision v4.0.2 with CDC, via easy ssh"
+cmd_subtest v4.0.2 true false
```
lonng, line 8: It's better to add a newline at the ending of the file.

pkg/cluster/executor/ssh.go
lucklove: If we don't try it this time when the user actually executes the command, it may block as long as the command timeout value since the wrong prompt of ssh. What's worse is that there will be no possibility to get clear what happened at that time.

For example, the user can change the ssh prompt to:
```
root@172.16.5.141's secret:
```
But the `sshpass` may expect it be:
```
root@172.16.5.141's password:
```
And then the command will hang forever (until receiving the kill signal).

.github/workflows/integrate.yaml
```diff
@@ -24,6 +24,10 @@ jobs:
           - "test_scale_core"
           - "test_scale_tools"
           - "test_upgrade"
+          - "test_cmd_with_native_ssh"
+          - "test_cmd_no_cdc_with_native_ssh"
```
AstroProfundis, line 28: This is not necessary, it's testing the logic of cluster deploying, not related to the SSH implementation.

components/cluster/command/root.go
```diff
@@ -79,6 +79,12 @@ func init() {
 	flags.ShowBacktrace = len(os.Getenv("TIUP_BACKTRACE")) > 0
 	cobra.EnableCommandSorting = false
 
+	if os.Getenv(localdata.EnvNameNativeSSHClient) == "true" ||
```
AstroProfundis, line 82: `strings.ToLower()`?

components/cluster/command/root.go
```diff
@@ -106,6 +112,10 @@ func init() {
 
 			teleCommand = getParentNames(cmd)
 
+			if gOpt.NativeSSH {
+				fmt.Println("Native ssh client will be used")
```
AstroProfundis, line 116: use log instead of fmt

components/cluster/command/root.go
```diff
@@ -123,6 +133,7 @@ func init() {
 	// start/stop operations is 90s, the default value of this argument is better be longer than that
 	rootCmd.PersistentFlags().Int64Var(&gOpt.OptTimeout, "wait-timeout", 120, "Timeout in seconds to wait for an operation to complete, ignored for operations that don't fit.")
 	rootCmd.PersistentFlags().BoolVarP(&skipConfirm, "yes", "y", false, "Skip all confirmations and assumes 'yes'")
+	rootCmd.PersistentFlags().BoolVar(&gOpt.NativeSSH, "native-ssh", gOpt.NativeSSH, "Use native local ssh client instead of the easy ssh client.")
```
AstroProfundis, line 136: "Use the native SSH client installed on local system instead of the build-in one."

pkg/cluster/executor/ssh.go
```diff
@@ -207,3 +237,127 @@ func (e *SSHExecutor) Transfer(src string, dst string, download bool) error {
...
+		}
+		return stdout.Bytes(), stderr.Bytes(), baseErr
+	}
+
+	return stdout.Bytes(), stderr.Bytes(), err
+}
+
+// Transfer copies files via SCP
+// This function depends on `scp` (a tool from OpenSSH or other SSH implementation)
+// This function is based on easyssh.MakeConfig.Scp() but with support of copying
```
AstroProfundis, line 336: This comment should be removed for native ssh

pkg/exec/run.go
```diff
@@ -28,6 +27,8 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/fatih/color"
+
```
AstroProfundis, line 31: remove this empty line and `go fmt` again

components/cluster/command/root.go
lonng: Nice catch

Issue #617, opened 2020-07-21
Misc adjustments of TiSpark support


AstroProfundis:
Some TODOs we found during testing of #531 

 - [x] Add validation of multiple tispark workers on the same host
 - [x] Add `java_home` to spec to set `JAVA_HOME` in systemd service file
 - [x] Set `SPARK_LOCAL_IP` in `tispark-env.sh` to the address of the node
 - [ ] Use better promts to inform user to install JRE

PR #621, opened 2020-07-22
cluster: optimize tispark supporting

AstroProfundis:
Close: #617


pkg/cluster/spec/spec.go
```diff
@@ -515,6 +516,17 @@ func (s *Specification) validateTiSparkSpec() error {
 		return ErrMultipleTiSparkMaster
 	}
 
+	// Multiple workers on the same host is not supported by Spark
+	if len(s.TiSparkWorkers) > 1 {
+		cnt := make(map[string]int)
+		for _, w := range s.TiSparkWorkers {
+			if cnt[w.Host] > 0 {
+				return ErrMultipleTisparkWorker
```
lonng, line 524: It's better to attach the `host` to the error message to help users figure out which host is duplicated.

PR #623, opened 2020-07-23
Refactor helper pacakges and fix spec validation

AstroProfundis:
The ports of monitoring agents are not checked with other existing clusters before deploying a new cluster.

 - Refactor the package `prepare` and `clusterutils.Retry` to make it less possible of having import cycle.
 - Refactor the pre-deploy cross cluster conflict checks to involve ports of monitoring agents and make them easier to test
 - Add test cases for cross cluster conflict checks

### Check List 

Tests 

 - Unit test

Code changes


pkg/cluster/spec/instance.go
```diff
@@ -330,11 +330,20 @@ func (i *instance) GetPort() int {
 }
 
 func (i *instance) UsedPorts() []int {
-	return i.usedPorts
+	return append(
+		i.usedPorts,
+		i.topo.MonitoredOptions.NodeExporterPort,
```
lonng, line 335: These ports should be unique by the host not instance.

lucklove:
Should we apply this logic in `CheckClusterDirConflict`?
We should check directory conflict, too.

pkg/cliutil/prepare/prepare.go
```diff
@@ -216,6 +217,10 @@ func CheckClusterPortConflict(clusterSpec *meta.SpecManager, clusterName string,
 			return errors.Trace(err)
 		}
 
+		uniqueHosts := set.NewStringSet()
```
lonng, line 220: Add some tests for this change.

lonng:
> Should we apply this logic in `CheckClusterDirConflict`?
> We should check directory conflict, too.

I think we should.

AstroProfundis:
There is already checks for dir conflicts with `hostDirAccessor`

pkg/cliutil/prepare/prepare.go
```diff
@@ -231,13 +232,30 @@ func CheckClusterPortConflict(clusterSpec *spec.SpecManager, clusterName string,
 		}
 
 		metadata.GetTopology().IterInstance(func(inst spec.Instance) {
+			uniqueHosts := set.NewStringSet()
```
lonng, line 235: Do you have tested this PR? I think the `uniqueHosts := set.NewStringSet()` should be placed out of `IterInstance`.

pkg/cliutil/prepare/prepare.go
```diff
@@ -248,6 +266,20 @@ func CheckClusterPortConflict(clusterSpec *spec.SpecManager, clusterName string,
 				port:     port,
 			})
 		}
+
+		uniqueHosts := set.NewStringSet()
+		if !uniqueHosts.Exist(inst.GetHost()) {
+			uniqueHosts.Insert(inst.GetHost())
```
lonng, line 272: ditto

pkg/cliutil/prepare/prepare.go
AstroProfundis: It should be in the loop, it "unique" for on cluster, not for all clusters

lonng: Yes, it should be placed in the `for` loop, but above the L234.

pkg/utils/retry.go
```diff
@@ -13,16 +13,99 @@
...
+		default:
+			time.Sleep(cfg.Delay)
+		}
+	}
+
+	return fmt.Errorf("operation exceeds the max retry attempts of %d", cfg.Attempts)
+}
+
+// IsTimeoutOrMaxRetry return true if it's timeout or reach max retry.
+func IsTimeoutOrMaxRetry(err error) bool {
```
lucklove, line 98: It's better to split into two functions: `IsTimeout` and `IsMaxRetry`. They have different semantics.

lucklove:
Rest LGTM

PR #634, opened 2020-07-28
Overwrite the default --help persistent flag

lonng:

Overwrite the default --help persistent flag

### Why do we change it?

Some components will define themself `-h` flag, eg: tiup dumpling -h ${host}. This PR try to leave the handling of `-h` flag to the component.


lonng:
@july2993 addressed.

july2993:
> @july2993 addressed.

I mean to add in the code.
Since it's hard to know why we need to do this and for what?

Issue #177, opened 2020-04-19
Tiup should ignore deprecated components


Fullstop000:
![image](https://user-images.githubusercontent.com/12471960/79682558-df652400-8255-11ea-9c66-2418bd8758f3.png)

Since the `tpc` has been integrated into `bench`,  I think it should not fire a `NOT FOUND` error. And it can be more user friendly to support cleaning up deprecated components automatically.

lonng:
@Fullstop000 Nice catch, can you submit a PR to fix it?

nrc:
Long term, this is covered by the design in #193 (called yanked, rather than deprecated)

AstroProfundis:
Yanked component is already implemented in #574 and we should adjust existing manifests to set values for them. cc @lucklove

PR #635, opened 2020-07-28
Handle yanked version in update

lucklove:
Fix https://github.com/pingcap/tiup/issues/177


tests/tiup/tiup.toml
```diff
@@ -1 +1 @@
-mirror = "https://tiup-mirrors.pingcap.com"
+mirror = "http://172.16.5.139:8988"
```
lonng, line 1: ?

pkg/repository/v1_repository.go
```diff
@@ -639,8 +646,8 @@ func (r *V1Repository) UpdateComponentManifests() error {
 	}
 
 	for name := range index.Components {
-		_, err = r.updateComponentManifest(name)
-		if err != nil {
+		_, err = r.updateComponentManifest(name, false)
+		if err != nil && err != errUnknownComponent {
```
lonng, line 650: It's better to unwrap the error to prevent the original error from being wrapped with stack or annotation.

tests/tiup/tiup.toml
lonng: Please secondary confirmation after commit pushed.

pkg/repository/v1_repository.go
```diff
@@ -427,7 +427,14 @@ func (r *V1Repository) updateComponentManifest(id string) (*v1manifest.Component
...
+	var components map[string]v1manifest.ComponentItem
+	if withYanked {
+		components = index.ComponentListWithYanked()
+	} else {
+		components = index.ComponentList()
+	}
+
+	item, ok := components[id]
 	if !ok {
 		return nil, errUnknownComponent
```
lonng, line 439: The returned error should contain call stack.
```suggestion
		return nil, errors.AddStack(errUnknownComponent)
```

PR #636, opened 2020-07-28
Enhance native ssh test

lucklove:
Add failpoint check in integration testing to assert the `--native-ssh` flag really works

屏幕快照 2020-07-29 上午10 12 45


pkg/cluster/executor/ssh.go
```diff
@@ -124,6 +125,15 @@ func NewSSHExecutor(c SSHConfig, sudo bool, native bool) Executor {
 		return e
 	}
 
+	// Used in integration testing, to check if native ssh client is really used when it need to be.
+	failpoint.Inject("assertNativeSSH", func() {
```
lonng, line 129: 👍 Gooood idea.

PR #643, opened 2020-07-29
Tweak the upgrade order of components

### LINKED ISSUE #641: Upgrade TiCDC first then upgrade TiKV
## Bug Report

1. What did you do?

During development, we have changed the protocol between TiCDC and TiKV, and there are some compatibility problems.
The solution is upgrading TiCDC first then upgrade TiKV.

Upgrade TiCDC first then upgrade TiKV.

3. What did you see instead?

TiUP upgrades TiKV first.

master.

lonng:

Fix #641


lonng:
@overvenus  PTAL

Issue #138, opened 2020-04-13
Partially destroy cluster


BusyJay:
## Feature Request

**Describe the feature you'd like:**

Tiup seems to only support destroy the whole cluster. But during development or re-deploy, not all data are expected to be cleaned, especially metrics. In short, we need a cluster version `clean`, working similar to `unsafe_cleanup_data` playbook from tidb-ansible.

lonng:
@BusyJay How about `tiup cluster clean <cluster-name>`

BusyJay:
LGTM

lonng:
Dup with #396 and fixed in #644

PR #644, opened 2020-07-30
Add clean sub command

lucklove:
Fix https://github.com/pingcap/tiup/issues/138

TODOLIST:
- [x] Primary cleanup: tidb/tikv/pd/promethus/grafana
- [x] Support skip specific role, eg: prometheus 
- [x] Support cleanup log

### DIFF
--- a/components/cluster/command/clean.go
+++ b/components/cluster/command/clean.go
@@ -14,22 +14,16 @@
 package command
 
 import (
-	perrs "github.com/pingcap/errors"
 	operator "github.com/pingcap/tiup/pkg/cluster/operation"
-	"github.com/pingcap/tiup/pkg/cluster/spec"
-	"github.com/pingcap/tiup/pkg/set"
 	"github.com/spf13/cobra"
 )
 
 func newCleanCmd() *cobra.Command {
-	destoyOpt := operator.Options{}
+	cleanOpt := operator.Options{}
+
 	cmd := &cobra.Command{
-		Use: "clean <cluster-name>",
-		Long: `Destroy a specified cluster, which will clean the deployment binaries and data.
-You can retain some nodes and roles data when clean cluster, eg:
-  
-  $ tiup cluster clean <cluster-name> --retain-role-data prometheus
-  $ tiup cluster clean <cluster-name> --retain-node-data 172.16.13.11:9000`,
+		Use:  "clean <cluster-name>",
+		Long: `clean cluster data without destroy cluster`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if len(args) != 1 {
 				return cmd.Help()
@@ -38,22 +32,9 @@ You can retain some nodes and roles data when clean cluster, eg:
 			clusterName := args[0]
 			teleCommand = append(teleCommand, scrubClusterName(clusterName))
 
-			// Validate the retained roles to prevent unexpected deleting data
-			if len(destoyOpt.RetainDataRoles) > 0 {
-				validRoles := set.NewStringSet(spec.AllComponentNames()...)
-				for _, role := range destoyOpt.RetainDataRoles {
-					if !validRoles.Exist(role) {
-						return perrs.Errorf("role name `%s` invalid", role)
-					}
-				}
-			}
-
-			return manager.DestroyCluster(clusterName, gOpt, destoyOpt, skipConfirm)
+			return manager.CleanCluster(clusterName, gOpt, cleanOpt, skipConfirm)
 		},
 	}
-
-	cmd.Flags().StringArrayVar(&destoyOpt.RetainDataNodes, "retain-node-data", nil, "Specify the nodes whose data will be retained")
-	cmd.Flags().StringArrayVar(&destoyOpt.RetainDataRoles, "retain-role-data", nil, "Specify the roles whose data will be retained")
 
 	return cmd
 }
--- a/components/cluster/command/clean.go
+++ b/components/cluster/command/clean.go
@@ -14,7 +14,6 @@
 package command
 
 import (
-	perrs "github.com/pingcap/errors"
 	operator "github.com/pingcap/tiup/pkg/cluster/operation"
 	"github.com/spf13/cobra"
 )
@@ -28,9 +27,9 @@ func newCleanCmd() *cobra.Command {
 		Long: `Cleanup a specified cluster without destroying it.
 	You can retain some nodes and roles data when cleanup the cluster, eg:
 		  
-	$ tiup cluster clean <cluster-name> --ignore-role prometheus
-	$ tiup cluster clean <cluster-name> --ignore-node 172.16.13.11:9000
-	$ tiup cluster clean <cluster-name> --ignore-node 172.16.13.12`,
+	$ tiup cluster clean <cluster-name> --all --ignore-role prometheus
+	$ tiup cluster clean <cluster-name> --log --ignore-node 172.16.13.11:9000
+	$ tiup cluster clean <cluster-name> --data --ignore-node 172.16.13.12`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if len(args) != 1 {
 				return cmd.Help()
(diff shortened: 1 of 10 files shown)


pkg/cluster/manager.go
```diff
@@ -199,6 +199,52 @@ func (m *Manager) ListCluster() error {
...
+	if err != nil {
+		return err
+	}
+
+	topo := metadata.GetTopology()
+	base := metadata.GetBaseMeta()
+
+	if !skipConfirm {
+		if err := cliutil.PromptForConfirmOrAbortError(
+			"This operation will clean %s %s cluster %s's data.\nDo you want to continue? [y/N]:",
```
lonng, line 214: Please print out the retained nodes and roles.

pkg/cluster/operation/destroy.go
```diff
@@ -171,6 +196,77 @@ func DestroyMonitored(getter ExecutorGetter, inst spec.Instance, options *spec.M
...
+
+	for _, ins := range instances {
+		// Some data of instances will be retained
+		dataRetained := retainDataRoles.Exist(ins.ComponentName()) ||
+			retainDataNodes.Exist(ins.ID())
+
+		e := getter.Get(ins.GetHost())
+		log.Infof("Cleanup instance %s", ins.GetHost())
+
+		if err := ins.WaitForDown(e, timeout); err != nil {
```
lonng, line 214: The `stop` operation will wait for down, do we need to wait again?

lonng: Do we need to add a `--force` flag to skip some error and force cleanup the cluster?

pkg/cluster/operation/destroy.go
```diff
@@ -171,6 +196,77 @@ func DestroyMonitored(getter ExecutorGetter, inst spec.Instance, options *spec.M
...
+		var logDirs []string
+		if len(ins.DataDir()) > 0 {
+			dataDirs = strings.Split(ins.DataDir(), ",")
+		}
+		if len(ins.LogDir()) > 0 {
+			logDirs = strings.Split(ins.LogDir(), ",")
+		}
+
+		delFiles := set.NewStringSet()
+		if !dataRetained {
```
lonng, line 233: I think we should skip the current iteration if the `dataRetained ==true`.

pkg/cluster/operation/destroy.go
lucklove: No, the `stop` operation only waits `systemctl` return, it doesn't `WaitForDown`

components/cluster/command/clean.go
```diff
@@ -0,0 +1,45 @@
...
+			clusterName := args[0]
+			teleCommand = append(teleCommand, scrubClusterName(clusterName))
+
+			return manager.CleanCluster(clusterName, gOpt, cleanOpt, skipConfirm, withLog)
+		},
+	}
+
+	cmd.Flags().StringArrayVar(&cleanOpt.RetainDataNodes, "retain-node-data", nil, "Specify the nodes whose data will be retained")
+	cmd.Flags().StringArrayVar(&cleanOpt.RetainDataRoles, "retain-role-data", nil, "Specify the roles whose data will be retained")
+	cmd.Flags().BoolVar(&withLog, "with-log", false, "Cleanup log too")
```
lonng, line 42: I think `with-log` is a bad flag name, @AstroProfundis @july2993 any suggestion?

july2993: what about `clean-log`?

lonng: @lucklove I think we should consider more scenarios and define the precise semantics for `tiup cluster clean`, e.g:
1. `tiup cluster clean` means clean all data and logs
2. `tiup cluster clean --only-log` means clean log only
3. `tiup cluster clean --only-data` means clean data only

AstroProfundis: I'd prefer:
 - `clean all`
 - `clean log`
 - `clean data`
 - `clean sth-else`

components/cluster/command/clean.go
```diff
@@ -33,13 +34,24 @@ func newCleanCmd() *cobra.Command {
...
 			teleCommand = append(teleCommand, scrubClusterName(clusterName))
 
-			return manager.CleanCluster(clusterName, gOpt, cleanOpt, skipConfirm, withLog)
+			if cleanALl {
+				cleanOpt.CleanupData = true
+				cleanOpt.CleanupLog = true
+			}
+
+			if !(cleanOpt.CleanupData || cleanOpt.CleanupLog) {
+				return perrs.Errorf("at least one of `--all` `--data` `--log` should be specified")
```
lonng, line 43: Use a more concrete suggestion instead of an error. e.g:
```
Please choose one of the following commands to clean the cluster:

    tiup cluster clean %s --data                  # Only clean data in the cluster
    tiup cluster clean %s --log                   # Only clean log in the cluster
    tiup cluster clean %s --all                   # Both data and log will be cleaned.
```

pkg/cluster/operation/destroy.go
```diff
@@ -171,6 +195,85 @@ func DestroyMonitored(getter ExecutorGetter, inst spec.Instance, options *spec.M
...
+				return errors.Annotatef(err, str)
+			}
+			log.Warnf("You may manually check if the process on %s:%d is still running", ins.GetHost(), ins.GetPort())
+		}
+
+		var dataDirs []string
+		var logDirs []string
+		if len(ins.DataDir()) > 0 {
+			dataDirs = strings.Split(ins.DataDir(), ",")
+		}
```
lonng, line 230: Please move L228-L230 into
```go
if options.CleanupData && len(ins.DataDir()) > 0 {
			for _, dataDir := range  strings.Split(ins.DataDir(), ",") {
				delFiles.Insert(path.Join(dataDir, "*"))
			}
		}
```

pkg/cluster/operation/destroy.go
```diff
@@ -171,6 +195,85 @@ func DestroyMonitored(getter ExecutorGetter, inst spec.Instance, options *spec.M
...
+		if len(ins.LogDir()) > 0 {
```
lonng, line 231: ditto

pkg/cluster/operation/destroy.go
```diff
@@ -187,7 +290,7 @@ func DestroyComponent(getter ExecutorGetter, instances []spec.Instance, cls spec
 	for _, ins := range instances {
 		// Some data of instances will be retained
 		dataRetained := retainDataRoles.Exist(ins.ComponentName()) ||
-			retainDataNodes.Exist(ins.ID())
+			retainDataNodes.Exist(ins.ID()) || retainDataNodes.Exist(ins.GetHost())
```
lonng, line 293: If we apply this change, we should change the documentation, this feature allows user to retain data by the host.

components/cluster/command/clean.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+func newCleanCmd() *cobra.Command {
+	cleanOpt := operator.Options{}
+	cleanALl := false
+
+	cmd := &cobra.Command{
+		Use: "clean <cluster-name>",
+		Long: `Cleanup a specified cluster without destroying it.
+	You can retain some nodes and roles data when cleanup the cluster, eg:
```
lonng, line 28: Please use 4 spaces to indent this content instead of a <TAB>.
Another issue is better to demonstrate `--ignore-xxx` in an individual command and doesn't hybrid them.

tests/tiup-dm/test_cmd.sh
```diff
@@ -15,6 +15,13 @@ tiup-dm --yes deploy $name $version $topo -i ~/.ssh/id_rsa
 
 tiup-dm list | grep "$name"
 
+# debug
```
lonng, line 18: Will we delete this debug command lines finally?

lucklove: Unless we make clear why CI brokes

lonng: OK, it's better to create an issue and link the issue here to help us track it and delete the debug command lines when the issue has been resolved.

tests/tiup-cluster/script/cmd_subtest.sh
```diff
@@ -54,5 +54,11 @@ function cmd_subtest() {
 
     tiup-cluster $client display $name
 
+    tiup-cluster $client --yes clean $name --data --all --ignore-node 172.19.0.102:20160
+
+    echo "checking cleanup data and log"
+    tiup-cluster $client exec $name -N 172.19.0.102 --command "ls /home/tidb/deploy/tikv-20160/log/tikv.log"
```
lonng, line 60: We have a `tiup-cluster _test` hidden sub-command, it's better to add a test for this PR. e.g:
1. insert some rows
2. clean up cluster
3. check the rows

lonng:
@lucklove Please write the release note for this PR.

PR #654, opened 2020-08-05
Add enable and disable sub command

lucklove:
issue: https://asktug.com/t/topic/36387
TODOLIST:
- [x] Don't enable systemd service on in `startInstance()`
- [x] Enable systems service in `deploy` command and  add `--skip-enable` as an option
- [x] Add `enable` and `disable` command

Unresoved issue: https://github.com/pingcap/tiup/issues/697

```release-note
Support tiup cluster enable and tiup cluster disable command
```


components/cluster/command/deploy.go
```diff
@@ -90,6 +91,7 @@ func newDeploy() *cobra.Command {
 	cmd.Flags().StringVarP(&opt.IdentityFile, "identity_file", "i", opt.IdentityFile, "The path of the SSH identity file. If specified, public key authentication will be used.")
 	cmd.Flags().BoolVarP(&opt.UsePassword, "password", "p", false, "Use password of target hosts. If specified, password authentication will be used.")
 	cmd.Flags().BoolVarP(&opt.IgnoreConfigCheck, "ignore-config-check", "", opt.IgnoreConfigCheck, "Ignore the config check result")
+	cmd.Flags().BoolVarP(&skipEnable, "skip-enable", "", skipEnable, "Don't start the cluster automaticlly after host restart")
```
lonng, line 94: Prefer to not provide this flag, or find a better name for it.

components/cluster/command/disable.go
```diff
@@ -0,0 +1,44 @@
...
+package command
+
+import (
+	"github.com/spf13/cobra"
+)
+
+func newDisableCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "disable <cluster-name>",
+		Short: "Stop starting a TiDB cluster automatically at boot",
```
lonng, line 23:
```suggestion
		Short: "Disable starting a TiDB cluster automatically at boot",
```

components/cluster/command/disable.go
```diff
@@ -0,0 +1,44 @@
...
+			}
+
+			clusterName := args[0]
+			teleCommand = append(teleCommand, scrubClusterName(clusterName))
+
+			return manager.EnableCluster(clusterName, gOpt, false)
+		},
+	}
+
+	cmd.Flags().StringSliceVarP(&gOpt.Roles, "role", "R", nil, "Only start specified roles")
```
lonng, line 40:
```suggestion
	cmd.Flags().StringSliceVarP(&gOpt.Roles, "role", "R", nil, "Only disable specified roles")
```

components/cluster/command/disable.go
```diff
@@ -0,0 +1,44 @@
...
+	cmd.Flags().StringSliceVarP(&gOpt.Nodes, "node", "N", nil, "Only start specified nodes")
```
lonng, line 41: ditto

components/cluster/command/enable.go
```diff
@@ -0,0 +1,44 @@
...
+package command
+
+import (
+	"github.com/spf13/cobra"
+)
+
+func newEnableCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "enable <cluster-name>",
+		Short: "Start a TiDB cluster automatically at boot",
```
lonng, line 23:
```suggestion
		Short: "Enable a TiDB cluster automatically at boot",
```

components/cluster/command/enable.go
```diff
@@ -0,0 +1,44 @@
...
+			}
+
+			clusterName := args[0]
+			teleCommand = append(teleCommand, scrubClusterName(clusterName))
+
+			return manager.EnableCluster(clusterName, gOpt, true)
+		},
+	}
+
+	cmd.Flags().StringSliceVarP(&gOpt.Roles, "role", "R", nil, "Only start specified roles")
```
lonng, line 40: ditto

components/cluster/command/deploy.go
AstroProfundis: I think it could be just named as `--disable`, like `deploy xxx vx.y.z xxx.yaml --disable`

tests/tiup-cluster/script/cmd_subtest.sh
```diff
@@ -67,6 +67,13 @@ function cmd_subtest() {
     tiup-cluster $client display "tmp-cluster-name"
     tiup-cluster $client rename "tmp-cluster-name" $name
 
+    # Test enable & disable
```
lonng, line 70: Add some test cases to disable all and enable all.

components/cluster/command/disable.go
AstroProfundis: on boot

pkg/cluster/operation/action.go
```diff
@@ -441,6 +475,52 @@ func RestartComponent(getter ExecutorGetter, instances []spec.Instance, timeout
...
+	c := module.SystemdModuleConfig{
+		Unit:    ins.ServiceName(),
+		Action:  action,
+		Timeout: time.Second * time.Duration(timeout),
+	}
+	systemd := module.NewSystemdModule(c)
+	stdout, stderr, err := systemd.Execute(e)
+
+	if len(stdout) > 0 {
+		fmt.Println(string(stdout))
```
AstroProfundis, line 501: Will this broke the output on parallel executing?

tests/tiup-cluster/script/cmd_subtest.sh
```diff
@@ -67,6 +67,13 @@ function cmd_subtest() {
...
+    tiup-cluster $client exec $name -R tidb --command="systemctl status tidb-4000|grep 'enabled;'"
```
AstroProfundis, line 71: I think the cluster name should also be added in the assertion.

components/cluster/command/deploy.go
lucklove: If it's `--disable`, we should execute disable for the cluster (because the default value may be enable in the cluster)

pkg/cluster/operation/action.go
lucklove: I think no. Because the `start`, `restart`, `stop` action works well with this.

tests/tiup-cluster/script/cmd_subtest.sh
lucklove: This action doesn't change cluster name

components/cluster/command/deploy.go
lonng: So, will we remove this flag?

lonng (commented):
This PR needs to take some consideration more carefully about how to handle the exporters processes.

lucklove:
> This PR needs to take some consideration more carefully about how to handle the exporters processes.

@lonng I added an unresolved issue: https://github.com/pingcap/tiup/issues/697

pkg/cluster/module/systemd.go
```diff
@@ -67,11 +66,6 @@ func NewSystemdModule(config SystemdModuleConfig) *SystemdModule {
 	cmd := fmt.Sprintf("%s %s %s",
 		systemctl, strings.ToLower(config.Action), config.Unit)
 
-	if config.Enabled {
```
lonng, line 70: It's better to enable in default.

components/cluster/command/deploy.go
lucklove: OK

lonng (approved):
LGTM

Issue #658, opened 2020-08-05
tiup update --self will break root.json when using offline package


lucklove:
The reason is that the `clone` command doesn't unpack and replace the `root.json` in `tiup.tar.gz`, so the `root.json` in `tiup.tar.gz` is the same as the online mirror.

The result is that after executing `tiup update --self` with the offline package, the manifests from the offline package will be recognized as invalid.

![image](https://user-images.githubusercontent.com/8718109/89411161-8911e780-d757-11ea-81ef-da9de47736d5.png)

PR #659, opened 2020-08-05
Fix tiup update --self with offline package

lucklove:
Fix https://github.com/pingcap/tiup/issues/658

### DIFF
--- a/pkg/environment/env.go
+++ b/pkg/environment/env.go
@@ -220,12 +220,7 @@ func (env *Environment) SelfUpdate() error {
 		}
 
 		// Cover the root.json from tiup.bar.gz
-		profile := localdata.InitProfile()
-		if err := profile.ResetMirror(Mirror(), ""); err != nil {
-			return err
-		}
-
-		return nil
+		return localdata.InitProfile().ResetMirror(Mirror(), "")
 	}
 
 	return env.repo.DownloadTiup(env.LocalPath("bin"))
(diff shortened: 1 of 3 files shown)


pkg/repository/clone_mirror.go
```diff
@@ -306,27 +306,6 @@ func cloneComponents(repo *V1Repository,
 		compManifests[name] = newManifest
 	}
 
-	// Download TiUP binary
-	for _, goos := range options.OSs {
```
lonng, line 310: Why delete this part?

lucklove: It's of no use in V1. This part doesn't change add `tiup` to manifests. But with V1 repository, tiup update self only read from manifests.

pkg/environment/env.go
```diff
@@ -215,7 +215,17 @@ func (env *Environment) PlatformString() string {
...
+			return err
+		}
+
+		// Cover the root.json from tiup.bar.gz
+		profile := localdata.InitProfile()
+		if err := profile.ResetMirror(Mirror(), ""); err != nil {
+			return err
+		}
+
+		return nil
```
lonng, line 228: The +224 to +228 can be simplified to `return profile.ResetMirror(Mirror(), "")`

lonng:
@lucklove Please write the release note for this PR.

Issue #661, opened 2020-08-06
tiup fails to reload when there is only one TiKV and one PD


BusyJay:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?
Deploy a cluster that has only one TiKV and one PD. And then restart the cluster.
```
% tiup cluster reload test-cluster
Restarting component pd
        Stopping instance 172.0.0.1
        Stop pd 172.0.0.1:2349 success
        Starting instance pd 172.0.0.1:2349
        Start pd 172.0.0.1:2349 success
Restarting component tikv
Still waitting for the PD leader to be elected
Evicting 1 leaders from store 172.0.0.1:20160...
Still waitting for 1 store leaders to transfer...
Still waitting for 1 store leaders to transfer...
```

2. What did you expect to see?
Reload success quickly.

3. What did you see instead?
tiup keeps waiting for TiKV to evict all its leader. However, there is only one TiKV, so evict leader will never succeed. It will wait until timeout.

4. What version of TiUP are you using (`tiup --version`)?
```
% tiup --version
v1.0.8 tiup
Go Version: go1.13
Git Branch: master
GitHash: 4276089
```

Note that the same deploy topology, reloading pd can finish instantly. tiup seems to skip evicting leader for PD already. It should do the same thing for TiKV.

PR #662, opened 2020-08-06
Skip evict store when only 1 tikv

lucklove:
Fix https://github.com/pingcap/tiup/issues/661


lonng:
Do we need to skip the evict leader for PD?

lucklove:
> Do we need to skip the evict leader for PD?

PD has already been skipped.

PR #668, opened 2020-08-10
Update examples/topology.example.yaml

### LINKED ISSUE #667: Misleading example configuration
## Bug Report

1. What did you do?

https://github.com/pingcap/tiup/blob/a1c16375264cb4cb0059b9b8c95a52e5f38da6f5/examples/topology.example.yaml#L16-L17

The example uses 20% without extra explanation. People usually think it's 20% of total CPU. In fact, it means 20% of a **single** CPU. I suggest to change it to 200% and add comment explaining what it means.

This is not an issue in imagination, I have seen two cases that make the same mistake that setting it to 50% and hopping to use half the machine resources.

An alternatives is TiUP uses the value as a percentage of all cores, calculates the final cores and sets it automatically. Though it may not be backward compatible.

july2993:
Update examples/topology.example.yaml

fix #667 
more directly reference links and add comments.

### DIFF
--- a/examples/topology.example.yaml
+++ b/examples/topology.example.yaml
@@ -11,10 +11,12 @@ global:
   # # See: https://www.freedesktop.org/software/systemd/man/systemd.resource-control.html
   # # Supports using instance-level `resource_control` to override global `resource_control`.
   # resource_control:
-  #   # See: https://www.freedesktop.org/software/systemd/man/systemd.resource-control.html#IOReadBandwidthMax=device%20bytes
+  #   # See: https://www.freedesktop.org/software/systemd/man/systemd.resource-control.html#MemoryLimit=bytes
   #   memory_limit: "2G"
-  #   # See: https://www.freedesktop.org/software/systemd/man/systemd.resource-control.html#IOReadBandwidthMax=device%20bytes
-  #   cpu_quota: "20%"
+  #   # See: https://www.freedesktop.org/software/systemd/man/systemd.resource-control.html#CPUQuota=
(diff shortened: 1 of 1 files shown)


examples/topology.example.yaml
```diff
@@ -11,10 +11,12 @@ global:
...
   # # Supports using instance-level `resource_control` to override global `resource_control`.
   # resource_control:
-  #   # See: https://www.freedesktop.org/software/systemd/man/systemd.resource-control.html#IOReadBandwidthMax=device%20bytes
+  #   # See: https://www.freedesktop.org/software/systemd/man/systemd.resource-control.html#MemoryLimit=bytes
   #   memory_limit: "2G"
-  #   # See: https://www.freedesktop.org/software/systemd/man/systemd.resource-control.html#IOReadBandwidthMax=device%20bytes
-  #   cpu_quota: "20%"
+  #   # See: https://www.freedesktop.org/software/systemd/man/systemd.resource-control.html#CPUQuota=
+  #   # The percentage specifies how much CPU time the unit shall get at maximum, relative to the total CPU time available on one CPU. Use values > 100% for allotting CPU time on more than one CPU.
+  #   # Example: CPUQuota=20% ensures that the executed processes will never get more than 20% CPU time on one CPU.
```
lonng, line 18: This comment is inconsistent with the next line. Oops, all the `see: xxxx` comments are wrong.

july2993, line 18:
```suggestion
  #   # Example: CPUQuota=200% ensures that the executed processes will never get more than two CPU time.
```

july2993: https://github.com/pingcap/tiup/pull/668/commits/2a9dc6a145ca2f2d97c95c2672e7474c2f5a6ab2

PR #670, opened 2020-08-11
tui: fix reverse confirmation prompt

AstroProfundis:
When `edit-config` fails on immutable field been changed, enter empty input on the `Do you want to continue editing` prompt does not trigger the default behavior of opening editor again, but instead returned.

 - Change the implementation of prompts to handle empty input correctly
 - Fix typo

### Check List 

Tests 

Related changes


lonng:
@AstroProfundis Please supplement the release note.

PR #671, opened 2020-08-11
Support rename cluster

lucklove:
Make it possible to rename cluster.

Rename the cluster storage folder and reload prometheus and grafana.
eg: `tiup cluster rename old new` equals to:
```sh
mv ~/.tiup/storage/cluster/clusters/old ~/.tiup/storage/cluster/clusters/new
tiup cluster reload new -R grafana,prometheus
```

### Check List 

Tests 

 - Integration test

Side effects

Grafana stores some data related to cluster name to its grafana.db. The rename action will NOT delete them. So there may be some useless panel need to be deleted manually.

Release notes:

```release-note
Support rename cluster by the command `tiup cluster rename <old-name> <new-name>`
> Grafana stores some data related to cluster name to its grafana.db. The rename action will NOT delete them. So there may be some useless panel need to be deleted manually.
```


components/cluster/command/rename.go
```diff
@@ -0,0 +1,29 @@
+package command
```
lonng, line 1: Missing license header.

pkg/cluster/manager.go
```diff
@@ -522,6 +522,18 @@ func (m *Manager) EditConfig(clusterName string, skipConfirm bool) error {
 	return nil
 }
 
+// Rename the cluster
+func (m *Manager) Rename(clusterName string, opt operator.Options, newName string) error {
+	if err := os.Rename(m.specManager.Path(clusterName), m.specManager.Path(newName)); err != nil {
```
lonng, line 527: You must ensure the old cluster exists and the new cluster name doesn't exist.

PR #672, opened 2020-08-12
Add import command for dm

july2993:
Add import command for tiup-dm

import command work:

- generate a topology from ansible, also copy the source config files which generated from the config of the worker to master,
these files are needed for the master to upgrade from 1.0 to 2.0
- deploy a new version of the dm (>= 2.0) using tiup-dm deploy command
- user start the cluster and just like deploy a cluster using tiup-dm

code change:
- add a `components/dm/ansible` pkg, contains an `Importer` for handling generate topology from ansible and copy source config file to master, `testdata` contains some files deployed by ansible for the unit test.
- add an import command
- add a test script `test_cmd.sh` which can auto-deploy a v1.0.6 cluster using dm-ansible and run import command of tiup-dm

### Check List 

Tests 

 - Unit test
 - Integration test
test_import.sh can auto deeply a dm 1.0.6 cluster using dm-ansible
then run the import command of tiup-dm


july2993:
@csuzhangxc PTAL

components/dm/ansible/testdata/deploy_dir/172.19.0.101/conf/dm-worker.toml
```diff
@@ -0,0 +1,27 @@
...
+
+
+#charset of DSN of source mysql/mariadb instance
+charset = ""
+meta-dir = ""
+
+[from]
+host = "mysql1"
+user = "root"
+password = ""
```
csuzhangxc, line 20: ditto?

components/dm/command/import.go
```diff
@@ -0,0 +1,109 @@
...
+
+func newImportCmd() *cobra.Command {
+	var ansibleDir string
+	var inventoryFileName string
+	var rename string
+	var clusterVersion string
+
+	cmd := &cobra.Command{
+		Use:   "import",
+		Short: "Import an exist DM 1.0 cluster from tidb-ansible and re-deploy 2.0 version",
```
csuzhangxc, line 26:
```suggestion
		Short: "Import an exist DM 1.0 cluster from dm-ansible and re-deploy 2.0 version",
```

components/dm/command/import.go
```diff
@@ -0,0 +1,109 @@
...
+			if err != nil {
+				return errors.AddStack(err)
+			}
+
+			data, err := yaml.Marshal(meta.Topology)
+			if err != nil {
+				return errors.AddStack(err)
+			}
+
+			f, err := ioutil.TempFile("", "tiup-*")
```
csuzhangxc, line 56: How about giving it a meaningful name? like `tiup-source-{source-id}.yaml`?

july2993: This is the temporal topology.yaml not about source  config of dm.

components/dm/ansible/testdata/ansible/inventory.ini
```diff
@@ -0,0 +1,36 @@
+## DM modules
+[dm_master_servers]
+dm_master ansible_host=172.19.0.101
+
+[dm_worker_servers]
+dm-worker1 ansible_host=172.19.0.101  server_id=101 source_id="mysql-replica-01" mysql_host=mysql1 mysql_user=root mysql_password='' mysql_port=3306
+
+dm-worker2 ansible_host=172.19.0.102  server_id=102 source_id="mysql-replica-02" mysql_host=mysql2 mysql_user=root mysql_password='' mysql_port=3306
```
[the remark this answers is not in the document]
july2993: Add in https://github.com/pingcap/tiup/pull/672/commits/e4ab90ede340c63536f853f27e32c4f34b33b65c
but it actually get from the config file of worker

components/dm/ansible/import.go
```diff
@@ -0,0 +1,605 @@
...
+		master.V1SourcePath = target
+
+		e := im.getExecutor(master.Host, master.SSHPort)
+		_, stderr, err := e.Execute("mkdir -p "+target, false)
+		if err != nil {
+			return errors.Annotatef(err, "failed to execute: %s", string(stderr))
+		}
+
+		for addr, source := range im.sources {
+			f, err := ioutil.TempFile("", "tiup-dm-*")
```
csuzhangxc, line 219: How about giving it a meaningful name? like `tiup-source-{source-id}.yaml`?

components/dm/command/import.go
csuzhangxc: I want to comment on this https://github.com/pingcap/tiup/pull/672#pullrequestreview-466687108. 😢

components/dm/ansible/import.go
july2993: as a temporal name, I think there's no need to do this, the target name in the master is '{host}:{port}.yml'(specified at L239, where host and port are the related worker's host and port.

csuzhangxc: OK

july2993:
@AstroProfundis @lonng PTAL

july2993:
> Please supplement the license header for new files.

Add by https://github.com/pingcap/tiup/pull/672/commits/e39f542b50f7251225d0bdefe5e1ee821959bf92

components/dm/ansible/import.go
```diff
@@ -0,0 +1,618 @@
...
+}
+
+func (im *Importer) getExecutor(host string, port int) (e executor.Executor) {
+	if im.testExecutorGetter != nil {
+		return im.testExecutorGetter.Get(host)
+	}
+
+	keypath := ansible.SSHKeyPath()
+
+	cfg := executor.SSHConfig{
```
lonng, line 152: Do we need to support native ssh?

july2993: added https://github.com/pingcap/tiup/pull/672/commits/bfc8fb609c90ee20a2e9a00bb0366d1b14560fe2

pkg/cluster/ansible/inventory.go
```diff
@@ -585,28 +585,36 @@ func readGroupVars(dir, filename string) (map[string]string, error) {
...
 		intPort, err := strconv.Atoi(port)
 		if err == nil {
-			sshPort = intPort
-		} // else just use the srv.Port
+			return intPort
+		}
 	}
-	return sshPort
+
+	// try to set value in global ansible config
```
lucklove, line 609:
```suggestion
	// try to get value from global ansible config
```

pkg/cluster/executor/local.go
```diff
@@ -0,0 +1,69 @@
...
+	command.Stderr = stderrBuf
+
+	err = command.Run()
+	stdout = stderrBuf.Bytes()
+	stderr = stderrBuf.Bytes()
+	return
+}
+
+// Transfer implements Executer interface.
+func (l *Local) Transfer(src string, dst string, download bool) error {
```
lucklove, line 57: I think we should call "cp -r"  in case of that the src and dst is a directory.

components/dm/ansible/import.go
```diff
@@ -0,0 +1,618 @@
...
+// ref https://docs.ansible.com/ansible/latest/reference_appendices/config.html#the-configuration-file
+// Changes can be made and used in a configuration file which will be searched for in the following order:
+//
+// ANSIBLE_CONFIG (environment variable if set)
+// ansible.cfg (in the current directory)
+// ~/.ansible.cfg (in the home directory)
+// /etc/ansible/ansible.cfg
+func searchConfigFile(dir string) (fname string, err error) {
+	// ANSIBLE_CONFIG (environment variable if set)
+	if v := os.Getenv("ANSIBLE_CONFIG"); len(v) > 0 {
```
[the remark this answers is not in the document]
july2993: No, search the file expected to be used, if it not exist, it should fail when we load it.

pkg/cluster/executor/local.go
july2993: depend on the expected behavior of `Transfer` in `Executor` interface`.
the default implementations we use `EasySSHExecutor` DO NOT support directory actually.
I will keep not support directory int this now in case of misuse.

PR #673, opened 2020-08-12
cluster: add support of TLS encryption for TiDB cluster

### LINKED ISSUE #529: Support deploying and managing TLS encryption enabled TiDB cluster
### Description

At present, when deploying a cluster with tiup-cluster, the tidb cluster (tidb <-> pd <-> tikv and pd <-> tiup) use plain messages to communicate, this may lead to potential security leaks. We should support TLS encryption as an option in the cluster topology to enable TLS encryption among components.

Similar support is already implemented in tidb-ansible and tidb-operator.

### Catagory

Security

### Value

Increase security of TiDB cluster, and avoid potential security leaks like MITM attack.

### TODO List

 - [x] Add support of TLS encryption in specs
    - [x] PD
    - [x] TiKV
    - [x] TiDB (Both server and client)
    - [x] Pump / Drainer
    - [x] CDC
    - [ ] binlog
    - [x] Conflict validation with unsupported components
 - [x] Generate TLS certificates and correct configs for components
 - [x] Support API calls with TLS encryption enabled

### Schedule

GanttStart: 2020-08-01
GanttDue: 2020-08-31
GanttProgress: 95%

AstroProfundis:

Add support for TLS encryption for TiDB cluster. Close #529 

 - [x] Specification checks for TLS enabled cluster
 - [x] Generate CA and certificate for cluster
 - [x] Deploy certificates to server
 - [x] Start/Stop/Restart TLS enabled cluster
 - [x] Scale in/out TLS enabled cluster
 - [x] Upgrade TLS enabled cluster
 - [x] Update monitoring configs for TLS enabled cluster
 - [x] Adjust `display` output for TLS enabled cluster
 - [x] Testing

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
Add support of TLS encryption for TiDB cluster
```


july2993 (commented):
> Error: failed to transfer CA cert to server: Process exited with status 1

pls fix ci

pkg/cluster/manager.go
```diff
@@ -406,8 +450,23 @@ func (m *Manager) Display(clusterName string, opt operator.Options) error {
 
 	// display cluster meta
 	cyan := color.New(color.FgCyan, color.Bold)
-	fmt.Printf("%s Cluster: %s\n", m.sysName, cyan.Sprint(clusterName))
-	fmt.Printf("%s Version: %s\n", m.sysName, cyan.Sprint(base.Version))
+	fmt.Printf("Cluster type:    %s\n", cyan.Sprint(m.sysName))
```
july2993, line 453: this seems redundant for the user.

pkg/cluster/spec/validate.go
```diff
@@ -659,9 +659,44 @@ func (s *Specification) validateTiSparkSpec() error {
...
+			ComponentTiDB,
+			ComponentTiKV,
+			ComponentPump,
+			ComponentDrainer,
+			ComponentCDC,
+			ComponentPrometheus,
+			ComponentAlertManager,
+			ComponentGrafana:
+		default:
+			return fmt.Errorf("component %s is not supported in TLS enabled cluster", c.Name())
```
july2993, line 687: use pingcap/errors Errorf  for stack trace

pkg/cluster/task/builder.go
```diff
@@ -364,6 +374,16 @@ func (b *Builder) DeploySpark(inst spec.Instance, version, srcPath, deployDir st
 	)
 }
 
+// TLSCert geenrates certificate for instance and transfer it to server
```
july2993, line 377: s/geenrates/generates

july2993: s/server/the server/ ?

pkg/cluster/task/tls.go
```diff
@@ -0,0 +1,112 @@
...
+
+// Rollback implements the Task interface
+func (c *TLSCert) Rollback(ctx *Context) error {
+	return ErrUnsupportedRollback
+}
+
+// String implements the fmt.Stringer interface
+func (c *TLSCert) String() string {
+	return fmt.Sprintf("TLSCert: host=%s role=%s cn=%s",
+		c.inst.GetHost(), c.inst.Role(), c.inst.ComponentName())
```
july2993, line 111: Role alway = ComponentName ?

pkg/cluster/spec/util.go
```diff
@@ -107,3 +109,27 @@ func ClusterMetadata(clusterName string) (*ClusterMeta, error) {
...
+}
+
+// statusByURL queries current status of the instance by http status api.
+func statusByURL(url string, tlsCfg *tls.Config) string {
+	client := utils.NewHTTPClient(statusQueryTimeout, tlsCfg)
+
+	// body doesn't have any status section needed
+	body, err := client.Get(url)
+	if err != nil {
+		return "Down"
```
july2993, line 129: I think we can add the error msg for more clarity？

pkg/cluster/task/tls.go
```diff
@@ -0,0 +1,112 @@
...
+	}), ""); err != nil {
+		return err
+	}
+
+	// transfer file to remote
+	e, ok := ctx.GetExecutor(c.inst.GetHost())
+	if !ok {
+		return ErrNoExecutor
+	}
+	if err := e.Transfer(caFile,
```
july2993, line 84: seems this API should be designed as `Transfer` and `Download` at first.

pkg/cluster/manager.go
AstroProfundis: It was `tidb Cluster: xxx` and `tidb Version: vx.y.z`

pkg/cluster/task/tls.go
AstroProfundis: No it's for example `tispark-master` as role and `tispark` as component name, identical for most components but not all.

pkg/cluster/spec/util.go
AstroProfundis: It has used to be, and been changed to this way for better understanding for users, having too many states might lead to confusion.

pkg/crypto/ca.go
```diff
@@ -0,0 +1,180 @@
...
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package crypto
+
```
july2993, line 15: pls use pingcap/errors

july2993 (commented):
I think the user should be able to specified the certificate&ca instead of all generated internally ?

AstroProfundis:
> I think the user should be able to specified the certificate&ca instead of all generated internally ?

That should be another feature request. In `tidb-operator` we support user specified certificate via cert manager, while in `tidb-ansible` only auto generated CA and certs are used.

pkg/crypto/ca.go
july2993: this is not resolved

pkg/crypto/ca.go
```diff
@@ -0,0 +1,180 @@
...
+	return x509.CreateCertificate(rand.Reader, template, ca.Cert, csr.PublicKey, ca.Key.Signer())
+}
+
+// ReadCA reads an existing CA certificate from disk
+func ReadCA(clsName, certPath, keyPath string) (*CertificateAuthority, error) {
+	// read private key
+	rawKey, err := ioutil.ReadFile(keyPath)
+	if err != nil {
+		return nil, fmt.Errorf("error reading CA private key for %s: %s", clsName, err)
```
july2993, line 140: use pingcap/errors
also many other place

july2993 (commented):
better not force push while reviewing? hard to find out what's updated

.github/workflows/ci.yaml
```diff
@@ -1,46 +0,0 @@
----
-
-name: ci
```
lonng, line 3: Why do you delete this file?

components/cluster/command/display.go
```diff
@@ -91,7 +92,14 @@ func displayDashboardInfo(clusterName string) error {
 		pdEndpoints = append(pdEndpoints, fmt.Sprintf("%s:%d", pd.Host, pd.ClientPort))
 	}
 
-	pdAPI := api.NewPDClient(pdEndpoints, 2*time.Second, nil)
+	var tlsCfg *tls.Config
```
lonng, line 95: There is so much duplicated code to check whether the `metadata.Topology.GlobalOptions.TLSEnabled` is true. It's better to define a method for `Topology`, e.g:
```
func (t *Topology) TSLConfig(path string)  (*tsl.Config, error) {
    if !t.GlobalOptions.TLSEnabled {
        return nil, nil
    }
    tlsCfg, err = spec.LoadClientCert(path)
    if err != nil {
        return perrs.AddStack(err)
    }
}

tslConfig, err := metadata.Topology.TSLConfig(tidbSpec.Path(clusterName, spec.TLSCertKeyDir))
```

.github/workflows/ci.yaml
AstroProfundis: To reduce the amount of CI tasks, I've moved the `check` to run in `integrate`.

components/cluster/command/display.go
```diff
@@ -54,20 +57,24 @@ func newDisplayCmd() *cobra.Command {
...
+			metadata, err := spec.ClusterMetadata(clusterName)
+			if err != nil && !errors.Is(perrs.Cause(err), meta.ErrValidate) &&
+				!errors.Is(perrs.Cause(err), spec.ErrNoTiSparkMaster) {
+				return perrs.AddStack(err)
+			}
 			if showDashboardOnly {
-				return displayDashboardInfo(clusterName)
+				tlsCfg, err := metadata.Topology.TLSConfig(tidbSpec.Path(clusterName, spec.TLSCertKeyDir))
+				if err != nil {
+					return perrs.AddStack(err)
```
lonng, line 68: It's better to add a stack in the function `metadata.Topology.TLSConfig` to avoid `perrs.AddStack` everywhere.

components/cluster/command/display.go
```diff
@@ -108,7 +115,21 @@ func displayDashboardInfo(clusterName string) error {
...
-	fmt.Println(u.String())
+
+	if tlsCfg != nil {
+		fmt.Println(
+			"Client certificate:",
+			color.CyanString(tidbSpec.Path(clusterName, spec.TLSCertKeyDir, spec.PFXClientCert)),
+		)
+		fmt.Println(
+			"Certificate password:",
+			color.CyanString(crypto.PKCS12Password),
```
lonng, line 126: I don't we need to print the password to the output.

components/cluster/command/display.go
```diff
@@ -108,7 +115,21 @@ func displayDashboardInfo(clusterName string) error {
 	}
 
 	u.Path = "/dashboard/"
-	fmt.Println(u.String())
+
+	if tlsCfg != nil {
+		fmt.Println(
```
lonng, line 120: Can you combine the two `fmt.Println`?

components/cluster/command/display.go
```diff
@@ -108,7 +115,21 @@ func displayDashboardInfo(clusterName string) error {
...
+			"Client certificate:",
+			color.CyanString(tidbSpec.Path(clusterName, spec.TLSCertKeyDir, spec.PFXClientCert)),
+		)
+		fmt.Println(
+			"Certificate password:",
+			color.CyanString(crypto.PKCS12Password),
+		)
+	}
+	fmt.Println(
```
lonng, line 129: Inline L129-L132.

components/dm/spec/alertmanager.go
```diff
@@ -107,16 +108,23 @@ func (i *AlertManagerInstance) Deploy(t *task.Builder, srcPath string, deployDir
...
 
 // InitConfig implement Instance interface
-func (i *AlertManagerInstance) InitConfig(e executor.Executor, clusterName, clusterVersion, deployUser string, paths meta.DirPaths) error {
+func (i *AlertManagerInstance) InitConfig(
+	e executor.Executor,
+	clusterName,
+	clusterVersion,
+	deployUser string,
+	paths meta.DirPaths,
+	enableTLS bool,
```
lonng, line 117: Why do we need the `enableTLS` parameter? We can retrieve it from `i.topo` directly.

pkg/cluster/manager.go
```diff
@@ -406,8 +436,23 @@ func (m *Manager) Display(clusterName string, opt operator.Options) error {
...
+	fmt.Printf("Cluster name:    %s\n", cyan.Sprint(clusterName))
+	fmt.Printf("Cluster version: %s\n", cyan.Sprint(base.Version))
```
lonng, line 441: The multiple `fmt.Printf` can be combined a single one.

pkg/cluster/manager.go
```diff
@@ -406,8 +436,23 @@ func (m *Manager) Display(clusterName string, opt operator.Options) error {
...
+	if topo.BaseTopo().GlobalOptions.TLSEnabled {
+		fmt.Printf("TLS encryption:  %s\n", cyan.Sprint("enabled"))
+		fmt.Printf("CA certificate:     %s\n", cyan.Sprint(
+			m.specManager.Path(clusterName, spec.TLSCertKeyDir, spec.TLSCACert),
+		))
+		fmt.Printf("Client private key: %s\n", cyan.Sprint(
+			m.specManager.Path(clusterName, spec.TLSCertKeyDir, spec.TLSClientKey),
+		))
+		fmt.Printf("Client certificate: %s\n", cyan.Sprint(
+			m.specManager.Path(clusterName, spec.TLSCertKeyDir, spec.TLSClientCert),
```
lonng, line 453: ditto

pkg/cluster/manager.go
```diff
@@ -601,6 +651,7 @@ func (m *Manager) Reload(clusterName string, opt operator.Options, skipRestart b
 			base.Version,
 			m.specManager,
 			inst, base.User,
+			topo.BaseTopo().GlobalOptions.TLSEnabled,
```
lonng, line 654: ditto

pkg/cluster/manager.go
```diff
@@ -1601,8 +1705,12 @@ func (m *Manager) confirmTopology(clusterName, version string, topo spec.Topolog
 	log.Infof("Please confirm your topology:")
 
 	cyan := color.New(color.FgCyan, color.Bold)
-	fmt.Printf("%s Cluster: %s\n", m.sysName, cyan.Sprint(clusterName))
-	fmt.Printf("%s Version: %s\n", m.sysName, cyan.Sprint(version))
+	fmt.Printf("Cluster type:    %s\n", cyan.Sprint(m.sysName))
+	fmt.Printf("Cluster name:    %s\n", cyan.Sprint(clusterName))
+	fmt.Printf("Cluster version: %s\n", cyan.Sprint(version))
```
lonng, line 1710: ditto

components/cluster/command/display.go
AstroProfundis: This is a hard coded password, and user need that to import the certificate to their browsers. It's by intend to be hard coded as the PKCS#12 encryption itself is weak and the password is not providing much extra protection, but just a compatibility feature.

pkg/cluster/manager.go
```diff
@@ -2020,3 +2188,64 @@ func refreshMonitoredConfigTask(
 	}
 	return tasks
 }
+
+func genAndSaveClusterCA(clusterName, tlsPath string) (*crypto.CertificateAuthority, error) {
```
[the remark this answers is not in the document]
lonng: +1

pkg/cluster/spec/validate.go
```diff
@@ -659,9 +659,44 @@ func (s *Specification) validateTiSparkSpec() error {
...
+		case ComponentPD,
+			ComponentTiDB,
+			ComponentTiKV,
+			ComponentPump,
+			ComponentDrainer,
+			ComponentCDC,
+			ComponentPrometheus,
+			ComponentAlertManager,
+			ComponentGrafana:
+		default:
```
lonng, line 686: TiFlash supports it after v4.0.4

pkg/cluster/manager.go
```diff
@@ -1034,13 +1118,18 @@ func (m *Manager) Deploy(
...
 		// Deploy component
 		// prepare deployment server
+		deployDirs := []string{
+			deployDir, logDir,
+			filepath.Join(deployDir, "bin"),
+			filepath.Join(deployDir, "conf"),
+			filepath.Join(deployDir, "scripts"),
+		}
+		if globalOptions.TLSEnabled {
+			deployDirs = append(deployDirs, filepath.Join(deployDir, "tls"))
```
lonng, line 1128: The `tls` directory should be deleted when destroying component with retaining data, see https://github.com/pingcap/tiup/blob/9a83b5f1513917c4561c759f9e2235d55f11e80c/pkg/cluster/operation/destroy.go#L312

components/cluster/command/display.go
AstroProfundis: It's a wrapped direct returning of etcd client's function, no named error variable in the call stack.

components/cluster/command/display.go
AstroProfundis: It could be better formatting by printing those content line by line instead of using a combined multi-line template, `go fmt` would force adding indent to the first line.

pkg/cluster/spec/validate.go
AstroProfundis: I didn't find any document about how to enable TLS for TiFlash, I think we could add that in another PR. It won't be too difficult after this PR been merged.

pkg/cluster/manager.go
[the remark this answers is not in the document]
AstroProfundis: We have test cases for `crypto.NewCA()` and `file.SaveFileWithBackup()` in their own packages, this function just call them in order and have no extra handling logic, I don't see any new cases necessary to be added.

lucklove (approved):
LGTM

lonng:
@AstroProfundis Please resolve the conflicts.

PR #676, opened 2020-08-12
Add advertise-status-addr for tiflash proxy

birdstorm:

Add `advertise-status-addr` for tiflash to support host name.

### Check List 

Tests 

 - No code

Release notes:

```release-note
Add `advertise-status-addr` for tiflash to support host name
```


birdstorm:
@solotzg PTAL

components/playground/instance/tiflash_proxy_config.go
```diff
@@ -43,11 +46,42 @@ data-dir = "%[6]s"
...
+[security]
+ca-path = ""
+cert-path = ""
+key-path = ""
+
+[server]
+addr = "0.0.0.0:%[4]d"
+advertise-addr = "%[2]s:%[4]d"
+engine-addr = "%[2]s:%[3]d"
+status-addr = "0.0.0.0:%[5]d"
```
solotzg, line 65: Maybe we can make some abstraction and reduce similar code

pkg/cluster/spec/tiflash.go
```diff
@@ -241,19 +242,26 @@ server_configs:
...
 
 // InitTiFlashLearnerConfig initializes TiFlash learner config file
-func (i *TiFlashInstance) InitTiFlashLearnerConfig(cfg *scripts.TiFlashScript, src map[string]interface{}) (map[string]interface{}, error) {
+func (i *TiFlashInstance) InitTiFlashLearnerConfig(cfg *scripts.TiFlashScript, clusterVersion string, src map[string]interface{}) (map[string]interface{}, error) {
 	topo := Specification{}
+	var statusAddr string
 
 	firstDataDir := strings.Split(cfg.DataDir, ",")[0]
 
+	if semver.Compare("v4.0.5", clusterVersion) <= 0 {
```
lonng, line 251: Prefer 
```suggestion
	if semver.Compare(clusterVersion, "v4.0.5") >= 0 {
```

july2993: > An invalid semantic version string is considered less than a valid one.

And "nightly" is an invalid one, do we need to handle it specifically?

birdstorm: nightly version is not affected. This config change is usable since v4.0.4, so is nightly.

birdstorm:
@lonng PTAL again, thanks.

Issue #677, opened 2020-08-13
Add cluster user deploy cluster feature，skip create user


WentaoJin:
1、In the existing tiup deployment logic --user will log in as the ssh user, then create the cluster user in the configuration, and then distribute the deployment. Some users only have the authority to manage and deploy the tidb cluster, but they do not have the authority to create users and other permissions. Deploying the cluster in this way will report an error
![image](https://user-images.githubusercontent.com/34805682/90095584-abad8d00-dd63-11ea-974e-92aee6a9ad21.png)

2、hope to support the direct designation of cluster users in the topology to deploy the cluster to avoid this deployment problem

PR #678, opened 2020-08-13
Improve env init on deploying and scale-out cluster

lucklove:
At this time, we will create the user which is specified in topology.yml on `deploy` and `scale-out` stage.
However, the topology.yml only allows specifying the user name excepting the user group name. Sometimes the user may want to create the user, but using an existed group name.

BTW: fix https://github.com/pingcap/tiup/issues/677
Add a `group` field in the topology file

### Check List 

Tests 

 - Integration testing

Release notes:

```release-note
Support specifying user group or skipping creating a user in the deploy and scale-out stage
- to specify the group: https://github.com/pingcap/tiup/blob/master/examples/topology.example.yaml#L7
- to skip creating the user: `tiup cluster deploy/scale-out --skip-create-user xxx`
```


pkg/cluster/task/mkdir.go
```diff
@@ -31,7 +31,7 @@ type Mkdir struct {
 func (m *Mkdir) Execute(ctx *Context) error {
 	exec, found := ctx.GetExecutor(m.host)
 	if !found {
-		return ErrNoExecutor
+		panic(ErrNoExecutor)
```
lonng, line 34: Why panic?

pkg/cluster/module/user.go
```diff
@@ -48,6 +49,7 @@ var (
 type UserModuleConfig struct {
 	Action string // add, del or modify user
 	Name   string // username
+	Group  string // group name
```
lonng, line 52: There is no command line flag to assign a value to `Group`.

pkg/cluster/task/mkdir.go
lucklove: Same with https://github.com/pingcap/tiup/blob/master/pkg/cluster/task/env_init.go#L54

I think it's easier to debug...

pkg/cluster/module/user.go
lucklove: There is no need to add them to command line because the EnvInit only read user from topology file.

pkg/cluster/task/mkdir.go
lonng: Both of them should be changed to Err, panic is always a bad idea in our scenarios and we shouldn't display the stack trace to users.

lucklove: Ok

pkg/cluster/module/user.go
```diff
@@ -80,10 +82,19 @@ func NewUserModule(config UserModuleConfig) *UserModule {
...
+
+		// groupadd -f <group-name>
+		groupAdd := fmt.Sprintf("%s -f %s", groupaddCmd, config.Group)
+
+		// useradd -g <group-name> <user-name>
+		cmd = fmt.Sprintf("%s -g %s %s", cmd, config.Group, config.Name)
 
 		// prevent errors when username already in use
-		cmd = fmt.Sprintf("id -u %s > /dev/null 2>&1 || %s", config.Name, cmd)
+		cmd = fmt.Sprintf("%s && (id -u %s > /dev/null 2>&1 || %s)", groupAdd, config.Name, cmd)
```
lonng, line 97: Will we create the group even though the user exists?

pkg/cluster/task/mkdir.go
lonng: Better to revert this part.

PR #689, opened 2020-08-17
Make it clear when transfer failed using native ssh

lucklove:
This is usefull for debug

Before:
image

Optimized:
image

Return error from remote

### Check List 

Tests 

Release notes:

```release-note
Support debug scp with --natvie-ssh flag
```


lonng:
The optimized output is more understandable, but it will be better if you format the output with some indent.

PR #690, opened 2020-08-18
Support saving the raw toml as a string field in topology

july2993:
Support saving the raw toml as a string filed in topo

Add `TomlConfig` in pkg/cluster/spec/config.go which compatible with the old format.
Replace all the config field to use `TomlConfig` instead of `map[string]interface{}`
Alway `Marshal` as a string field with the raw toml data.

Note, if merge this, after edit-config, all config will be converted to the raw toml format.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Related changes

Release notes:

```release-note
Support saving the raw toml config data as a string field in topology
```


pkg/cluster/spec/tispark.go
```diff
@@ -31,18 +31,21 @@ import (
 
 // TiSparkMasterSpec is the topology specification for TiSpark master node
 type TiSparkMasterSpec struct {
```
july2993, line 33: change nothing, just add a comment, no sure whether we should change it yet.

pkg/cluster/ansible/import.go
```diff
@@ -84,7 +84,8 @@ func parseInventoryFile(invFile io.Reader) (string, *spec.ClusterMeta, *aini.Inv
 			clsMeta.Version = host.Vars["tidb_version"]
 
 			if enableBinlog, err := strconv.ParseBool(host.Vars["enable_binlog"]); err == nil && enableBinlog {
-				clsMeta.Topology.ServerConfigs.TiDB["binlog.enable"] = enableBinlog
+				clsMeta.Topology.ServerConfigs.TiDB = new(spec.TomlConfig)
```
lonng, line 87: If the `clsMeta.Topology.ServerConfigs.TiDB` isn't nil, we should  use `clsMeta.Topology.ServerConfigs.TiDB.Set()` to overwrite the configuration.

pkg/cluster/spec/config.go
```diff
@@ -0,0 +1,97 @@
+package spec
```
lonng, line 1: The license header is missing.

pkg/cluster/spec/config.go
```diff
@@ -0,0 +1,97 @@
...
+		c.mp, err = merge(value)
+		if err != nil {
+			return errors.AddStack(err)
+		}
+	default:
+		return errors.Errorf("unknown type: %v", reflect.TypeOf(value))
+
+	}
+
+	return nil
```
lonng, line 50: We should unflatten the unshelled `map[string]interface{}` because we can guarantee all keys is `xxx.yyy` form. e.g:
```yaml
server_configs:
  pd:
    xxx.yyy: 1
    zzz:
      z1: 2
      z2: 3
```

pkg/cluster/spec/config_test.go
```diff
@@ -0,0 +1,115 @@
+package spec
```
lonng, line 1: ditto.

pkg/cluster/spec/config.go
```diff
@@ -0,0 +1,97 @@
...
+// Set the config item.
+// k can be the format with "."
+func (c *TomlConfig) Set(k string, v interface{}) {
+	if c.mp == nil {
+		c.mp = make(map[string]interface{})
+	}
+	c.mp[k] = v
+
+	var err error
+	c.mp, err = flattenMap(c.mp)
```
lonng, line 93: I made a mistake to name the function `flattenMap` ever. The correct name should be `unflattenMap`.

pkg/cluster/spec/config_test.go
```diff
@@ -0,0 +1,115 @@
...
+	err = yaml.Unmarshal(out, &b)
+	assert.Nil(err)
+	bout, err := yaml.Marshal(&a)
+	assert.Nil(err)
+	assert.Equal(expectOut, string(bout))
+}
+
+func TestTomlConfigMap(t *testing.T) {
+	out := `a: 1
+config:
```
lonng, line 47: Add some test cases like:
```yaml
server_configs:
  pd:
    xxx.yyy: 1
    zzz:
      z1: 2
      z2: 3
```

PR #695, opened 2020-08-19
Allow empty key file

lucklove:
As https://github.com/pingcap/tiup/pull/615 described, we may need to use some ssh plugin, so the native ssh client is supported. But there is a little bug: *tiup-cluster doesn't allow empty `-i` flag and it has a default value*. So the `-i` flag will always be passed to ssh client unless the user use password. In this case, the LDAP don't work.

This pull request allows the user to spec `-i ""` and the ssh client will not receive a `-i` flag in this case.


lucklove:
Document about this: https://github.com/pingcap/docs-cn/pull/4303

lucklove:
TODO: use "" as the default value if --native-ssh selected

pkg/utils/utils.go
```diff
@@ -27,3 +29,14 @@ func JoinInt(nums []int, delim string) string {
 	}
 	return strings.TrimSuffix(result, delim)
 }
+
+// IsFlagSetByUser check if the a flag is set by user explicitly
+func IsFlagSetByUser(flagSet *pflag.FlagSet, flagName string) bool {
+	setByUser := false
```
AstroProfundis, line 35: It's not necessary to use a named variable here, just return true and false would be enough.

AstroProfundis (approved):
LGTM

lonng:
@lucklove Please resolve the conflicts.

PR #696, opened 2020-08-20
Do not overwrite alertmanager.yml for dm when InitConfig

july2993:
when reload or scale the altermanager.yml will be overwritten. 

never overwrite altermanager.yml, only init the example config when deploying the instance.

also, handle the custom deploying an instant in upgrade and scale-out
### Check List 

Tests 
deploy a cluster and update the config -> try reload -> check config file not overwritten

```release-note
NONE
```

### DIFF
--- a/pkg/cluster/manager.go
+++ b/pkg/cluster/manager.go
@@ -730,20 +730,24 @@ func (m *Manager) Upgrade(clusterName string, clusterVersion string, opt operato
 			// backup files of the old version
 			tb = tb.BackupComponent(inst.ComponentName(), base.Version, inst.GetHost(), deployDir)
 
-			// copy dependency component if needed
-			switch inst.ComponentName() {
-			case spec.ComponentTiSpark:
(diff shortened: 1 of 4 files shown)


pkg/cluster/manager.go
```diff
@@ -730,20 +730,24 @@ func (m *Manager) Upgrade(clusterName string, clusterVersion string, opt operato
...
-					inst.GetHost(),
-					deployDir,
-				)
+			if deployerInstance, ok := inst.(DeployerInstance); ok {
+				deployerInstance.Deploy(tb, "", deployDir, version, clusterName, clusterVersion)
+			} else {
+				// copy dependency component if needed
+				switch inst.ComponentName() {
+				case spec.ComponentTiSpark:
+					tb = tb.DeploySpark(inst, version, "" /* default srcPath */, deployDir, m.bindVersion)
```
lonng, line 739: It's better to implement the `Deployer` interface for the TiSpark instance.

july2993: Will do this in later pr

PR #698, opened 2020-08-20
Remove component specified in tryURLs

july2993:
Screen Shot 2020-08-20 at 4 24 13 PM

Remote component specified in tryURLs

also, add the `name` in example
Release notes:

```release-note
NONE
```

### DIFF
--- a/examples/dm/topology.example.yaml
+++ b/examples/dm/topology.example.yaml
@@ -20,6 +20,7 @@ server_configs:
 
 master_servers:
   - host: 10.0.1.11
+    name: master1
     # ssh_port: 22
     # port: 8261
     # peer_port: 8291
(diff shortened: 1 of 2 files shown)


examples/dm/topology.example.yaml
```diff
@@ -34,7 +35,9 @@ master_servers:
       # rpc-rate-limit: 10.0
       # rpc-rate-burst: 40
   - host: 10.0.1.18
+    name: master2
```
lonng, line 38: Can we set a default name for master node?

july2993: will set as some kind like {host}:{port} as default if it's empty, but I recommend to set it explicitly by the user.

lonng (approved):

PR #704, opened 2020-08-21
cluster/import: read and set deploy_dir properly on import

AstroProfundis:
The global `deploy_dir` is not set during the process of importing from ansible deployed cluster, however, as we detect and set deploy dir for each instance individually, this only affect monitoring agents (node_exporter and blackbox_exporter).

Read the value of `deploy_dir` in `inventory.ini` and set it to `global.deploy_dir` and `monitored.deploy_dir` global variables.

Note that as we read and set dirs for each instance individually, this change does not effect them.

### Check List 

Tests 

 - Unit test

Code changes

 - Has persistent data change

Side effects
None

Related changes

Release notes:

```release-note
Set correct `deploy_dir` of monitoring agents when importing ansible deployed clusters
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -50,7 +50,7 @@ require (
 	github.com/pingcap/pd/v4 v4.0.0
 	github.com/pingcap/tidb-insight v0.3.1
 	github.com/r3labs/diff v0.0.0-20200627101315-aecd9dd05dd2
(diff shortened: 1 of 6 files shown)


lonng (approved):
LGTM

PR #705, opened 2020-08-22
fix the hard-coded parameter: storage.tsdb.retention

darkelf21cn:
There is a configuration `storage_retention` under `monitoring_servers` section in the meta.toml file which allows users to change the data retention period of Prometheus. However, The variable `--storage.tsdb.retention` in the Prometheus template has a hard-code value `30d`. So the configuration never takes effect. 

Add `WithRetention` to the `PrometheusScript` structure to set the data retention in days. It still uses `30d` by default when the parameter is not configured or the value is not valid. Otherwise, it will use the given value to configure the Prometheus.

### Check List 

Tests 

Manual test (add detailed scripts or steps below)

Did tests below and validated the result by reading the value of `--storage.tsdb.retention` in the output from the `ps aux | grep prometheus` command.
- remove the `storage_retention` from the `meta.toml` and run `./tiup-cluster reload test`
- set an invalid value `45y` to the `storage_retention` and run `./tiup-cluster reload test`
- set an valid value `45d` to the `storage_retention` and run `./tiup-cluster reload test`
Monitor the result by reading the `--storage.tsdb.retention` output from the `ps aux | grep prometheus` command

Code changes

 - Has persistent data change (template file)

Side effects
 - No effects are expected

Related changes

Release notes:

```release-note
NONE
```


lonng:
@darkelf21cn Thanks for your contribution. 👍

darkelf21cn:
> @darkelf21cn Thanks for your contribution. 👍

My pleasure.

PR #707, opened 2020-08-24
cluster: fix some issues of tidb cluster with tispark component

AstroProfundis:
Fix some issues found during test.

 - Fix component role handling, `-R` in sub commands now works properly on tispark nodes
 - Fix wrong validation tag for tispark config fields
 - Fix listening address handling for tispark workers

### Check List 

Tests 

 - Unit test

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```


pkg/cluster/spec/spec.go
```diff
@@ -202,7 +202,12 @@ func (s *Specification) BaseTopo() *BaseTopo {
 func AllComponentNames() (roles []string) {
 	tp := &Specification{}
 	tp.IterComponent(func(c Component) {
-		roles = append(roles, c.Name())
+		switch c.Name() {
+		case ComponentTiSpark: // tispark-{master, worker}
```
lonng, line 206: I think we should treat `tispark-{master,worker}` as a different component type.

lonng: And we don't need `Role` method anymore.

AstroProfundis: We need something called `tispark` to download the package

pkg/cluster/spec/tispark.go
```diff
@@ -39,8 +39,8 @@ type TiSparkMasterSpec struct {
 	WebPort      int                    `yaml:"web_port" default:"8080"`
 	DeployDir    string                 `yaml:"deploy_dir,omitempty"`
 	JavaHome     string                 `yaml:"java_home,omitempty" validate:"java_home:editable"`
-	SparkConfigs map[string]interface{} `yaml:"spark_config,omitempty" validate:"spark_config:editable"`
-	SparkEnvs    map[string]string      `yaml:"spark_env,omitempty" validate:"spark_env:editable"`
+	SparkConfigs map[string]interface{} `yaml:"spark_config,omitempty" validate:"spark_config:ignore"`
```
lonng, line 42:
```suggestion
	SparkConfigs map[string]interface{} `yaml:"spark_config,omitempty" validate:"-"`
```

pkg/cluster/spec/tispark.go
```diff
@@ -39,8 +39,8 @@ type TiSparkMasterSpec struct {
...
+	SparkEnvs    map[string]string      `yaml:"spark_env,omitempty" validate:"spark_env:ignore"`
```
lonng, line 43: ditto

pkg/cluster/spec/tispark.go
AstroProfundis: `-` does not make any sense. The `validate` tag is handled in `pkg/utils/diff.go` and only `ignore` makes the content of these config fields freely editable.

lonng (approved):
LGTM
But I think the `Instance` interface is getting bigger, we should simplify it.

PR #709, opened 2020-08-24
cluster/check: fix checking of existing cluster

AstroProfundis:
 - `check` sub command is not working properly with existing cluster
 - checks are not performed for instance with no data dir defined

 - Fix credential settings for existing clusters
 - Handle data dir properly
 - `ss -ltn` may report multiple records for on port (with different fd, may because of `TCP_PORT_REUSE`), only one is enough for the check

### Check List 

Tests 

Code changes

Side effects
None

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/components/cluster/command/check.go
+++ b/components/cluster/command/check.go
@@ -184,6 +184,7 @@ func checkSystemInfo(s *cliutil.SSHConnectionProps, topo *spec.Specification, op
 
 			// build checking tasks
 			t2 := task.NewBuilder().
+				// check for general system info
 				CheckSys(
 					inst.GetHost(),
 					"",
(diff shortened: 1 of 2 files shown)


components/cluster/command/check.go
```diff
@@ -180,86 +182,88 @@ func checkSystemInfo(s *cliutil.SSHConnectionProps, topo *spec.Specification, op
...
+					inst.GetHost(),
+					"ss -lnt",
+					false,
+				).CheckSys(
+				inst.GetHost(),
+				"",
+				task.CheckTypePort,
+				topo,
+				opt.opr,
+			).
```
lonng, line 211: Seems the format is incorrect?

lonng (approved):
LGTM

PR #713, opened 2020-08-25
cluster: set owner to current user when decompressing tarballs

AstroProfundis:
Make sure the decompressed files are owned by the current user

Add `--no-same-owner` flag to `tar` command when decompressing

### Check List 

Tests 

 - Integration test

Code changes

 - Has persistent data change

Side effects
None

Related changes

Release notes:

```release-note
NONE
```


july2993 (commented):
LGTM
can you describe the case this will case problem

lonng (approved):

AstroProfundis:
I don't have a specific case but our DBA fellow informs me that they've had seen directory permission issues when decompressing tarballs without this argument.

PR #515, opened 2020-06-19
Redirect std output to log file

july2993:
1. redirect std output to the log file, only print the latest 10 lines of
output form log file if the instance quit abnormally.  maybe fix https://github.com/pingcap/tiup/issues/342
2. make the scale-in/scale-out/display not hidden and improve help
message.
3. enable monitor default. fix #491 
4. add uptime in display

example of 1:
Screen Shot 2020-06-19 at 3 10 42 PM

example of 4:
Screen Shot 2020-06-19 at 4 32 59 PM

PR #715, opened 2020-08-26
Improve playground and test

july2993:
- Fix the issue that playground display can not show correct uptime status if the process is exited.
- Fix the issue that playground keep silent if a process is exited with non zero exit code, It should Tail the latest log after #515
- better make sure kill all process and improve test.

this is broken after #588 

Do not call the Cmd.Process.Wait directly or we can not get the
Cmd.ProcessState and the Uptime show in display can't find out
the process is quit. Also, Cmd.Process return the ProcessState and
nil if the process exit with not 0 code, so if the instance quits
unexpectedly will tail the last log of the instance will never work.

Setup signal handler before we boot the cluster. Before this PR, in bootCluster, if
return before we setup the signal handler and receive signal(like user
press ctrl-c), playground will quit directly, but we may have started
some subprocess which become orphaned process.

Improve the integration test to make sure scale-in/out, display
basically work.

### Check List 

Tests 

 - Integration test

Release notes:

```release-note
NONE
```


components/playground/grafana.go
```diff
@@ -149,6 +152,7 @@ func makeSureDir(fname string) error {
 var clusterName string = "playground"
 
 // dir should contains files untar the grafana.
+// return not error iff the Cmd is started successfully.
```
lonng, line 155:
```suggestion
// return not error if the Cmd is started successfully.
```

july2993: iff = if and only if

components/playground/instance/process.go
```diff
@@ -50,8 +57,9 @@ func (p *process) Pid() int {
 // Uptime implements Instance interface.
 func (p *process) Uptime() string {
 	s := p.cmd.ProcessState
-	if s != nil && s.Exited() {
-		return "exited"
+
+	if s != nil {
```
lonng, line 61: Can we retrieve the `ProcessState` if the current process is running? If so this method won't return the real `Uptime`.

components/playground/main.go
```diff
@@ -126,7 +130,66 @@ Examples:
...
+				go p.terminate(sig)
+				// If user try double ctrl+c, force quit
+				sig = (<-sc).(syscall.Signal)
+				atomic.StoreInt32(&p.curSig, int32(syscall.SIGKILL))
+				if sig == syscall.SIGINT {
+					p.terminate(syscall.SIGKILL)
+				}
+			}()
+
+			// TODO: we can set Pdeathsig of Cmd.SysProcAttr(linux only) in all the Cmd we started to kill
```
lonng, line 171: Can we implement this in the current PR?

components/playground/playground.go
```diff
@@ -46,27 +44,33 @@ import (
...
+	tidbs            []*instance.TiDBInstance
+	tiflashs         []*instance.TiFlashInstance
+	pumps            []*instance.Pump
+	drainers         []*instance.Drainer
+	startedInstances []instance.Instance
 
 	idAlloc        map[string]int
 	instanceWaiter errgroup.Group
 
+	// not nil iff we start the exec.Cmd successfully.
```
lonng, line 69:
```suggestion
	// not nil if we start the exec.Cmd successfully.
```

components/playground/instance/process.go
july2993: 131     // ProcessState contains information about an exited process,
132     // available after a call to Wait or Run.
133     ProcessState *os.ProcessState

It is only available after the proccss exited and we call Wait on the Cmd.

components/playground/main.go
july2993: #719 #720 
I will try but not in this PR.
Keeping this pr small and do one thing.

lonng (approved):
LGTM

Issue #703, opened 2020-08-21
wrong output of `tiup dm help`


csuzhangxc:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?

```
~ tiup dm help
Starting component `dm`: /home/zhangxc/.tiup/components/dm/v1.0.9/tiup-dm help
Deploy a DM cluster for production

Usage:
  tiup cluster cluster [flags]
  tiup cluster [command]
```

2. What did you expect to see?

```
Usage:
  tiup dm [command]
```

3. What did you see instead?

```
Usage:
  tiup cluster cluster [flags]
  tiup cluster [command]
```

4. What version of TiUP are you using (`tiup --version`)?

```
~ tiup --version
v1.0.9 tiup
Go Version: go1.13
Git Branch: master
GitHash: 41fbacf
```

PR #716, opened 2020-08-26
Fix showing  misleading usage for cluster and dm

july2993:
Set the magic arg0 for dm. Remove RunE for both cluster and dm because
they are just not runable. Remove the wrong custom UseLine although we nolonger
show it.(the raw useline we get "tiup cluster [flags]", the custom
useline will be "tiup cluster cluster [flags]")

fix https://github.com/pingcap/tiup/issues/703

    Set the magic arg0 for dm. Remove RunE for both cluster and dm because
    they are just not runable. Remove the wrong custom UseLine although we nolonger
    show it.(the raw useline we get "tiup cluster [flags]", the custom
    useline will be "tiup cluster cluster [flags]")

### Check List 

Tests 

Screen Shot 2020-08-26 at 7 17 43 PM
Screen Shot 2020-08-26 at 7 18 00 PM

Release notes:

```release-note
NONE
```

### DIFF
--- a/components/cluster/command/root.go
+++ b/components/cluster/command/root.go
@@ -121,9 +121,6 @@ func init() {
 
 			return nil
 		},
-		RunE: func(cmd *cobra.Command, args []string) error {
-			return cmd.Help()
-		},
 		PersistentPostRunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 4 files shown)


lucklove (approved):
LGTM

PR #717, opened 2020-08-27
cluster: restart prometheus after scaling the cluster

AstroProfundis:
The prometheus config is not reloaded correctly after scaling in a cluster.

Restart the prometheus instance after updating configs, otherwise new configs are not applied.

### Check List 

Tests 

 - Integration test

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```


lonng (approved):
LGTM

PR #682, opened 2020-08-14
Do not change the run component's wd

july2993:
### What problem does this PR solve? 
Do not change the run component's wd

### Check List 

Tests 

 - Manual test (add detailed scripts or steps below)
before: (main.go exist in the current directory)
```
root@control:/tiup-cluster# tiup dmctl:nightly --config ./main.go
Starting component `dmctl`: /root/.tiup/components/dmctl/nightly/dmctl/dmctl --config ./main.go
parse cmd flags err: open ./main.go: no such file or directory
```

after:
```
root@control:/tiup-cluster# tiup dmctl:nightly --config ./main.go
Starting component `dmctl`: /root/.tiup/components/dmctl/nightly/dmctl/dmctl --config ./main.go
parse cmd flags err: Near line 0 (last key parsed ''): bare keys cannot contain '/'
Error: run `/root/.tiup/components/dmctl/nightly/dmctl/dmctl` (wd:) failed: exit status 2
```

Release notes:

```release-note
NONE
```

PR #718, opened 2020-08-27
Distinct instanceDir and wd

july2993:
Fix the issue that `tiup_process_meta` is generated at the wd and `tiup status` can not work.

After #682, we do not change wd for the run components.
It make it save the tiup_process_meta file in wd instead of
InstanceDataDir like before. This seems broke "tiup clean" and "tiup
status" which check all subdirectory in ~/.tiup/data for the
tiup_process_meta.

After this PR, we Distinct InstanceDataDir and wd in launchComponent and
make Dir in tiup_process_meta as the InstanceDataDir instead of wd.

### Check List 

Tests 

run `tiup tidb` and check `tiup status`

Release notes:

```release-note
NONE
```


pkg/exec/run.go
```diff
@@ -135,7 +135,7 @@ func PrepareCommand(
 	ctx context.Context,
 	component string,
 	version v0manifest.Version,
-	binPath, tag, wd string,
+	binPath, tag, instanceDir string, wd string,
```
lonng, line 138: Seems we don't need `wd` anymore.

july2993: playground is using this `PrepareCommand` too and set the wd as ~/.data/{playground}/{instance}

lonng (approved):
LGTM

PR #724, opened 2020-08-28
Keep deploy dir for worker while import from ansible

july2993:
dm-worker 1.0 will write some data  at  {wd}/dm_worker_meta.
and 2.0 bootstrap need this directory(after bootstrap, this directory will be deleted by worker)
So we must always keep using the same wd when upgrade to 2.0

Deploy dir MUST always keep the same and CAN NOT change.
dm-worker will save the data in the wording directory and there's no configuration
to specify the directory.
We will always set the wd as DeployDir.

also, mark `cluster-version` required instead of nightly default.

### Check List 

Tests 

 - Unit test
 

Release notes:

```release-note
NONE
```

### DIFF
--- a/components/dm/ansible/import.go
+++ b/components/dm/ansible/import.go
@@ -407,7 +407,12 @@ func (im *Importer) ImportFromAnsibleDir() (clusterName string, meta *spec.Metad
 					}
 				}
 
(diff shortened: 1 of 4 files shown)


july2993:
@csuzhangxc PTAL

lucklove (approved):
LGTM

csuzhangxc (commented):
LGTM

Issue #112, opened 2020-04-08
Traking issue: developer support


nrc:
Running and testing TiDB and TiKV when developing is tricky. TiUp has potential to make it easy and thus to encourage contribution. This issue is for tracking ideas for ways in which TiUp could help developers. Good ideas should be spun out into their own issues.

nrc:
Some initial ideas:

* Specify that specified or all nodes in a cluster should use a locally supplied (presumably built by the dev) version of TiDB/TiKV/PD
* Support for automatically setting up clusters and running common benchmark suites (e.g., YCSB, sysbench). Seems like the tpc component does this?
* run tools like a debugger or perf
* support for finding and reading logs
* scripting for setting up clusters

lonng:
@nrc Great ideas.

lonng:
I will close this issue because most of them have been implemented.

PR #726, opened 2020-08-28
Support ycsb bench

longfangsong:

Currently tiup bench just support the tpc series benchmarking, which benches both tidb and tikv, but sometimes we just want to bench tikv.

[go-ycsb](https://github.com/pingcap/go-ycsb) can do the bench, but it is not very convieniet to run go-ycsb, we need to pull its code and `make` it, and pass in a lot of properties which is hard to remember and frustrating.

So I think it will be cool to support do ycsb benchmarking directly in tiup.

This is part of #112.

Add supprot for ycsb.

### Check List 

Tests 

  - Do the bench with `tiup bench ycsb prepare` and `tiup bench ycsb run`

Code changes

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
Support benchmarking tikv with `go-ycsb`
```


longfangsong:
Current run
```shell
tiup bench ycsb prepare
```
and 
```
tiup bench ycsb run
```
will make it work.

However, it's my first time to contribute to TiUP, I'm not familiar with the code base, so I need help with how can I make this feature a sound one. Thank you in advance.

components/bench/ycsb.go
```diff
@@ -0,0 +1,132 @@
+package main
```
lonng, line 1: The license header is missing.

components/bench/ycsb.go
```diff
@@ -0,0 +1,145 @@
...
+	switch action {
+	case "prepare":
+		_, _, _ = configProp.Set("dotransactions", "false")
+	case "run":
+		_, _, _ = configProp.Set("dotransactions", "true")
+	}
+	workloadCreator := ycsb.GetWorkloadCreator("core")
+	measurement.InitMeasure(configProp)
+	workload, err := workloadCreator.Create(configProp)
+	if err != nil {
```
july2993, line 133: return error instead of panic?
We can use `RunE` of Command instead of `Run`

components/bench/ycsb.go
```diff
@@ -0,0 +1,145 @@
...
+		panic(err)
+	}
+	dbCreator := ycsb.GetDBCreator("tikv")
+	db, err := dbCreator.Create(configProp)
+	if err != nil {
```
july2993, line 138: ditto

components/bench/ycsb.go
```diff
@@ -0,0 +1,145 @@
...
+	cmd.AddCommand(cmdPrepare, cmdRun)
+	root.AddCommand(cmd)
+}
+
+func executeYcsb(action string) {
+	runtime.GOMAXPROCS(maxProcs)
+	configProp := config.ToProperties()
+	switch action {
+	case "prepare":
+		_, _, _ = configProp.Set("dotransactions", "false")
```
july2993, line 126: do we need to check the err?

components/bench/ycsb.go
```diff
@@ -0,0 +1,145 @@
...
+	ReadModifyWriteProportion float32
+
+	RequestDistribution string
+	Verbose             bool
+	Pd                  string
+	ConnCount           uint32
+	BatchSize           uint32
+}
+
+func (config ycsbConfig) ToProperties() *properties.Properties {
```
july2993, line 47: do we need to check the err?

components/bench/ycsb.go
longfangsong: The only circumstances an error can occur is "Property expansion", since we just use properties internal here, I think instead of checking error here, we should disable the expansion instead.

components/bench/ycsb.go
```diff
@@ -0,0 +1,195 @@
...
+			for _, content := range []string{"a", "b", "c", "d", "e", "f"} {
+				if config.PropertyFile == content {
+					config.PropertyFile = commonWorkloadURL + content
+					break
+				}
+			}
+		}
+
+		if strings.HasPrefix(config.PropertyFile, "http") {
+			result = properties.MustLoadURL(config.PropertyFile)
```
july2993, line 66: Does this mean if user passes a wrong url or file will panic?
prefer LoadURL and handle err

components/bench/ycsb.go
```diff
@@ -0,0 +1,195 @@
...
+		} else {
+			result = properties.MustLoadFile(config.PropertyFile, properties.UTF8)
```
july2993, line 68: ditto

july2993 (commented):
rest LGTM

july2993 (approved):
LGTM
Thanks for your contribution.

PR #733, opened 2020-08-31
Set Pdeathsig to kill subprocess if playground dead

### LINKED ISSUE #720: set Pdeathsig as KILL for all Command start by playground
## Feature Request

- start playground
- kill -9 playground
All the subprocess of playground orphaned child process adopted by init.

We can set Pdeathsig(Linux only) as KILL for all Command start by the playground to kill all processes.
ref: https://golang.org/pkg/syscall/#SysProcAttr
and https://medium.com/@ganeshmaharaj/clean-exit-of-golangs-exec-command-897832ac3fa5

july2993:
fix #720 
Set Pdeathsig to kill subprocess if playground dead
After this pr, if playground dead and some subprocess still running,
the subprocess will be killed by OS instead of becoming an orphaned child
adopted by init. This feature is linux only.

### Check List 

Tests 

before this pr:
Screen Shot 2020-08-31 at 2 57 45 PM
after this pr:

Release notes:

```release-note
Set Pdeathsig to kill subprocess if playground dead
```

### DIFF
--- a/pkg/exec/run.go
+++ b/pkg/exec/run.go
@@ -137,6 +137,7 @@ func PrepareCommand(
 	version v0manifest.Version,
 	binPath, tag, instanceDir string, wd string,
 	args []string,
+	sysProcAttr *syscall.SysProcAttr,
 	env *environment.Environment,
 	checkUpdate ...bool,
 ) (*exec.Cmd, error) {
@@ -226,6 +227,7 @@ func PrepareCommand(
 	c.Stdout = os.Stdout
 	c.Stderr = os.Stderr
 	c.Dir = wd
+	c.SysProcAttr = sysProcAttr
 
 	return c, nil
 }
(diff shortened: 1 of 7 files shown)


pkg/exec/run.go
```diff
@@ -137,6 +137,7 @@ func PrepareCommand(
 	version v0manifest.Version,
 	binPath, tag, instanceDir string, wd string,
 	args []string,
+	sysProcAttr *syscall.SysProcAttr,
```
lonng, line 140: The function `PrepareCommand` has 11 parameters already, it's better to use a struct to define. How do you think?

lonng (approved):
Rest LGTM

lonng:
Is any solution for macOS to handle the same scenario?

july2993:
> Is any solution for macOS to handle the same scenario?

No as far as I know.

pkg/exec/run.go
july2993: https://github.com/pingcap/tiup/pull/733/commits/3ddcf04d4ad043213bc46ad80855398018a92701

lucklove:
LGTM

Maybe it's better to add related issue into a milestone before we starting work on that... Otherwise the workload may be out of control.

Issue #739, opened 2020-09-01
Use the SIGHUB signal to reload prometheus configuration instead of restarting


lonng:
## Feature Request

Currently, we restart the Prometheus process to reload the latest configuration when reloading. The Prometheus support dynamic configuration reloading, see: https://www.robustperception.io/reloading-prometheus-configuration`.

It's better to use the SIGHUB signal to reload Prometheus configuration instead of restarting.

I think we can implement it via adding a new task implementation, e.g: `PromethuesReload` and execute the `systemctl kill -s SIGWHATEVER prometheus-9100.service` command when the task execution.

PR #740, opened 2020-09-01
Reload prometheus configuration dynamically

lonng:

Close: #739

Currently, we restart the Prometheus process to reload the latest configuration when reloading. The Prometheus support dynamic configuration reloading, see: https://www.robustperception.io/reloading-prometheus-configuration.

It's better to use the SIGHUB signal to reload Prometheus configuration instead of restarting.

### Check List 

Tests 

 - Unit test
 - Integration test

```
level=info ts=2020-09-01T12:37:01.595538035Z caller=main.go:724 msg="Loading configuration file" filename=/home/simple/deploy/prometheus-9090/conf/prometheus.yml
level=info ts=2020-09-01T12:37:01.613266736Z caller=main.go:751 msg="Completed loading of configuration file" filename=/home/simple/deploy/prometheus-9090/conf/prometheus.yml
level=info ts=2020-09-01T12:37:01.613296336Z caller=main.go:609 msg="Server is ready to receive web requests."
level=info ts=2020-09-01T12:37:39.925046179Z caller=main.go:724 msg="Loading configuration file" filename=/home/simple/deploy/prometheus-9090/conf/prometheus.yml
level=info ts=2020-09-01T12:37:39.944044587Z caller=main.go:751 msg="Completed loading of configuration file" filename=/home/simple/deploy/prometheus-9090/conf/prometheus.yml
level=info ts=2020-09-01T12:56:33.660782089Z caller=main.go:724 msg="Loading configuration file" filename=/home/simple/deploy/prometheus-9090/conf/prometheus.yml
level=info ts=2020-09-01T12:56:33.74352596Z caller=main.go:751 msg="Completed loading of configuration file" filename=/home/simple/deploy/prometheus-9090/conf/prometheus.yml
```

Release notes:

```release-note
Reload prometheus configuration dynamically when scale-out/scale-in
```

Support since: v1.1.2


pkg/cluster/task/systemd.go
```diff
@@ -15,6 +15,7 @@ package task
 
 import (
 	"fmt"
+	"github.com/pingcap/errors"
```
lonng, line 19:
```suggestion

	"github.com/pingcap/errors"
```

pkg/cluster/spec/spec.go
```diff
@@ -129,6 +129,8 @@ type Topology interface {
 	// count how many time a path is used by instances in cluster
 	CountDir(host string, dir string) int
 
+	Component(name string) Component
```
july2993, line 132: There's no need to add this to the interface. better keep interface small.
This can be:
```
func  Component(s Topology, name string) Component {
	for _, com := range s.ComponentsByStartOrder() {
		if com.Name() == name {
			return com
		}
	}
	return nil
}
```

pkg/cluster/task/systemd.go
```diff
@@ -24,6 +25,7 @@ type SystemCtl struct {
 	host   string
 	unit   string
 	action string
+	reload bool
```
AstroProfundis, line 28: just set `action` to `reload`

templates/systemd/system.service.tpl
```diff
@@ -21,6 +21,9 @@ LimitSTACK=10485760
 
 User={{.User}}
 ExecStart={{.DeployDir}}/scripts/run_{{.ServiceName}}.sh
+{{- if eq .ServiceName "prometheus"}}
```
july2993, line 24: surprised to find out that eq is a prefix match

july2993 (commented):
rest LGTM

pkg/cluster/task/systemd.go
lonng: It means `daemon-reload`.

pkg/cluster/spec/spec.go
```diff
@@ -510,6 +512,16 @@ func (s *Specification) ComponentsByUpdateOrder() (comps []Component) {
 	return
 }
 
+// Component implements the Topology interface
+func (s *Specification) Component(name string) Component {
```
lucklove, line 516: This repeat https://github.com/pingcap/tiup/pull/740/files#diff-196f616bff16839a1b07c563e558cb00R309

pkg/cluster/task/builder.go
```diff
@@ -281,11 +281,12 @@ func (b *Builder) Shell(host, command string, sudo bool) *Builder {
 }
 
 // SystemCtl run systemctl on host
-func (b *Builder) SystemCtl(host, unit, action string) *Builder {
+func (b *Builder) SystemCtl(host, unit, action string, reload bool) *Builder {
 	b.tasks = append(b.tasks, &SystemCtl{
 		host:   host,
 		unit:   unit,
 		action: action,
+		reload: reload,
```
lucklove, line 289: can we use "reload" as an action? maybe dynreload

pkg/cluster/task/systemd.go
AstroProfundis: Um, then it'd better be named as `reloadDaemon`

pkg/cluster/task/builder.go
```diff
@@ -366,12 +367,18 @@ func (b *Builder) DeploySpark(inst spec.Instance, version, srcPath, deployDir st
 
 // Parallel appends a parallel task to the current task collection
 func (b *Builder) Parallel(tasks ...Task) *Builder {
+	if len(tasks) == 0 {
```
AstroProfundis, line 370: how about
```
if len(tasks) > 0 {
    append
}
return b
```

templates/systemd/system.service.tpl
```diff
@@ -21,6 +21,9 @@ LimitSTACK=10485760
...
+ExecReload=/bin/kill -HUP $MAINPID
```
AstroProfundis, line 25: It's `/usr/bin/kill` on CentOS 7, openSUSE Tumbleweed and ArchLinux, but `/bin/kill` on Debian Stretch and Buster.

AstroProfundis: OK seems they all have `/bin/kill` as an alias

templates/systemd/system.service.tpl
lonng: The `.ServiceName` is not the `prometheus-9100.service` form, the name is misleading you (I think we should blame @lucklove ). The correct field name should be `ComponentName`

july2993: It is `prometheus-9100.service`.

I first thought `eq` must be a full match but found out its prefix match.

lonng: @july2993 No, it's is `prometheus`, it's the component name and not service name.

july2993: 😳 all right.

lucklove (approved):
LGTM

PR #747, opened 2020-09-02
fix(docker): ompatible with macOS's echo

9547:

The `/bin/echo` in macOS defaults to BSD's, and which can't print out a message with the colorful output:

image

_ps_: If you want to use colorful echo output in macOS, you can install `coreutils` and put the binutils  first in PATH

```bash
brew install coreutils
export PATH="/usr/local/opt/coreutils/libexec/gnubin:$PATH"
```

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/docker/up.sh
+++ b/docker/up.sh
@@ -9,15 +9,15 @@ set -o nounset
 # set -o xtrace
 
 ERROR() {
-    /bin/echo -e "\e[101m\e[97m[ERROR]\e[49m\e[39m" "$@"
+    echo -e "\e[101m\e[97m[ERROR]\e[49m\e[39m" "$@"
 }
 
 WARNING() {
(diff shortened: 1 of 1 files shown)


docker/up.sh
```diff
@@ -9,15 +9,15 @@ set -o nounset
 # set -o xtrace
```
lonng, line 10: Can you add the comment here
```
brew install coreutils
export PATH="/usr/local/opt/coreutils/libexec/gnubin:$PATH"
```

july2993 (approved):

PR #749, opened 2020-09-03
Dump dns & port file at data dir

july2993:
Playground dump dsn&port file at wd.
After we changed not to set wd as instance data dir, playground dump
this file at wd but not instance data dir(not same dir like before).
This broke playground subcommand and "tiup client", which search this
files at ~/.tiup/data/../

### Check List 

Tests 

 - Integration test

Release notes:

```release-note
Fix dsn and port file not dumped at instance directory of playground
```


lonng (approved):
LGTM

lucklove (approved):
LGTM

ti-srebot (approved):

PR #758, opened 2020-09-06
cliutil: set cmdLine if registered

### LINKED ISSUE #753: Help message is inconsistent
## Bug Report

1. What did you do?

`tiup cluster check -h` will invoke the subprocess of `tiup-cluster`, and the help message print out as below:

image

missing the `cluster` sub command

```bash
Usage:
    tiup cluster check <topology.yml | cluster-name> [flags]
```

3. What did you see instead?

9547:

fix #753 

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

**Manual test**

```bash
(base) root@rabbit:tiup <fix/cluster-help-inconsistent(v1.1.0-15-gf645c99)>$ ./bin/tiup cluster check -h
Perform preflight checks for the cluster. By default, it checks deploy servers
before a cluster is deployed, the input is the topology.yaml for the cluster.
If '--cluster' is set, it will perform checks for an existing cluster, the input
is the cluster name. Some checks are ignore in this mode, such as port and dir
conflict checks with other clusters

Usage:
  tiup cluster check <topology.yml | cluster-name> [flags]

Flags:
      --apply                  Try to fix failed checks
      --cluster                Check existing cluster, the input is a cluster name.
      --enable-cpu             Enable CPU thread count check
      --enable-disk            Enable disk IO (fio) check
      --enable-mem             Enable memory size check
  -h, --help                   help for check
  -i, --identity_file string   The path of the SSH identity file. If specified, public key authentication will be used. (default "/home/.ssh/id_rsa")
  -p, --password               Use password of target hosts. If specified, password authentication will be used.
  -u, --user string            The user name to login via SSH. The user must has root (or sudo) privilege. (default "root")

Global Flags:
      --native-ssh         Use the native SSH client installed on local system instead of the build-in one (experimental).
      --wait-timeout int   Timeout in seconds to wait for an operation to complete, ignored for operations that don't fit. (default 120)
```

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/components/cluster/main.go
+++ b/components/cluster/main.go
@@ -13,8 +13,12 @@
 
 package main
 
-import "github.com/pingcap/tiup/components/cluster/command"
+import (
+	"github.com/pingcap/tiup/components/cluster/command"
+	"github.com/pingcap/tiup/pkg/cliutil"
+)
 
 func main() {
+	cliutil.RegisterArg0("tiup cluster")
 	command.Execute()
 }
--- a/pkg/cliutil/cliutil.go
+++ b/pkg/cliutil/cliutil.go
@@ -124,7 +124,7 @@ func SuggestionFromFormat(format string, a ...interface{}) (errorx.Property, str
 // BeautifyCobraUsageAndHelp beautifies cobra usages and help.
 func BeautifyCobraUsageAndHelp(rootCmd *cobra.Command) {
 	s := `Usage:{{if .Runnable}}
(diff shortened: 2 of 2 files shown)


9547:
@july2993 PTAL

july2993 (approved):

Issue #761, opened 2020-09-07
PD returns 0 leader count for tikv store after restart


lucklove:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?

Reload the cluster
2. What did you expect to see?

The leader should be transfered before restart TiKV

3. What did you see instead?

TiUP think there is no leader on a specified tikv and restart it directly. And the QPS drop to 0.

4. What version of TiUP are you using (`tiup --version`)?

v1.1.1

PR #762, opened 2020-09-07
Fix wrong tikv store leader count

lucklove:
Fix https://github.com/pingcap/tiup/issues/761

After pd restart, the region leader information will be lost and tiup can't evict tikv leaders correctly.

Use kv metrics api to count leader instead of read it from PDs

### Check List 

Tests 

Both tls and non-tls works:
屏幕快照 2020-09-07 下午7 31 18

Related changes

Release notes:

```release-note
NONE
```


pkg/cluster/api/pdapi.go
```diff
@@ -345,13 +346,23 @@ func (pc *PDClient) EvictStoreLeader(host string, retryOpt *utils.RetryOption) e
 		break
 	}
 
-	if latestStore == nil || latestStore.Status.LeaderCount == 0 {
+	if latestStore == nil {
+		return nil
+	}
+
+	// XXX: the status address in store will be something like 0.0.0.0:20180
```
AstroProfundis, line 353: What is `XXX`?

pkg/cluster/api/pdapi.go
```diff
@@ -345,13 +346,23 @@ func (pc *PDClient) EvictStoreLeader(host string, retryOpt *utils.RetryOption) e
...
+	statusAddress := strings.Split(latestStore.Store.Address, ":")
+	statusAddress[1] = strings.Split(latestStore.Store.StatusAddress, ":")[1]
```
AstroProfundis, line 355: Why not use `StatusAddress` directly?

pkg/cluster/api/pdapi.go
```diff
@@ -345,13 +346,23 @@ func (pc *PDClient) EvictStoreLeader(host string, retryOpt *utils.RetryOption) e
...
+	var leaderCount int
+	if leaderCount, err = countLeader(strings.Join(statusAddress, ":")); err != nil {
```
AstroProfundis, line 357: If there's no default implementation of `countLeader()`, I think we don't have to pass it from argument.

pkg/cluster/spec/tikv.go
```diff
@@ -330,3 +338,59 @@ func addr(ins Instance) string {
 	}
 	return ins.GetHost() + ":" + strconv.Itoa(ins.GetPort())
 }
+
+func genLeaderCounter(tlsCfg *tls.Config) func(string) (int, error) {
```
AstroProfundis, line 342: ditto, why not put this in `pdapi`?

pkg/cluster/api/pdapi.go
lucklove: Because StatusAddress is something like 0.0.0.0:20180, whose ip address is not correct

pkg/cluster/api/pdapi.go
lucklove: i pass it from argument is because that it's the tikv's logic and it's not good to access tikv in pdapi.go

pkg/cluster/api/pdapi.go
lucklove: https://www.snellman.net/blog/archive/2017-04-17-xxx-fixme/

> The token XXX is frequently used in source code comments as a way of marking some code as needing attention.

AstroProfundis (approved):
LGTM

pkg/cluster/spec/tikv.go
```diff
@@ -330,3 +338,78 @@ func addr(ins Instance) string {
...
+					continue
+				} else if m.Labels["type"] != labelNameLeaderCount {
+					continue
+				} else {
+					return strconv.Atoi(m.Value)
+				}
+			}
+		}
+
+		return 0, errors.AddStack(fmt.Errorf("metric %s{type=\"%s\"} not found", metricNameRegionCount, labelNameLeaderCount))
```
lonng, line 398:
```suggestion
		return 0, errors.Errorf("metric %s{type=\"%s\"} not found", metricNameRegionCount, labelNameLeaderCount)
```

lonng: `errors.Errorf` will add stack information to the error.

pkg/cluster/spec/tikv.go
```diff
@@ -330,3 +338,78 @@ func addr(ins Instance) string {
...
+				log.Errorf("failed counting leader on %s (status addr %s), %v", id, addr, err)
+			}
+		}()
+
+		fms := []*prom2json.Family{}
+		for mf := range mfChan {
+			fm := prom2json.NewFamily(mf)
+			fms = append(fms, fm)
+		}
+		for _, fm := range fms {
```
lonng, line 383: Why not merge the two iterations to avoid use a slice to save the `*prom2json.Family`?

pkg/cluster/spec/tikv.go
```diff
@@ -330,3 +338,78 @@ func addr(ins Instance) string {
...
+					return strconv.Atoi(m.Value)
+				}
+			}
+		}
+
+		return 0, errors.AddStack(fmt.Errorf("metric %s{type=\"%s\"} not found", metricNameRegionCount, labelNameLeaderCount))
+	}
+}
+
+func makeTransport(tlsCfg *tls.Config) (*http.Transport, error) {
```
lonng, line 402: Why this function signature is `(tlsCfg *tls.Config) (*http.Transport, error)`? Seems `(tlsCfg *tls.Config) *http.Transport` enough.

pkg/cluster/spec/tikv.go
```diff
@@ -330,3 +338,78 @@ func addr(ins Instance) string {
...
+				continue
+			}
+			for _, m := range fm.Metrics {
+				if m, ok := m.(prom2json.Metric); !ok {
+					continue
+				} else if m.Labels["type"] != labelNameLeaderCount {
+					continue
+				} else {
+					return strconv.Atoi(m.Value)
+				}
```
lonng, line 394: This part can be simplified:
```go
if m, ok := m.(prom2json.Metric); !ok && m.Labels["type"] == labelNameLeaderCount {
    return strconv.Atoi(m.Value)
}
```

pkg/cluster/spec/tikv.go
lucklove: I'm afraid that there may be goroutine leak. eg: the loop found target metric and returns but the sender has not finish all metrics

pkg/cluster/spec/tikv.go
lucklove: That's cool

pkg/cluster/spec/tikv.go
lucklove:
```
if m, ok := m.(prom2json.Metric); ok && m.Labels["type"] == labelNameLeaderCount {
    return strconv.Atoi(m.Value)
}
```

lucklove: Nice

pkg/cluster/spec/tikv.go
lonng: I think we can resolve it via:
```go
defer close(mfChan)
for {
}
```

lonng (approved):
LGTM

ti-srebot (approved):

PR #768, opened 2020-09-09
Fix the issue that tiflash's data is not clean up when tombstone

july2993:
Fix the issue that tiflash's data is not clean up when tombstone

when tombstone
before this pr:
Screen Shot 2020-09-09 at 4 12 21 PM
after this pr:
Screen Shot 2020-09-09 at 5 24 31 PM

We use ServicePort in DestroyClusterTombstone to construct id and filter by id,
but the id of instances of the cluster is construct using Port(TCPPort for
	tiflash),
so it will not get the instance and clean its data in DestroyClusterTombstone.

Note the GetMainPort() is TCPPort now, change the code to both
using GetMainPort() to avoid inconsistent.

### Check List 

Tests 

deploy -> scale-in -> check work as expect

Release notes:

```release-note
Fix the issue that tiflash's data is not clean up when tombstone
```

### DIFF
--- a/pkg/cluster/operation/action.go
+++ b/pkg/cluster/operation/action.go
@@ -261,6 +261,7 @@ func DestroyClusterTombstone(
 		}
 
 		instances := (&spec.TiFlashComponent{Specification: cluster}).Instances()
+		id = s.Host + ":" + strconv.Itoa(s.GetMainPort())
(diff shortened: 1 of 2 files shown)


lonng (approved):
LGTM

PR #773, opened 2020-09-10
Refine log if failed to scp

july2993:
Address problem:
1. No stack trace if Scp failed, and many caller just return err
   directly no any more info.
2. Distinct log level for failed executed command, so we can quickly
search the error ones.

### Check List 

Tests 

chown root:root run_tidb.sh at the target host then reload cluster:

Screen Shot 2020-09-10 at 7 14 00 PM

Related changes

Release notes:

```release-note
NONE
```


pkg/cluster/executor/ssh.go
```diff
@@ -176,7 +181,11 @@ func (e *EasySSHExecutor) Execute(cmd string, sudo bool, timeout ...time.Duratio
 // file from remote to local.
 func (e *EasySSHExecutor) Transfer(src string, dst string, download bool) error {
 	if !download {
-		return e.Config.Scp(src, dst)
+		err := e.Config.Scp(src, dst)
+		if err != nil {
+			return errors.Annotatef(err, "failed to scp %s to (%s@%s):%s", src, e.Config.User, e.Config.Server, dst)
```
lonng, line 186:
```suggestion
			return errors.Annotatef(err, "failed to scp %s to %s@%s:%s", src, e.Config.User, e.Config.Server, dst)
```
We can reuse ` %s@%s:%s` as out CLI arguments if we want to reproduce manually.

lonng (approved):
Rest LGTM

PR #778, opened 2020-09-13
Fix dependences version binding when cloning mirror

lonng:

Currently, if the versions of the non-core components not specified, the latest version will be chosen. The behavior will abort the `tiup cluster deploy` because it binds version to a specific version to `tispark/spark`.

This PR fixes dependences version binding when cloning mirrors.

Release notes:

```release-note
Fix dependences version binding when cloning mirror
```


lucklove:
Close this since https://github.com/pingcap/tiup/pull/779

PR #779, opened 2020-09-13
Fix dependences version binding when cloning mirror

lucklove:

Currently, if the versions of the non-core components not specified, the latest version will be chosen. The behavior will abort the `tiup cluster deploy` because it binds version to a specific version to `tispark/spark`.

This PR fixes dependences version binding when cloning mirrors.

Release notes:

```release-note
Fix dependences version binding when cloning mirror
```


lonng:
Can we use the latest version of tispark by default?

pkg/cluster/spec/bindversion.go
```diff
@@ -31,8 +31,6 @@ func TiDBComponentVersion(comp, version string) string {
 		return "v0.7.0"
 	case ComponentCheckCollector:
 		return "v0.3.1"
-	case ComponentTiSpark:
-		return "v2.3.1"
 	case ComponentSpark:
 		return "v2.4.3"
```
lonng, line 35: It's better to use the latest version of spark.

pkg/cluster/prepare.go
```diff
@@ -26,20 +26,23 @@ type InstanceIter interface {
...
 	instanceIter.IterInstance(func(inst spec.Instance) {
 		key := fmt.Sprintf("%s-%s-%s", inst.ComponentName(), inst.OS(), inst.Arch())
 		if _, found := uniqueTaskList[key]; !found {
 			uniqueTaskList[key] = struct{}{}
 
-			// download spark as dependency of tispark
-			if inst.ComponentName() == spec.ComponentTiSpark {
+			// we don't set version for tispark, so the lastest tispark will be used
+			var version string
+			if inst.ComponentName() != spec.ComponentTiSpark {
```
lonng, line 39: Please use `inst.ComponentName() == spec.ComponentTiSpark` instead of `!=`.

pkg/repository/v1_repository.go
```diff
@@ -40,11 +42,17 @@ import (
...
 var errUnknownComponent = stderrors.New("unknown component")
 
+const (
+	componentManifestCacheKey = "component-manifest"
+)
+
 // V1Repository represents a remote repository viewed with the v1 manifest design.
 type V1Repository struct {
 	Options
+	sync.Mutex
```
lonng, line 52: Do we need to introduce a `Mutex` for `V1Repository`?

lonng (approved):
LGTM

Issue #632, opened 2020-07-27
Add `tiup env` subcommand to show all environment variables of tiup used.


lonng:
## Feature Request

Add `tiup env` subcommand to show all environment variables of TiUP used. e.g:
```
$ tiup env
TIUP_XXX=value
...
```

PR #788, opened 2020-09-17
Support tiup env command

lucklove:
Fix https://github.com/pingcap/tiup/issues/632

Fix https://github.com/pingcap/tiup/issues/632

Add the env subcommand

### Check List 

Tests 

 - Integration test

Related changes

Release notes:

```release-note
Support tiup env sub command
```

### DIFF
--- a/cmd/env.go
+++ b/cmd/env.go
@@ -1,0 +1,53 @@
+// Copyright 2020 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cmd
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/pingcap/tiup/pkg/localdata"
+	"github.com/spf13/cobra"
+)
+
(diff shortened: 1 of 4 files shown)


cmd/env.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"fmt"
+	"os"
+
+	"github.com/pingcap/tiup/pkg/localdata"
+	"github.com/spf13/cobra"
+)
+
+var envList = []string{
+	localdata.EnvNameHome,
```
lonng, line 25: Prefer to add a comment in the `localdata` package and link here.

lonng (approved):
Rest LGTM

cmd/env.go
```diff
@@ -0,0 +1,53 @@
...
+			return nil
+		},
+	}
+
+	return cmd
+}
+
+func showEnvList(names ...string) {
+	for _, name := range names {
+		fmt.Printf("%s=\"%s\"\n", name, os.Getenv(name))
```
lonng, line 51: Prefer to show only value if the user specifies environment variables explicitly.
```shell
➜  devel go env GOPATH
/c/devel/gopath
➜  devel go env GOPATH GOROOT
/c/devel/gopath
/usr/local/go
```

lucklove: Yes, I have noticed the golang's behavior, but I don't know what's the benefit of this. Can you explain this?

lonng: I think to output the value of an envvar instead of `key=value` may help get the value in the script, and the script doesn't need to split the string.

lucklove: OK

lonng (approved):
LGTM

Issue #565, opened 2020-07-02
nightly components re-download always even they are just downloaded when using tiup-cluster


baurine:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?

Deploy a cluster by `tiup cluster` with command `tiup cluster deploy singleHost nightly topology.yaml -i ../_shared/vagrant_key -y --user vagrant` but failed, then I deployed it again with the same command immediately.

2. What did you expect to see?

It should skip the download process.

3. What did you see instead?

Most of the components (except alertmanager / node_exporter / blackbox_exporter) were downloaded again and cost some time to wait.

![企业微信20200702114505](https://user-images.githubusercontent.com/1284531/86317825-ffa45c80-bc62-11ea-9c5d-549fd000aca8.png)

4. What version of TiUP are you using (`tiup --version`)?

```
> tiup --version
v1.0.7 tiup
Go Version: go1.13
Git Branch: v1.0
GitHash: 7f51794
```

PR #793, opened 2020-09-17
Skip download nightly package when it exists

lucklove:
Fix https://github.com/pingcap/tiup/issues/565

Fix https://github.com/pingcap/tiup/issues/565

Stop ignoring nightly when check if we should download. This is ok because the `VerifyComponent` step will remove old nightly version.

### Check List 

Tests 

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/pkg/cluster/operation/download.go
+++ b/pkg/cluster/operation/download.go
@@ -22,7 +22,6 @@ import (
 	"github.com/pingcap/errors"
 	"github.com/pingcap/tiup/pkg/cluster/spec"
 	"github.com/pingcap/tiup/pkg/utils"
-	tiupver "github.com/pingcap/tiup/pkg/version"
 )
 
 // Download the specific version of a component from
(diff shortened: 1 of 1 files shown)


july2993 (commented):
I think you can not skip, the nightly will be updated.
need to check the checksum for skip download

lonng:
@lucklove Yes, as @july2993 said, you cannot skip the downloading for the nightly version even though the tarball exists.

lucklove:
@july2993 @lonng Hey guys, pay attention on [this line](https://github.com/pingcap/tiup/blob/528dc6126f88322962e41cf043a2f96940a7368f/pkg/cluster/operation/download.go#L51), it will check if the tarball have correct checksum, if not, the tarball will be deleted. So if the nightly has been updated, the old one will be deleted and the condition `utils.IsNotExist(srcPath)` will be true and the newest nightly will be downloaded again.

lonng (approved):
LGTM

PR #795, opened 2020-09-18
Fix panic if the query key not exist

july2993:
Fix panic if the query key not exist

check and return error

### Check List 

Tests 

Release notes:

```release-note
NONE
```


lonng (approved):
LGTM

PR #799, opened 2020-09-21
Use alternative privilege escalation command to initialize the environment

darkelf21cn:
The `su` command is used in the `EnvInit` structure to initialize the tidb service account. However, the command is not always allowed in all the environments. For example, in our Linux environment, `sudo` is allowed but `su` and `runuser` are prohibited. So it's better to determine what is supported first then use the supported one to initialize the environment. 

The `PrivilegeEscalationMethod` is added to the `EnvInit` structure which defines all the available privilege escalation methods. When init_env starts, it will first try all the methods until a supported method is found. Then it will use that method to finish the initialization. An error will be raised if there is no privilege escalation method available.

### Check List 

Tests 
I'm happy to write unit tests but I didn't write one because the command need to interact with operation system. So I just wrote some code to print the command and run it manually on my test lab. Please advice if you have better suggestions.
```
package main

import (
	"fmt"
)

// PrivilegeEscalationMethod defiles availabile linux privilege escalation commands
type PrivilegeEscalationMethod int

const (
	// Invalid privilege escalation mathod as a default
	Invalid PrivilegeEscalationMethod = 0
	// Sudo command
	Sudo PrivilegeEscalationMethod = 1
	//Su command
	Su PrivilegeEscalationMethod = 2
	//Runuser command
	Runuser PrivilegeEscalationMethod = 3
)

type EnvInit struct {
	host           string
	deployUser     string
	userGroup      string
	skipCreateUser bool
}

func main() {
	e := &EnvInit{
		host:           "127.0.0.1",
		deployUser:     "tidb",
		userGroup:      "tidb_group",
		skipCreateUser: false,
	}

	// some of the privilege escalation commands might be prohibited by the linux administrator
	// try different methods and choose the one that works (su, runuser, sudo)
	pems := map[PrivilegeEscalationMethod]string{
		Su:      fmt.Sprintf(`su - %s -c 'echo 0'`, e.deployUser),
		Runuser: fmt.Sprintf(`runuser -l %s -c 'echo 0'`, e.deployUser),
		Sudo:    `sudo echo 0`,
	}
	for _, cmd := range pems {
		fmt.Println(cmd)
	}

	fmt.Printf(`su - %s -c 'test -d ~/.ssh || mkdir -p ~/.ssh && chmod 700 ~/.ssh'`+"\n", e.deployUser)
	fmt.Printf(`runuser -l %s -c 'test -d ~/.ssh || mkdir -p ~/.ssh && chmod 700 ~/.ssh'`+"\n", e.deployUser)
	fmt.Printf(`sudo test -d ~/.ssh || sudo mkdir -p ~/.ssh && sudo chmod 700 ~/.ssh && sudo chown %s:%s ~/.ssh`+"\n", e.deployUser, e.userGroup)

	sshAuthorizedKeys := "authorized_keys"
	pk := "AAAAAAA"
	fmt.Printf(`su - %[1]s -c 'grep $(echo %[2]s) %[3]s || echo %[2]s >> %[3]s && chmod 600 %[3]s'`+"\n",
		e.deployUser, pk, sshAuthorizedKeys)
	fmt.Printf(`runuser -l %[1]s -c 'grep $(echo %[2]s) %[3]s || echo %[2]s >> %[3]s && chmod 600 %[3]s'`+"\n",
		e.deployUser, pk, sshAuthorizedKeys)
	fmt.Printf(`sudo grep $(echo %[1]s) %[2]s || sudo echo %[1]s >> %[2]s && chmod 600 %[2]s && sudo chown %[3]s:%[4]s %[2]s`+"\n",
		pk, sshAuthorizedKeys, e.deployUser, e.userGroup)
}

```

Code changes

- Changes within the function

Side effects

 - Increased code complexity

Related changes

None

Release notes:

```release-note
NONE
```

### DIFF
--- a/pkg/cluster/task/env_init.go
+++ b/pkg/cluster/task/env_init.go
@@ -23,6 +23,20 @@ import (
 	"github.com/pingcap/tiup/pkg/cluster/module"
 )
 
+// PrivilegeEscalationMethod defiles availabile linux privilege escalation commands
+type PrivilegeEscalationMethod int
+
+const (
+	// Invalid privilege escalation mathod as a default
+	Invalid PrivilegeEscalationMethod = 0
+	// Sudo command
+	Sudo PrivilegeEscalationMethod = 1
+	//Su command
+	Su PrivilegeEscalationMethod = 2
+	//Runuser command
+	Runuser PrivilegeEscalationMethod = 3
+)
+
 var (
 	errNSEnvInit               = errNS.NewSubNamespace("env_init")
 	errEnvInitSubCommandFailed = errNSEnvInit.NewType("sub_command_failed")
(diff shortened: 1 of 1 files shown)


AstroProfundis:
We have many more locations that uses `sudo` to get super user privilege, I think we should make this approach more generic. Maybe in executor?

darkelf21cn:
> We have many more locations that uses `sudo` to get super user privilege, I think we should make this approach more generic. Maybe in executor?

Thank you Allen for reviewing this. I had a quick chat with Long Heng yesterday and he suggested the same thing. There is another issue here that actually we don't need `sudo` privilege for the tidb service account at all. I noticed that TiUP uses the user account to initialize the TiDB service account and grant the sudo access to it. Then it switches to the service account to continue the deployment. IMHO, it is not necessary because the user account has the full power to do anything on the server. TiUP can just use it to do all the installation stuff then change the file ownership to the service account. In that case, the sudo privilege is not required by the service account at all so the system will be more secure. When we achieve that, this PR will be no longer needed. Long stated he will plan for the change so please feel free to close this one. Cheers~

AstroProfundis:
Yes the only thing we need from the initial root account is to create the service account (`tidb` by default) for the deployment, so it's technically possible for us to use that account following on (if it has sudo privilege, or not doing any super-power actions anymore). I believe @lucklove is working on a validation of the idea.

AstroProfundis:
Hi, is it possible to implement this in SSH executor rather than only for the init process?

darkelf21cn:
> Hi, is it possible to make the option global (like `--native-ssh`) and implement this in SSH executor rather than only for the `env_init` process?

Sure, but I'm quite busy recently. Will take a look at this later.

Issue #787, opened 2020-09-17
Add host label at deploy stage by default


Win-Man:
## Feature Request

**Is your feature request related to a problem? Please describe:**

I deployed a cluster which had 2 tikv instances on one server and didn't add the label config. Then one of the servers crashed. It result some regions can't access and may be some data lost.

**Describe the feature you'd like:**

I suggest tiup could add host label config when deploy a cluster. It's better to add an options like `--no-lables` to decide not to add host label config. `--no-lables` should be false by default.

But there are still some scenarios need to be considered:
* If user config the labels in `topology.yaml`, tiup should use the user's config.
* If user upgrade from previous version, It should add lables or not. 

**Describe alternatives you've considered:**

Alert a warning message if users deploy a cluster which has more than 2 instances on one server and don't add lables config.

**Teachability, Documentation, Adoption, Migration Strategy:**

PR #800, opened 2020-09-21
Support tikv label check

lucklove:
Check if tikv's label is set correctly: fix https://github.com/pingcap/tiup/issues/787

- [x] apply check on deploy
- [x] apply check on scale-out
- [x] apply check on display
- [x] add test

### Check List 

Tests 

 - Unit test
- Manual test

##### If both TiKV labels and PD labels not set:
屏幕快照 2020-09-22 下午4 57 31

##### If TiVK labels are set but PD's not:
屏幕快照 2020-09-22 下午4 59 01

##### If both TiKV labels and PD labels not set:
屏幕快照 2020-09-22 下午6 42 55

##### If TiVK labels are set but PD's not:
In this case, TiKV will refuse to start
```
[2020/09/22 18:30:07.660 +08:00] [FATAL] [server.rs:576] ["failed to start node: Grpc(RpcFailure(RpcStatus { status: 2-UNKNOWN, details: Some(\"key matching the label was not found in the PD, store label key: host \") }))"]
```

#### ScaleOut
The result is the same with deploy

Side effects
 - Increased code complexity

Related changes

Release notes:

```release-note
Automaticlly check if TiKV's label is set
```


pkg/cluster/manager.go
```diff
@@ -529,6 +531,10 @@ func (m *Manager) Display(clusterName string, opt operator.Options) error {
 	filterRoles := set.NewStringSet(opt.Roles...)
 	filterNodes := set.NewStringSet(opt.Nodes...)
 	pdList := topo.BaseTopo().MasterList
+	tlsCfg, err := topo.TLSConfig(m.specManager.Path(clusterName, spec.TLSCertKeyDir))
+	if err != nil {
+		return perrs.AddStack(err)
```
lonng, line 536: The `TLSConfig` has contained stack information.

pkg/cluster/manager.go
```diff
@@ -1057,6 +1070,15 @@ func (m *Manager) Deploy(
 		base.GlobalOptions.SSHType = sshType
 	}
 
+	if !opt.NoLabels {
+		// Check if TiKV's label set correctly
+		lbs := topo.(*spec.Specification).LocationLabels()
+		kvs := topo.(*spec.Specification).TiKVServers
+		if err := spec.CheckTiKVLocationLabels(lbs, kvs); err != nil {
+			return perrs.Errorf("check TiKV label failed, please fix that before continue:\n%s", err)
```
lonng, line 1078: Prefer to attach the label information to the error.

pkg/cluster/spec/spec.go
```diff
@@ -210,6 +210,32 @@ func (s *Specification) BaseTopo() *BaseTopo {
 	}
 }
 
+// LocationLabels returns replication.location-labels in PD config
+func (s *Specification) LocationLabels() []string {
+	lbs := []string{}
+
+	if pdReplica := s.ServerConfigs.PD["replication"]; pdReplica != nil {
```
lonng, line 217: It's better to `merge(s.serverconfigs, pd)`. We allow users to config the `replication.location-labels` at the instance `config`. e.g:
```yaml
pd_servers:
- host: 1.2.3.4
  config:
    replication.location-labels: ...
- host: 1.2.3.5
  config:
    replication.location-labels: ...
````

File: pkg/cluster/spec/tikv.go [hunk 4]
```diff
@@ -120,6 +120,32 @@ func (s TiKVSpec) IsImported() bool {
 	return s.Imported
 }
 
+// Labels returns the labels of TiKV
+func (s TiKVSpec) Labels() map[string]string {
+	lbs := make(map[string]string)
+
+	if serverCfg := s.Config["server"]; serverCfg != nil {
```
lonng, line 127: ditto.

pkg/cluster/spec/validate.go
```diff
@@ -307,6 +307,64 @@ Please change to use another port or another host.
...
+	}
+	return str
+}
+
+// CheckTiKVLocationLabels will check if tikv missing label or have wrong label
+func CheckTiKVLocationLabels(pdLocLabels []string, kvs []TiKVSpec) error {
+	lerr := &TiKVLabelError{
+		TiKVInstances: make(map[string][]error),
+	}
+	lbs := set.NewStringSet(pdLocLabels...)
```
lonng, line 336: It's better to check whether the `pdLocLabels` is empty, so does `kvs`.

pkg/cluster/manager.go
lucklove: the `err` contains label info
image

pkg/cluster/spec/validate.go
lucklove: 1. If the `kvs` labels is empty and two kv deployed to the same host, it will report an error
image
2.If the `kvs` labels is empty but the `pdLocLabels` is, it will report the label is not found in PD config
image

So the empty `pdLocLabels` is not a validate case.

pkg/cluster/spec/spec.go
lucklove: Have forbiden this case

pkg/cluster/api/pdapi.go
```diff
@@ -661,6 +662,21 @@ func (pc *PDClient) GetReplicateConfig() ([]byte, error) {
...
+// GetLocationLabels gets the replication.location-labels config from pd server
+func (pc *PDClient) GetLocationLabels() ([]string, error) {
+	config, err := pc.GetReplicateConfig()
+	if err != nil {
+		return nil, err
+	}
+
+	rc := pdconfig.ReplicationConfig{}
+	if err := json.Unmarshal(config, &rc); err != nil {
+		return nil, errors.Annotate(err, "unmarshal replication config")
```
lonng, line 674: Do we need to attach the raw byte string to the errors? (avoid to ask users fetch data via HTTP API)

pkg/cluster/spec/spec.go
```diff
@@ -210,6 +210,30 @@ func (s *Specification) BaseTopo() *BaseTopo {
...
+				"replication.location-labels can't be defined in instance %s:%d, please move it to the global server_configs field",
+				pd.Host,
+				pd.GetMainPort(),
+			)
+		}
+	}
+
+	if repLbs := GetValueFromPath(s.ServerConfigs.PD, "replication.location-labels"); repLbs != nil {
+		for _, l := range repLbs.([]interface{}) {
+			lbs = append(lbs, l.(string))
```
lonng, line 230: I think `l.(string)` may cause paniked.

pkg/cluster/spec/validate.go
```diff
@@ -307,6 +307,64 @@ Please change to use another port or another host.
...
+	for _, kv := range kvs {
+		hosts[kv.Host] = hosts[kv.Host] + 1
+	}
+	for _, kv := range kvs {
+		id := fmt.Sprintf("%s:%d", kv.Host, kv.GetMainPort())
+		ls := kv.Labels()
+		if len(ls) == 0 && hosts[kv.Host] > 1 {
+			lerr.TiKVInstances[id] = append(
+				lerr.TiKVInstances[id],
+				errors.New("location label missing"),
```
lonng, line 348:
```suggestion
				errors.New("two TiKV instances are deployed at the same host but location label missing"),
```

pkg/cluster/spec/validate.go
```diff
@@ -307,6 +307,64 @@ Please change to use another port or another host.
 	return nil
 }
 
+// TiKVLabelError indicates that some TiKV servers don't have correct labels
+type TiKVLabelError struct {
+	TiKVInstances map[string][]error
```
lonng, line 312: Prefer to use `slice` to keep the result stable.

lonng (commented):
Rest LGTM

pkg/cluster/spec/validate.go
lucklove: To make a stable output, I think we should sort the error in the `Error` [method](https://github.com/pingcap/tiup/pull/800/files#diff-3ee04046ce3e73cd5ee6020d945d17f4R316) instead of using a slice here. Because the [origin labels data](https://github.com/pingcap/tiup/pull/800/files#diff-ae7f1374cdcf99ec34f0f79ba4d51b08R127) we fetched is a map, so even if we make it a slice, it's still unstable

lonng (approved):
LGTM

PR #806, opened 2020-09-23
cluster: fix custom port handling for drainer on importing

AstroProfundis:
The custom `drainer_port` for host is not read on import, so the value in `group_vars` is used instead.

### Check List 

Tests 

 - Unit test

Code changes

 - Has persistent data change

Side effects

 - None

Related changes

Release notes:

```release-note
Fix per-host custom port for drainer not handled correctly on importing
```


lonng (approved):
LGTM

PR #816, opened 2020-09-25
Fix self update

lucklove:

The code in the master branch will get the `tiup` binary lost after execution `tiup update --self`:
屏幕快照 2020-09-25 下午5 11 42

This is because we treat `tiup` as a normal component and install it to `~/.tiup/components/tiup/{version}/` in [SelfUpdate](https://github.com/pingcap/tiup/blob/master/pkg/environment/env.go#L219) and once it's installed `SelfUpdate` will return nil and the old `tiup` binary will be remove from ~/.tup/bin.

In a word, it should be installed to "~/.tiup/bin" instead of "~/.tiup/components/tiup/{version}/"

Read the `TargetDir` field of `ComponentSpec`, if it's not empty, install component to where it's specified.

### Check List 

Tests 

 - Integration test

Related changes

Release notes:

```release-note
NONE
```


lonng (approved):
lgtm

Issue #562, opened 2020-07-02
Use absolute path of DataDir/DeployDir/LogDir when assign default value


lonng:
## Feature Request

Use absolute path of DataDir/DeployDir/LogDir when assign default value.

PR #822, opened 2020-09-27
Use absolute path of DataDir/DeployDir/LogDir

lucklove:
Fix https://github.com/pingcap/tiup/issues/562

https://github.com/pingcap/tiup/issues/562

Check all `DeployDir`, `DataDir`, `LogDir` fields after `ParseTopologyYaml` on the stage of
- [x] deploy
- [x] scale-out

### Check List 

Tests 

 - Unit test

Related changes

Release notes:

```release-note
NONE
```


pkg/cluster/spec/parse_topology.go
```diff
@@ -63,3 +67,66 @@ Please check the syntax of your topology file {{ColorKeyword}}{{.File}}{{ColorRe
 
 	return nil
 }
+
+// FixRelativeDir fill DeployDir, DataDir and LogDir to absolute path
+func FixRelativeDir(topo Topology) {
```
AstroProfundis, line 72: I'd prefer rename it to `ExtendRelativeDir` as it's not fixing anything wrong.

pkg/cluster/spec/parse_topology.go
```diff
@@ -63,3 +67,66 @@ Please check the syntax of your topology file {{ColorKeyword}}{{.File}}{{ColorRe
...
+	fixRelativePath(deployUser(topo), topo)
+}
+
+func fixRelativePath(user string, topo interface{}) {
```
AstroProfundis, line 76: ditto

AstroProfundis (commented):
We should add some more tests for absolute paths in global and for data dir and log dirs.

AstroProfundis (approved):

PR #857, opened 2020-10-21
Fix: destroy monitor if no more components in host

### LINKED ISSUE #842: When you scale-in PD, node_exporter of the PD node is still present
## Bug Report

1. What did you do?

There is only pd service on 172.16.5.142, after executing tiup cluster scale-in lwt-cluster -N 172.16.5.142:2494, you can still see node_exporter and blackbox_exporter  exist on 172.16.5.142.

yaml:
image

tiup cluster scale-in lwt-cluster -N 172.16.5.142:2494
image

When you deploy without mixing, scale-in needs to clean up node_exporter and blackbox_exporter.

3. What did you see instead?

image
image

[tidb@node5162 ~]$ tiup cluster --version
Starting component `cluster`: /home/tidb/.tiup/components/cluster/v1.2.0/tiup-cluster --version
tiup version v1.2.0 tiup
Go Version: go1.13
Git Branch: release-1.2
GitHash: 1a4fbe7
[tidb@node5162 ~]$ tiup --version
v1.1.2 tiup
Go Version: go1.13
Git Branch: release-1.1
GitHash: 7cb85b9-dirty

9547:
fix #842 

After scale-in component, check whether the host has any more components, if not, stop and destroy the monitor components.

### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test 

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```


9547:
@AstroProfundis PTAL

components/dm/command/scale_in.go
```diff
@@ -100,23 +102,46 @@ func ScaleInDMCluster(
 	}
 
 	// Cannot delete all DM DMMaster servers
-	if len(deletedDiff[dm.ComponentDMMaster]) == len(spec.Masters) {
+	if len(deletedDiff[dm.ComponentDMMaster]) == len(topo.Masters) {
 		return errors.New("cannot delete all dm-master servers")
 	}
 
+	maybeDestroyMonitor := func(instance spec.Instance) error {
```
lucklove, line 109: I think there is no need to destroy monitored instance in dm cluster since dm doesn't support monitored instance at all: https://github.com/pingcap/tiup/blob/master/components/dm/spec/topology_dm.go#L89

pkg/cluster/operation/scale_in.go
```diff
@@ -122,29 +124,68 @@ func ScaleInCluster(
...
+		if err := DestroyComponent(getter, []spec.Instance{instance}, cluster, options); err != nil {
+			if !ignoreErr {
+				return errors.Annotatef(err, "failed to destroy %s", component.Name())
+			}
+			log.Warnf("failed to destroy %s: %v", component.Name(), err)
+		}
+
+		monitoredOptions := cluster.GetMonitoredOptions()
+		instCount[instance.GetHost()]--
+		if instCount[instance.GetHost()] == 0 && cluster.GetMonitoredOptions() != nil {
```
lucklove, line 158: cluster.GetMonitoredOptions() != nil is not needed because it's always true here: https://github.com/pingcap/tiup/blob/master/pkg/cluster/spec/spec.go#L192

pkg/cluster/operation/scale_in.go
```diff
@@ -230,11 +255,8 @@ func ScaleInCluster(
...
 			if !asyncOfflineComps.Exist(instance.ComponentName()) {
-				if err := StopComponent(getter, []spec.Instance{instance}, options.OptTimeout); err != nil {
-					return errors.Annotatef(err, "failed to stop %s", component.Name())
-				}
-				if err := DestroyComponent(getter, []spec.Instance{instance}, cluster, options); err != nil {
-					return errors.Annotatef(err, "failed to destroy %s", component.Name())
+				if err := offlineInstance(component, instance); err != nil {
+					return err
 				}
 			} else {
```
lucklove, line 261: We should do the same thing for asyncOfflineComps too. Example: when you scale-in a tikv node, it will not be remove at once, instead, the PD server will transfer it's data to another tikv node, after that, the user can execute the prune command (https://github.com/pingcap/tiup/blob/master/components/cluster/command/prune.go#L45) to remove that TiKV (stop & destroy). The prune command will call `DestroyClusterTombstone` finally, so I think we should implement the same logic in DestroyClusterTombstone(https://github.com/pingcap/tiup/blob/master/pkg/cluster/operation/action.go#L178)

lucklove (commented):
Thanks for your contributes, 9547, it really awesome!

components/dm/command/scale_in.go
9547: Thanks for the advice, revert this commit

pkg/cluster/operation/scale_in.go
9547: Sorry, this logical is forked from https://github.com/pingcap/tiup/blob/7aa4ea40705925f32e3033854e3bb786e87395e9/pkg/cluster/operation/destroy.go#L75-L79

In which case, the cluster may be tidb or dm, don't think too much, just copied it out.

pkg/cluster/operation/scale_in.go
9547: implemented in daf3e3a

9547:
@lucklove PTAL

pkg/cluster/operation/destroy.go
```diff
@@ -91,6 +90,46 @@ func Destroy(
 	return nil
 }
 
+// StopAndDestroyInstance stop and destroy the instance,
+// if this instance is the host's last one, and the host has monitor deployed,
+// we need to destroy the monitor, either
+func StopAndDestroyInstance(getter ExecutorGetter, component spec.Component, cluster spec.Topology, instance spec.Instance, options Options, destroyMonitor bool) error {
```
lucklove, line 96: I think you can remove the `component` arg and use `instance.ComponentName()` or `instance.ID()` instead.

9547: This arg is reserved, maybe has other usage later

lucklove: Then what about add it back when we really need it?

9547: LGTM

9547:
@lucklove PTAL again

pkg/cluster/operation/destroy.go
```diff
@@ -91,6 +90,46 @@ func Destroy(
 	return nil
 }
 
+// StopAndDestroyInstance stop and destroy the instance,
+// if this instance is the host's last one, and the host has monitor deployed,
+// we need to destroy the monitor, either
+func StopAndDestroyInstance(getter ExecutorGetter, compName string, cluster spec.Topology, instance spec.Instance, options Options, destroyMonitor bool) error {
```
lucklove, line 96: We can fetch the compName from `instance.ComponentName()`, so it can be removed at this time....

pkg/cluster/operation/destroy.go
```diff
@@ -368,3 +407,191 @@ func DestroyComponent(getter ExecutorGetter, instances []spec.Instance, cls spec
...
+	filterID := func(instance []spec.Instance, id string) (res []spec.Instance) {
+		for _, ins := range instance {
+			if ins.ID() == id {
+				res = append(res, ins)
+			}
+		}
+		return
+	}
+
+	maybeDestroyMonitor := func(comp spec.Component, instances []spec.Instance, id string) error {
```
lucklove, line 455: And if we remove the arg `compName` from `StopAndDestroyInstance`, the `comp` arg for `maybeDestroyMonitor` is not needed

pkg/cluster/operation/destroy.go
```diff
@@ -368,3 +407,191 @@ func DestroyComponent(getter ExecutorGetter, instances []spec.Instance, cls spec
...
+		}
+
+		nodes = append(nodes, id)
+		if returNodesOnly {
+			continue
+		}
+
+		instances := (&spec.TiFlashComponent{Specification: cluster}).Instances()
+		id = s.Host + ":" + strconv.Itoa(s.GetMainPort())
+		if err := maybeDestroyMonitor(&spec.TiKVComponent{}, instances, id); err != nil {
```
lucklove, line 524: It should not be `&spec.TiKVComponent{}` and it should be removed

pkg/cluster/operation/destroy.go
```diff
@@ -368,3 +407,191 @@ func DestroyComponent(getter ExecutorGetter, instances []spec.Instance, cls spec
...
+			drainerServers = append(drainerServers, s)
+		}
+
+		nodes = append(nodes, id)
+		if returNodesOnly {
+			continue
+		}
+
+		instances := (&spec.DrainerComponent{Specification: cluster}).Instances()
+		if err := maybeDestroyMonitor(&spec.TiKVComponent{}, instances, id); err != nil {
```
lucklove, line 582: ditto

lucklove (commented):
Rest LGTM

components/dm/command/scale_in.go
```diff
@@ -100,23 +99,17 @@ func ScaleInDMCluster(
...
 					continue
 				}
-				// just try stop and destroy
-				if err := operator.StopComponent(getter, []dm.Instance{instance}, options.OptTimeout); err != nil {
-					log.Warnf("failed to stop %s: %v", component.Name(), err)
-				}
-				if err := operator.DestroyComponent(getter, []dm.Instance{instance}, spec, options); err != nil {
-					log.Warnf("failed to destroy %s: %v", component.Name(), err)
-				}
+				_ = operator.StopAndDestroyInstance(getter, component.Name(), topo, instance, options, false)
```
lonng, line 112: We shouldn't ignore the error, please use `log.Warnf` if this error is not critical.

pkg/cluster/operation/scale_in.go
```diff
@@ -122,66 +124,55 @@ func ScaleInCluster(
...
 
-				// just try stop and destroy
-				if err := StopComponent(getter, []spec.Instance{instance}, options.OptTimeout); err != nil {
-					log.Warnf("failed to stop %s: %v", component.Name(), err)
-				}
-				if err := DestroyComponent(getter, []spec.Instance{instance}, cluster, options); err != nil {
-					log.Warnf("failed to destroy %s: %v", component.Name(), err)
-				}
+				instCount[instance.GetHost()]--
+				_ = StopAndDestroyInstance(getter, compName, cluster, instance, options, instCount[instance.GetHost()] == 0)
```
AstroProfundis, line 154: This error should be handled too

pkg/cluster/operation/destroy.go
9547: Thanks for the advice.

pkg/cluster/operation/destroy.go
```diff
@@ -368,3 +407,191 @@ func DestroyComponent(getter ExecutorGetter, instances []spec.Instance, cls spec
...
+			pumpServers = append(pumpServers, s)
+		}
+
+		nodes = append(nodes, id)
+		if returNodesOnly {
+			continue
+		}
+
+		instances := (&spec.PumpComponent{Specification: cluster}).Instances()
+		if err := maybeDestroyMonitor(&spec.TiKVComponent{}, instances, id); err != nil {
```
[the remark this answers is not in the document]
9547: This arg will be removed from `maybeDestroyMonitor`, and thanks to @lucklove 's advice

9547:
@lucklove Please help me about the test failure, curious about the test failure reason (exit 255), and it runs successful on my local dev env

lucklove:
> @lucklove Please help me about the test failure, curious about the test failure reason (exit 255), and it runs successful on my local dev env

It's because of this:
> fail to wait instance number reach 20, count 21, retry num: 120

lucklove:
It's because there is a TiKV node that is still in offline process:
```
172.19.0.102:20160  tikv            172.19.0.102  20160/20180                      linux/x86_64  Pending Offline
```

tests/tiup-cluster/script/scale_core.sh
```diff
@@ -62,13 +64,38 @@ function scale_core() {
...
-
     # after scalue-out, ensure this instance come back
     tiup-cluster $client exec $name -N $ipprefix.101 --command "grep -q $ipprefix.103:2379 /home/tidb/deploy/tidb-4000/scripts/run_tidb.sh"
 
+    echo "start scale in node-2 with [tidb, tikv]"
+    tiup-cluster $client --yes scale-in $name -N $ipprefix.102:4000
+    wait_instance_num_reach $name $total_sub_one $native_ssh
+    tiup-cluster $client --yes scale-in $name -N $ipprefix.102:20160
+    echo "after sclae in(without prune) tikv, the monitors should not be destroyed"
+    wait_instance_num_reach $name $total_sub_one $native_ssh
```
lucklove, line 80: > after sclae in(without prune) tikv, the monitors should not be destroyed

`waiit_instance_num_reach` will call prune automaticlly, this is not what you expected

9547: but why it was passed in my local dev

9547: I'll dig into later tonight, and thanks for your help 😀

lucklove: It passed in your local dev, I think it may related to the opportunity:
- Only when the tikv's state is tombstone, the prune command cleanup it
- After scale in a tikv node, it will take some time to transfer to tombstone state (seconds to minutes)
So I guess your local host, when the `wait_instance_num_reach` execute, the tikv node is still in `pending offline` state, so it passed, but in CI, the tikv node is `tombstone`, so it was cleanup.

9547: 1. Seems in Github CI, it's hard to predicate when TiKV will finish the transferring process and go to tombstone state, after that `prune` can evict the TiKV. 
2.  have reserved only one TIDB node on the 102 node for testing the issue #842

lucklove (approved):
LGTM

PR #858, opened 2020-10-22
More safe way to cleanup tombstone nodes

lucklove:
With TiUP, if we want to offline a TiKV server, we need do this first:

```sh
tiup cluster scale-in <cluster-name> -N <tikv-node-id>
```

This will send a request to PD servers to tell them that we want to offline this TiKV server, then PD servers will transfer all regions in this TiKV to some other place. After transferring data, the state of this TiKV server will be tombstone. At this point, the data of this server will be deleted (see the yellow line in the picture below).

屏幕快照 2020-10-22 下午12 20 46

 But oncel the user execute the `display` command:

```sh
tiup cluster display <cluster-name>
```

The TiKV server in tombstone state will be destroyed and cleanup at once, this may confuse users and it's unsafe.

DO NOT cleanup tombstone TiKV servers in display command, but add a prune command to do this, and give the user a choice to cancel the action:

屏幕快照 2020-10-22 下午12 23 34

### Check List 

Tests 

 - Integration test

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/components/cluster/command/prune.go
+++ b/components/cluster/command/prune.go
@@ -82,11 +82,13 @@ func destroyTombstoneIfNeed(clusterName string, metadata *spec.ClusterMeta, opt 
 		return nil
 	}
 
-	err = cliutil.PromptForConfirmOrAbortError(
-		color.HiYellowString(fmt.Sprintf("Will destroy these nodes: %v\nDo you confirm this action? [y/N]:", nodes)),
-	)
-	if err != nil {
-		return err
+	if !skipConfirm {
+		err = cliutil.PromptForConfirmOrAbortError(
+			color.HiYellowString(fmt.Sprintf("Will destroy these nodes: %v\nDo you confirm this action? [y/N]:", nodes)),
+		)
+		if err != nil {
+			return err
+		}
 	}
 
 	log.Infof("Start destroy Tombstone nodes: %v ...", nodes)
--- a/components/cluster/command/prune.go
+++ b/components/cluster/command/prune.go
@@ -15,7 +15,6 @@ package command
 
 import (
 	"fmt"
-	"strconv"
 
 	"github.com/fatih/color"
 	perrs "github.com/pingcap/errors"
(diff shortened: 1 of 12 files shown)


components/cluster/command/prune.go
```diff
@@ -0,0 +1,126 @@
...
+	printTable := [][]string{}
+	printTable = append(printTable, []string{
+		"ID", "Role", "Host", "Main Port",
+	})
+
+	for _, inst := range tombs {
+		host, _ := inst.SSH()
+		role := inst.Role()
+		port := inst.GetMainPort()
+		id := fmt.Sprintf("%s:%d", host, port)
```
lonng, line 63: It's better to use `inst.ID()` method.

components/cluster/command/prune.go
```diff
@@ -0,0 +1,126 @@
...
+
+	err = ctx.SetClusterSSH(topo, metadata.User, gOpt.SSHTimeout, gOpt.SSHType, topo.BaseTopo().GlobalOptions.SSHType)
+	if err != nil {
+		return perrs.AddStack(err)
+	}
+
+	nodes, err := operator.DestroyTombstone(ctx, topo, true /* returnNodesOnly */, opt, tlsCfg)
+	if err != nil {
+		return perrs.AddStack(err)
+	}
```
lonng, line 110: I think we should move the `confirm` to here.

lonng (commented):
LGTM

pkg/cluster/manager.go
```diff
@@ -599,14 +599,20 @@ func (m *Manager) Display(clusterName string, opt operator.Options) error {
...
 		if lbs, err := pdClient.GetLocationLabels(); err != nil {
 			color.Yellow("\nWARN: get location labels from pd failed: %v", err)
 		} else if err := spec.CheckTiKVLocationLabels(lbs, pdClient); err != nil {
 			color.Yellow("\nWARN: there is something wrong with TiKV labels, which may cause data losing:\n%v", err)
 		}
+
+		// Check if there is some instance in tombstone state
+		nodes, _ := operator.DestroyTombstone(ctx, t, true /* returnNodesOnly */, opt, tlsCfg)
+		if len(nodes) != 0 {
+			color.Green("There is some nodes in state Tombstone, you can destroy them with the command `tiup cluster prune %s`", clusterName)
```
lonng, line 614:
```suggestion
			color.Green("There are some nodes in state: `Tombstone`\n\tNodes: %+v\n\tYou can destroy them with the command: `tiup cluster prune %s`", nodes, clusterName)
```

AstroProfundis (approved):

ti-srebot (approved):

Issue #814, opened 2020-09-25
Support merge offline packages


lucklove:
## Feature Request

**Is your feature request related to a problem? Please describe:**

At current, if the user use a offline package to deploy his cluster, when he want to upgrade it to another version, or he want to deploy a cluster with a different version. He must switch the mirror to the right offline package via the command `tiup mirror set`.

**Describe the feature you'd like:**

There should be a way to merge two different offline package to a bigger package which contains all the components from both offline packages, so that we don't need to execute `tiup mirror set` to manager different version cluster.

PR #860, opened 2020-10-22
Refactor mirror command

lucklove:
https://github.com/pingcap/tiup/issues/814

- [x] Support publish to local mirror
- [x] Support merge mirror
- [x] Refactor tiup-server
- [x] Support add owner
- [x] Add test

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

Release notes:

```release-note
Support merge offline packages
```


cmd/mirror.go
```diff
@@ -221,80 +118,192 @@ func newMirrorSetCmd() *cobra.Command {
 	return cmd
 }
 
+// the `mirror introduce` sub command
+func newMirrorIntroduceCmd() *cobra.Command {
+	name := ""
+	privPath := ""
+
+	cmd := &cobra.Command{
+		Use:   "introduce <id>",
```
lonng, line 127: How about using` tiup mirror grant/revoke` to add a new owner and remove exists owner.

cmd/mirror.go
```diff
@@ -221,80 +118,192 @@ func newMirrorSetCmd() *cobra.Command {
...
+		Short: "introduce a new owner",
+		Long:  "introduce a new owner to current mirror",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) < 1 {
+				return cmd.Help()
+			}
+
+			id := args[0]
+			if name == "" {
+				name = id
```
lonng, line 137: We assigned a default value for name, do we need to print some message to hint the user? e.g:
```go
fmt.Printf("The --name hasn't specified value, default to %s\n", id).
```

cmd/mirror.go
```diff
@@ -221,80 +118,192 @@ func newMirrorSetCmd() *cobra.Command {
...
+				return err
+			}
+
+			env := environment.GlobalEnv()
+			return env.V1Repository().Mirror().Introduce(id, name, pubKey)
+		},
+	}
+
+	cmd.Flags().StringVarP(&name, "name", "n", "", "the name of the owner, default: id of the owner")
+	cmd.Flags().StringVarP(&privPath, "key", "k", "", "private/public key path of the owner")
```
lonng, line 156: I think we should make the description more clear if only `private key` can be assigned.
`private/public key path of the owner` -> `Specify the private key path of the owner`.

IMO, all descriptions for flags should be started with a upper case.

cmd/mirror.go
```diff
@@ -221,80 +118,192 @@ func newMirrorSetCmd() *cobra.Command {
...
 	desc := ""
 	standalone := false
 	hidden := false
 	yanked := false
 
 	cmd := &cobra.Command{
-		Use:  "modify <component>[:version] [flags]",
-		Long: "modify component attributes (hidden, standalone, yanked)",
+		Use:   "modify <component>[:version] [flags]",
+		Short: "modify published component",
```
lonng, line 171:
```suggestion
		Short: "Modify published component",
```

cmd/mirror.go
```diff
@@ -221,80 +118,192 @@ func newMirrorSetCmd() *cobra.Command {
...
+		Long:  "modify component attributes (hidden, standalone, yanked)",
```
lonng, line 172:
```suggestion
		Long:  "Modify component attributes (hidden, standalone, yanked)",
```

cmd/mirror.go
```diff
@@ -27,18 +28,16 @@ import (
 	"github.com/pingcap/tiup/pkg/environment"
 	"github.com/pingcap/tiup/pkg/localdata"
 	"github.com/pingcap/tiup/pkg/repository"
-	"github.com/pingcap/tiup/pkg/repository/remote"
+	"github.com/pingcap/tiup/pkg/repository/model"
+	ru "github.com/pingcap/tiup/pkg/repository/utils"
+	"github.com/pingcap/tiup/pkg/repository/v0manifest"
```
lonng, line 33: I think we can remove the `v0manifest` in the `v1.4.0`. @AstroProfundis

cmd/mirror.go
```diff
@@ -310,72 +319,60 @@ func newMirrorPublishCmd() *cobra.Command {
...
-			if privPath == "" {
-				privPath = env.Profile().Path(localdata.KeyInfoParentDir, "private.json")
+			hashes, length, err := ru.HashFile(tarpath)
+			if err != nil {
+				return err
 			}
 
-			// Get the private key
-			f, err := os.Open(privPath)
+			tarfile, err := os.Open(args[2])
```
lonng, line 333: Please use the variable declared at L322.

cmd/mirror.go
```diff
@@ -310,72 +319,60 @@ func newMirrorPublishCmd() *cobra.Command {
...
 			if err != nil {
-				return err
+				return errors.Annotate(err, "open tarbal")
```
lonng, line 335:
```suggestion
				return errors.Annotatef(err, "open tarball: %s", tarpath)
```

cmd/mirror.go
lucklove: I don't think remove an owner is possible. The owner doesn't has a 'yank' field so we can just remove it from the manifest file. After that, the existed components owned by that owner can't be verified (since the key is removed with that owner) and there will be signature error when user execute `tiup list` or execute the installed component.

cmd/mirror.go
lucklove: OK

cmd/mirror.go
lucklove: Maybe we should ignore the components missing owner, and report warning when user already install that component.

lucklove: But what if something depends on it? 

IMO, delete owner directly breaks the `yank` mechanism

cmd/mirror.go
AstroProfundis: I agree, https://github.com/pingcap/tiup/issues/906

cmd/mirror.go
```diff
@@ -221,80 +118,193 @@ func newMirrorSetCmd() *cobra.Command {
...
+		Short: "introduce a new owner",
+		Long:  "introduce a new owner to current mirror",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) < 1 {
+				return cmd.Help()
+			}
+
+			id := args[0]
+			if name == "" {
+				fmt.Printf("The --name hasn't specified value, default to %s\n", id)
```
AstroProfundis, line 137:
```suggestion
				fmt.Printf("The --name is not specified, using %s as default\n", id)
```

cmd/mirror.go
```diff
@@ -221,80 +118,193 @@ func newMirrorSetCmd() *cobra.Command {
...
+			comp, ver := environment.ParseCompVersion(component)
 			m, err := env.V1Repository().FetchComponentManifest(comp, true)
 			if err != nil {
 				return err
 			}
 
-			if endpoint == "" {
-				endpoint = environment.Mirror()
+			v1manifest.RenewManifest(m, time.Now())
+			m.Version++
```
AstroProfundis, line 190: Why not bump version in `RenewManifest()`?

pkg/cluster/audit/audit.go
```diff
@@ -81,7 +82,7 @@ func ShowAuditList(dir string) error {
 
 // OutputAuditLog outputs audit log.
 func OutputAuditLog(dir string, data []byte) error {
-	fname := filepath.Join(dir, base52.Encode(time.Now().UnixNano()))
+	fname := filepath.Join(dir, base52.Encode(time.Now().UnixNano()+rand.Int63n(1000)))
```
AstroProfundis, line 85: Better split this to another PR

pkg/repository/merge_mirror.go
```diff
@@ -0,0 +1,295 @@
...
+	name          string
+	componentItem v1manifest.ComponentItem
+	versionItem   v1manifest.VersionItem
+	version       string
+	os            string
+	arch          string
+	desc          string
+}
+
+// returns component that exists in source but not in base
```
AstroProfundis, line 38: "source" and "base" are bad names, they have similar meaning and might be confusing. how about "base" and "input" or "addition"

pkg/repository/merge_mirror.go
```diff
@@ -0,0 +1,295 @@
...
+
+		baseComponent, err := fetchComponentManifestFromMirror(base, name)
+		if err != nil {
+			return nil, err
+		}
+		sourceComponent, err := fetchComponentManifestFromMirror(source, name)
+		if err != nil {
+			return nil, err
+		}
+		if sourceComponent == nil {
```
AstroProfundis, line 67: Is this possible?

pkg/repository/merge_mirror.go
```diff
@@ -0,0 +1,295 @@
...
+		for ver, verinfo := range versions {
+			// Don't merge nightly this time
+			if v0manifest.Version(ver).IsNightly() {
+				continue
+			}
+
+			// this version not exits in base
+			if baseManifest.VersionListWithYanked(plat)[ver].URL == "" {
+				osArch := strings.Split(plat, "/")
+				if len(osArch) != 2 {
```
AstroProfundis, line 92: What's this check for?

pkg/repository/merge_mirror.go
```diff
@@ -0,0 +1,295 @@
...
+	defer r.Close()
+
+	snap := v1manifest.Snapshot{}
+	if _, err := v1manifest.ReadNoVerify(r, &snap); err != nil {
+		return nil, err
+	}
+
+	v := snap.Meta[fmt.Sprintf("/%s.json", component)].Version
+	if v == 0 {
+		return nil, nil
```
AstroProfundis, line 181: I think this is not reasonable to return `nil, nil`

pkg/repository/merge_mirror.go
```diff
@@ -0,0 +1,295 @@
...
+	name, ver, entry, os, arch, desc string,
+	filehash v1manifest.FileHash) *v1manifest.Component {
+	initTime := time.Now()
+
+	// update manifest
+	if m == nil {
+		m = v1manifest.NewComponent(name, desc, initTime)
+	} else {
+		v1manifest.RenewManifest(m, initTime)
+		m.Version++
```
AstroProfundis, line 263: ditto, bumping version is not atomic here

AstroProfundis (commented):
And don't change the file mode of `pkg/cluster/executor/executor.go`

pkg/repository/mirror.go
```diff
@@ -249,6 +326,71 @@ func (l *httpMirror) prepareURL(resource string) string {
...
+
+	if info.Filename() != "" {
+		tarAddr := fmt.Sprintf("%s/api/v1/tarball/%s", l.Source(), sid)
+		resp, err := utils.PostFile(info, tarAddr, "file", info.Filename())
+		if err != nil {
+			return err
+		}
+		defer resp.Body.Close()
+
+		if resp.StatusCode >= 300 {
```
AstroProfundis, line 346: What about 301, 302, 303 and 307?

pkg/repository/mirror.go
```diff
@@ -249,6 +326,71 @@ func (l *httpMirror) prepareURL(resource string) string {
...
+	}
+
+	payload, err := json.Marshal(manifest)
+	if err != nil {
+		return err
+	}
+	bodyBuf := bytes.NewBuffer(payload)
+	qpairs := []string{}
+	if info.Yanked() != nil {
+		qpairs = append(qpairs, fmt.Sprintf("%s=%t", "yanked", *info.Yanked()))
```
AstroProfundis, line 358: Maybe it's better to define those keys as global vars or consts

pkg/repository/mirror.go
```diff
@@ -249,6 +326,71 @@ func (l *httpMirror) prepareURL(resource string) string {
...
+	}
+	if info.Standalone() != nil {
+		qpairs = append(qpairs, fmt.Sprintf("%s=%t", "standalone", *info.Standalone()))
+	}
+	if info.Hidden() != nil {
+		qpairs = append(qpairs, fmt.Sprintf("%s=%t", "hidden", *info.Hidden()))
+	}
+	qstr := ""
+	if len(qpairs) > 0 {
+		qstr = "?" + strings.Join(qpairs, "&")
```
AstroProfundis, line 368: Better use `net/url.Values.Encode()`

pkg/repository/mirror.go
```diff
@@ -249,6 +326,71 @@ func (l *httpMirror) prepareURL(resource string) string {
...
+	}
+	manifestAddr := fmt.Sprintf("%s/api/v1/component/%s/%s%s", l.Source(), sid, manifest.Signed.(*v1manifest.Component).ID, qstr)
+
+	resp, err := http.Post(manifestAddr, "text/json", bodyBuf)
```
AstroProfundis, line 372: The default HTTP client has no timeout set, we should create a client with proper timeout setting and use it for requests.

pkg/repository/mirror.go
```diff
@@ -249,6 +326,71 @@ func (l *httpMirror) prepareURL(resource string) string {
...
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode < 300 {
```
AstroProfundis, line 378: ditto, what about other redirect codes?

pkg/repository/model/model.go
```diff
@@ -0,0 +1,406 @@
...
+	cjson "github.com/gibson042/canonicaljson-go"
+	"github.com/juju/errors"
+	"github.com/pingcap/tiup/pkg/logger/log"
+	"github.com/pingcap/tiup/pkg/repository/store"
+	"github.com/pingcap/tiup/pkg/repository/v1manifest"
+	"github.com/pingcap/tiup/pkg/utils"
+)
+
+// Model defines operations on the manifests
+type Model interface {
```
AstroProfundis, line 29: I think it could be something like `model.Filesystem` or `model.Storage` instead of `model.Model`

pkg/repository/store/txn.go
```diff
@@ -0,0 +1,268 @@
...
+	filepath := t.store.path(filename)
+	if utils.IsExist(path.Join(t.root, filename)) {
+		filepath = path.Join(t.root, filename)
+	}
+	var wc io.ReadCloser
+	if file, err := os.Open(filepath); err == nil {
+		wc = file
+	} else if os.IsNotExist(err) && t.store.upstream != "" {
+		url := fmt.Sprintf("%s/%s", t.store.upstream, filename)
+		if resp, err := http.Get(url); err == nil {
```
AstroProfundis, line 137: Set timeout

AstroProfundis (commented):
We should add more comments in `txn.go` to describe how the merging process work.

cmd/mirror.go
lucklove: good idea

pkg/cluster/audit/audit.go
lucklove: https://github.com/pingcap/tiup/pull/916

pkg/repository/mirror.go
lucklove: It's unexpected, the server shouldn't return this.

AstroProfundis (approved):

PR #873, opened 2020-10-30
Correct 'Contribution Guide' URL

dveeden:
The old URL was returning a HTTP 404 Not Found.

This seems to be the result of https://github.com/pingcap/community/commit/41fd697b523adace5f717e97d433201336653fc4

Updated the URL to what seems to be the correct URL.

### Check List 

Tests 

 - No code

Release notes:

```release-note
NONE
```


dveeden:
https://github.com/pingcap/tidb/blob/master/README.md links to the same url that returns a 404. I assume the old url might be used in multiple places. If this PR gets merged I'll do a search-and-replace in the TiDB repo as well. The other option would to bring back the old URL.

lonng (approved):
LGTM

PR #878, opened 2020-11-02
clone: retry on download while network error

lucklove:
When clone from the mirror, it some times return the network error and abort.

Retry several times while there is network errors:
- unexpected EOF 
- stream error
- server returned 502 Bad Gateway

### Check List 

Tests 

 - Manual test

Related changes

Release notes:

```release-note
NONE
```


pkg/repository/mirror.go
```diff
@@ -248,18 +249,46 @@ func (l *httpMirror) prepareURL(resource string) string {
...
 // Download implements the Mirror interface
 func (l *httpMirror) Download(resource, targetDir string) error {
 	tmpFilePath := filepath.Join(l.tmpDir, resource)
 	dstFilePath := filepath.Join(targetDir, resource)
 	// downloaded file is stored in a temp directory and the temp directory is
 	// deleted at Close(), in this way an interrupted download won't remain
 	// any partial file on the disk
-	r, err := l.download(l.prepareURL(resource), tmpFilePath, 0)
+	var err error
+	_ = utils.Retry(func() error {
```
lonng, line 275: I think we should limit the retry count and retry duration.

lucklove: The utils.Retry has default value for that:
```golang
func Retry(doFunc func() error, opts ...RetryOption) error {
	var cfg RetryOption
	if len(opts) > 0 {
		cfg = opts[0]
	} else {
		cfg = RetryOption{
			Attempts: defaultAttempts,
			Delay:    defaultDelay,
			Timeout:  defaultTimeout,
		}
	}
```

pkg/repository/mirror.go
```diff
@@ -248,18 +249,46 @@ func (l *httpMirror) prepareURL(resource string) string {
...
+		var r io.ReadCloser
+		if err != nil && l.isRetryable(err) {
+			log.Warnf("failed to download %s(%s), retrying...", resource, err.Error())
+		}
+		if r, err = l.download(l.prepareURL(resource), tmpFilePath, 0); err != nil {
```
lonng, line 280: If `err !=nil && !l.isRetryable(err)`, the `download` will execute also.

pkg/repository/mirror.go
```diff
@@ -248,18 +249,46 @@ func (l *httpMirror) prepareURL(resource string) string {
...
 	dstFilePath := filepath.Join(targetDir, resource)
 	// downloaded file is stored in a temp directory and the temp directory is
 	// deleted at Close(), in this way an interrupted download won't remain
 	// any partial file on the disk
-	r, err := l.download(l.prepareURL(resource), tmpFilePath, 0)
+	var err error
+	_ = utils.Retry(func() error {
+		var r io.ReadCloser
+		if err != nil && l.isRetryable(err) {
+			log.Warnf("failed to download %s(%s), retrying...", resource, err.Error())
```
lonng, line 278: Why not move this line after L281?

lonng (commented):
Rest LGTM

pkg/repository/mirror.go
lucklove: Because it may reach to the max retry attempts, so we must print log before we really retry instead of after retry.

lonng (approved):
LGTM

Issue #666, opened 2020-08-10
CI may fail on tiup-dm audit


lucklove:
In integrating test,  the grep command will fail:

```sh
tiup-dm --yes deploy $name $version $topo -i ~/.ssh/id_rsa

tiup-dm list | grep "$name"

tiup-dm audit | grep "deploy $name $version"
```

After the `tiup-dm list` command, the `tiup-dm audit` only contains the `list` operation excepting the `deploy` operation.

lucklove:
This is not a problem any more.

PR #879, opened 2020-11-02
cluster/audit: audit log in seconds maybe leads to file overwriten

9547:
In my local dev's env, I've found the audit log file was base52 encoded with `time.Now().Unix()`, while at the same time, if the command is run fast enough, then the later file with the same name will cover the previous one.

In my local dev, I'm running `tiup-dm test-cmd`

```bash
bash /tiup-cluster/tests/tiup-dm/run.sh --native-ssh --do-cases test_cmd
# ... wait for this case finished
exit 1
```
the below case **maybe**(occasional) failed of this line https://github.com/pingcap/tiup/blob/11535d272a90f904ad5b8f73af5dddc224c2894d/tests/tiup-dm/test_cmd.sh#L28

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
Fix the issue which leads the audit log file be overwritten.
```

### DIFF
--- a/pkg/cluster/audit/audit.go
+++ b/pkg/cluster/audit/audit.go
@@ -60,7 +60,7 @@ func ShowAuditList(dir string) error {
 		if err != nil {
 			continue
 		}
-		t := time.Unix(ts, 0)
+		t := time.Unix(ts/1e9, 0)
 		cmd, err := firstLine(fi.Name())
 		if err != nil {
 			continue
@@ -82,7 +82,7 @@ func ShowAuditList(dir string) error {
 
 // OutputAuditLog outputs audit log.
 func OutputAuditLog(dir string, data []byte) error {
-	fname := filepath.Join(dir, base52.Encode(time.Now().Unix()))
+	fname := filepath.Join(dir, base52.Encode(time.Now().UnixNano()))
 	return ioutil.WriteFile(fname, data, 0644)
 }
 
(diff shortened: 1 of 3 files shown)


9547:
@lucklove  PTAL

AstroProfundis (approved):
LGTM, brilliant!

lucklove:
Awesome!!! I think this is what bother us for a very long time: https://github.com/pingcap/tiup/issues/666 (I closed that because it didn't reproduce these days, but I'm sure your PR finally fixed that)

pkg/cluster/audit/audit.go
```diff
@@ -103,7 +103,7 @@ func ShowAuditLog(dir string, auditID string) error {
 		return errors.Trace(err)
 	}
 
-	t := time.Unix(ts, 0)
+	t := time.Unix(ts/1e9, 0)
```
july2993, line 106: This is not compatible with the existing audit files ?

lucklove: Sorry for my careless, @9547 can your please open another PR to fix this?

9547: Sorry, I was inconsiderate, I'll fix in another PR

9547: fixed in #882 @lucklove @july2993 PTAL

Issue #853, opened 2020-10-20
Merge DM into cluster


9547:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Seems most of the business logic of [components/dm](components/dm/) and [pkg/cluster](pkg/cluster) overlaps. Should we consider separating the two parts of common logic to reduce code duplication, and reduce maintenance costs.

**Describe the feature you'd like:**

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

lucklove:
Hi~ @9547 
We split them as two components because dm has different version number with TiDB (the newest TiDB is v4.x but DM is 2.x). 
The `cluster` is responsible for tidb cluster while the `dm` is responsible for dm cluster.
However, we believe less is more, so reducing code duplication is always a goal. Maybe we can merge the duplicate logic instead of the components.

9547:
We can move the monitors logic first

PR #890, opened 2020-11-06
dry: merge dm's {alertmanager,grafana,prometheus} into cluster

9547:

Try to implement #853 

The logic of dm's monitor components almost the same as cluster's, merge them into one place, less code, easier to maintain.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
merge dm's {alertmanager,grafana,prometheus} into cluster
```


9547:
@lucklove PTAL, thanks

pkg/cluster/spec/spec.go
```diff
@@ -192,6 +194,11 @@ func (s *Specification) GetMonitoredOptions() *MonitoredOptions {
 	return &s.MonitoredOptions
 }
 
+// GetGlobalOptions implements Topology interface.
+func (s *Specification) GetGlobalOptions() GlobalOptions {
```
lucklove, line 198: This function is dup with `BaseTopo().GlobalOptions`

pkg/cluster/spec/alertmanager.go
```diff
@@ -62,7 +63,7 @@ func (s AlertManagerSpec) IsImported() bool {
 }
 
 // AlertManagerComponent represents Alertmanager component.
-type AlertManagerComponent struct{ *Specification }
+type AlertManagerComponent struct{ Topology }
```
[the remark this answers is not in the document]
9547: IMO, `Topology` is an interface which could hold for implementation is the concrete `Component`, so I change `AlertManagerComponent`, `MonitorComponent` and `GrafanaComponent` into this style, there are little differene between cluster and dm's topology.  While `Specification` stands for the specific `Component`, such as TiDB, TiKV, PD... So the other Components changed to `*Specification`

pkg/cluster/spec/alertmanager.go
```diff
@@ -76,8 +77,15 @@ func (c *AlertManagerComponent) Role() string {
 
 // Instances implements Component interface.
 func (c *AlertManagerComponent) Instances() []Instance {
-	ins := make([]Instance, 0, len(c.Alertmanager))
-	for _, s := range c.Alertmanager {
+	val, found := findSliceField(c.Topology, "Alertmanager")
```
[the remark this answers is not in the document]
9547: Good idea 👍

pkg/cluster/spec/grafana.go
```diff
@@ -160,13 +164,22 @@ func (i *GrafanaInstance) InitConfig(
 		return err
 	}
 
+	topo := reflect.ValueOf(i.topo)
+	if topo.Kind() == reflect.Ptr {
+		topo = topo.Elem()
+	}
+	val := topo.FieldByName("Monitors")
+	if (val == reflect.Value{}) {
```
AstroProfundis, line 172: Is this check necessary? There is check for monitor list length just below.

9547: This sentinel checks if the struct contains field `Monitors` or not, if not, then `val.Interface()` will panic

AstroProfundis (approved):

Issue #797, opened 2020-09-21
NodeID of pump/drainer is not import and keep consistent when import from ansible


july2993:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?

import a cluster from ansible.
restart pump or drainer.

2. What did you expect to see?
`./binlogctl -cmd pumps` or `./binlogctl -cmd drainers`
things keep the same as use ansible before.

3. What did you see instead?

new instance of pump/drainer with the new NodeID(the one we see from `tiup cluster display`)

ansible use the {host}:port, and when imported to tiup, we use the {ip:port} and NodeID

4. What version of TiUP are you using (`tiup --version`)?

lucklove:
We can get the real id from `pump` and `drainer` api.
I think we can transform the tiup-node-id to ansible-node-id via these apis before we take any action on pump and drainer.

july2993:
should add a `NodeID` field in spec, and set it when import form ansible

glkappe:
Regarding this improvement, I hope to ensure that the modification of tiup-node-id is consistent with the official document (also consistent with the deployment of tidb-ansible), It is recommended to continue to use `host:port` to set unspecified `pump_node-id`,and avoid secondary learning

https://docs.pingcap.com/zh/tidb/stable/tidb-binlog-configuration-file#node-id

Thanks, please feel free to contact me if you have any questions

lucklove:
@glkappe We want to change it to ip:port because all instances in TiUP is addressed by ip:port pair, if Drainer not, it breaks the consistency.

PR #903, opened 2020-11-12
Fix imported pump & drainer start script

lucklove:

https://github.com/pingcap/tiup/issues/797

- don't add `--node-id` flag for imported instance in start scripts (to avoid node-id change)
- fetch the real node id from pd(etcd) instead of hard coding one

### Check List 

Tests 

    - case 1
        - deploy a cluster with pump
        - modify the `--node-id` flag to anything others
        - modify meta.yaml and set `imported: true` for pump
        - try to scale-in that pump server
        - destroy the cluster
    - case 2
        - deploy a cluster with pump
        - modify meta.yaml and set `imported: true` for pump
        - reload cluster
        - the generated run_pump.sh should not contain `--node-id` flag
    - case 3
        - deploy a cluster with pump
        - scale-in pump

Related changes

Release notes:

```release-note
NONE
```


july2993 (approved):

july2993 (approved):

july2993 (commented):
better test with the clusters imported since this issue fixes those clusters.

lucklove:
> better test with the clusters imported since this issue fixes those clusters.

OK

Issue #877, opened 2020-11-02
node_exporter enable memory buddyinfo monitoring


wangkedba:
## Feature Request
node_export can collect exposes statistics of memory fragments as reported by /proc/buddyinfo. But this feature is disable by default. It's usefull to locate memory problems,so we hope to open it by default , both tiup and ansible deploy. 

Add parameters '--collector.buddyinfo' to the node_exporter startup script.
**Is your feature request related to a problem? Please describe:**

Locate memory problems
**Describe the feature you'd like:**

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

9547:
Good idea, I'll try to implement it those days.

PR #904, opened 2020-11-12
feat(node_exporter): add --collector.buddyinfo

9547:
implement #877 

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```


lucklove (approved):
LGTM

lucklove:
It seems that you changed the template file but forget to run `make pkger`. Please run `make pkger` and commit again.

9547:
Thanks for the advice, and i think maybe we can use tools like `pre-commit` to automatical run `make pkger` to generate the pkg if those template files changed, before commit

lucklove:
> Thanks for the advice, and i think maybe we can use tools like `pre-commit` to automatical run `make pkger` to generate the pkg if those template files changed, before commit

That's a good ideal, but it seems git hook only works on local, do you know is there anyway to run it in github server?

9547:
Seems github didn't support git  hooks, the  only we can use is web hooks

lucklove:
> Seems github didn't support git hooks, the only we can use is web hooks

Web hook can only do post-check, there is no need to add it since we have unit test to do it (and what I mean is that you should make pkger and push again to pass the unit test)

Issue #898, opened 2020-11-10
TiUP destroy cluster ，ssh-key public residual


buggithubs:
## Feature Request

1. TiUP 删除 TiDB 集群后，目标机器中残留 cluster key 的公钥信息 ；是否可以清理掉？
2. TiUP deploy 部署时，在 + Generate SSH keys ... Done 步骤执行 ctrl + c ，再次部署时，目标机器会出现两条相同的 cluster key 公钥信息。

9547:
Good idea, sounds we need two features:

1. de-dup the same  public key;
2. add an argument to remove the public key or not(default is false). But how do you know that the key was added by tiup deploy and not by another user?

lucklove:
> But how do you know that the key was added by tiup deploy and not by another user?

@9547 We can, TiUP will store generated key in ~/.tiup/storage/cluster/clusters/<cluster-name>/ssh，which is used by only that cluster. once destroyed, the public key added to remote is not useful anymore. Because of this, I think there is no need to add an argument to decide removing or not, just remove it, Or the argument's default value can be true

lucklove:
> /close

@hi-rustin Can we have this command?

PR #910, opened 2020-11-16
Feature/remove public key after destroyed

9547:
Try to implement #898 

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
Remove SSH public key if the cluster was destroyed 
```

### DIFF
--- a/components/cluster/command/prune.go
+++ b/components/cluster/command/prune.go
@@ -73,7 +73,7 @@ func destroyTombstoneIfNeed(clusterName string, metadata *spec.ClusterMeta, opt 
 		return perrs.AddStack(err)
 	}
 
-	nodes, err := operator.DestroyTombstone(ctx, topo, true /* returnNodesOnly */, opt, tlsCfg)
+	nodes, err := operator.DestroyTombstone(ctx, topo, true /* returnNodesOnly */, opt, tlsCfg, ctx.PublicKeyPath)
 	if err != nil {
 		return perrs.AddStack(err)
 	}
@@ -93,7 +93,7 @@ func destroyTombstoneIfNeed(clusterName string, metadata *spec.ClusterMeta, opt 
 
 	log.Infof("Start destroy Tombstone nodes: %v ...", nodes)
 
-	_, err = operator.DestroyTombstone(ctx, topo, false /* returnNodesOnly */, opt, tlsCfg)
+	_, err = operator.DestroyTombstone(ctx, topo, false /* returnNodesOnly */, opt, tlsCfg, ctx.PublicKeyPath)
 	if err != nil {
 		return perrs.AddStack(err)
 	}
--- a/components/cluster/command/prune.go
+++ b/components/cluster/command/prune.go
@@ -73,7 +73,7 @@ func destroyTombstoneIfNeed(clusterName string, metadata *spec.ClusterMeta, opt 
 		return perrs.AddStack(err)
 	}
 
-	nodes, err := operator.DestroyTombstone(ctx, topo, true /* returnNodesOnly */, opt, tlsCfg, ctx.PublicKeyPath)
+	nodes, err := operator.DestroyTombstone(ctx, topo, true /* returnNodesOnly */, opt, tlsCfg)
 	if err != nil {
 		return perrs.AddStack(err)
 	}
@@ -93,7 +93,7 @@ func destroyTombstoneIfNeed(clusterName string, metadata *spec.ClusterMeta, opt 
 
 	log.Infof("Start destroy Tombstone nodes: %v ...", nodes)
 
-	_, err = operator.DestroyTombstone(ctx, topo, false /* returnNodesOnly */, opt, tlsCfg, ctx.PublicKeyPath)
+	_, err = operator.DestroyTombstone(ctx, topo, false /* returnNodesOnly */, opt, tlsCfg)
 	if err != nil {
 		return perrs.AddStack(err)
 	}
--- a/components/dm/command/scale_in.go
+++ b/components/dm/command/scale_in.go
@@ -46,7 +46,7 @@ func newScaleInCmd() *cobra.Command {
 				b.Func(
 					fmt.Sprintf("ScaleInCluster: options=%+v", gOpt),
 					func(ctx *task.Context) error {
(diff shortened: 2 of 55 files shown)


components/cluster/command/prune.go
```diff
@@ -73,7 +73,7 @@ func destroyTombstoneIfNeed(clusterName string, metadata *spec.ClusterMeta, opt
 		return perrs.AddStack(err)
 	}
 
-	nodes, err := operator.DestroyTombstone(ctx, topo, true /* returnNodesOnly */, opt, tlsCfg)
+	nodes, err := operator.DestroyTombstone(ctx, topo, true /* returnNodesOnly */, opt, tlsCfg, ctx.PublicKeyPath)
```
AstroProfundis, line 76: `ctx` is passed to `DestroyTombstone()` so I don't think it's necessary to pass `ctx.PublicKeyPath` again.

components/dm/command/scale_in.go
```diff
@@ -46,7 +46,7 @@ func newScaleInCmd() *cobra.Command {
 				b.Func(
 					fmt.Sprintf("ScaleInCluster: options=%+v", gOpt),
 					func(ctx *task.Context) error {
-						return ScaleInDMCluster(ctx, metadata.Topology, gOpt)
+						return ScaleInDMCluster(ctx, metadata.Topology, gOpt, ctx.PublicKeyPath)
```
AstroProfundis, line 49: ditto

pkg/cluster/manager.go
```diff
@@ -252,7 +252,8 @@ func (m *Manager) RestartCluster(clusterName string, options operator.Options) e
 func (m *Manager) ListCluster() error {
 	names, err := m.specManager.List()
 	if err != nil {
-		return perrs.AddStack(err)
+		log.Warnf("List cluster error %v", err)
+		// return perrs.AddStack(err)
```
AstroProfundis, line 256: Why not returning here?

pkg/cluster/manager.go
```diff
@@ -381,7 +382,7 @@ func (m *Manager) DestroyCluster(clusterName string, gOpt operator.Options, dest
 			}, tlsCfg)
 		}).
 		Func("DestroyCluster", func(ctx *task.Context) error {
-			return operator.Destroy(ctx, topo, destroyOpt)
+			return operator.Destroy(ctx, topo, ctx.PublicKeyPath, destroyOpt)
```
AstroProfundis, line 385: ditto, the new param is not necessary.

pkg/cluster/operation/destroy.go
```diff
@@ -171,6 +195,40 @@ func DeleteGlobalDirs(getter ExecutorGetter, host string, options *spec.GlobalOp
...
+	publicKey, err := ioutil.ReadFile(pubKeyPath)
+	if err != nil {
+		return perrs.Trace(err)
+	}
+	pubKey := string(bytes.TrimSpace(publicKey))
+	pubKey = strings.ReplaceAll(pubKey, "/", "\\/")
+	pubKeysFile := executor.FindSSHAuthorizedKeysFile(e)
+
+	c := module.ShellModuleConfig{
+		Command:  fmt.Sprintf("sed -i '/%s/d' %s", pubKey, pubKeysFile),
```
AstroProfundis, line 211: I think it might be better to just comment that line.

AstroProfundis (commented):
And as we support deploying cluster with user specified SSH key (instead of generating new), so it would be better to check for duplicate during deploy process, and also not deleting the public key by default on scaling in and destroying.

lucklove:
Seems test broken: `cannot stat '~/.tiup/storage/dm/clusters/test_cmd/ssh/id_rsa'`
May related to this PR.

components/cluster/command/prune.go
9547: The first argument inside `DestoryTombestone()` is `ExecutorGetter` interface: https://github.com/pingcap/tiup/blob/cc41ad5af87bde821ecde48d63dd7c1c40d72ea2/pkg/cluster/operation/destroy.go#L414-L420  So I need to cast it into `task.Context`, but this results in `import cycle` error. Please let me know if I've misssed something.

pkg/cluster/manager.go
9547: Sorry, this is a mistake, will be reverted back.

components/cluster/command/prune.go
lucklove: Maybe we can define another interface for the first param of `DestroyTombstone`:
```golang
type SSHExecutorGetter interface {
    ExecutorGetter
    GetSSHKeySet() (string, string) // return both priv and public key
}
```

9547: @lucklove Or should we extend the original `ExecutorGetter` interface with:

```go
type ExecutorGetter interface {
	Get(host string) (e executor.Executor)
	GetSSHKeySet() (string, string)
}
```

pkg/cluster/operation/destroy.go
9547: Thanks for the advice.

components/cluster/command/prune.go
9547: @lucklove I did it this way first, if you have any question, please let me know.

lucklove: > @lucklove Or should we extend the original `ExecutorGetter` interface with:
> 
> ```go
> type ExecutorGetter interface {
> 	Get(host string) (e executor.Executor)
> 	GetSSHKeySet() (string, string)
> }
> ```

It's OK at this time. In the future there may be some executors that don't relay on ssh, at that time we can refactor this.

lucklove:
What's the status of this? It seems still using ctx.PublicKeyPath? https://github.com/pingcap/tiup/pull/910/files#diff-c2cb8d42eb23179c5056c3d5c626c7f2ddabad3db5171c8f869d59f75ddffc42R76

9547:
@lucklove PTAL

lucklove (commented):
Rest LGTM

lucklove (approved):

AstroProfundis (approved):

PR #908, opened 2020-11-16
Feature/add log dir

9547:
### What problem does this PR solve? 

fix #897 

### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test (add detailed scripts or steps below)
 - No code

Code changes

 - Has exported function/method change
 - Has exported variable/fields change
 - Has interface methods change
 - Has persistent data change

Side effects

 - Possible performance regression
 - Increased code complexity
 - Breaking backward compatibility

Related changes

 - Need to cherry-pick to the release branch
 - Need to update the documentation

Release notes:

```release-note
Add `$TIUP_LOG_PATH` to store log file
```

PR #917, opened 2020-11-17
Fix: github workflow upload logs dir error

9547:

As #908 set the default log dir to `$TIUP_HOME/logs`, we need to change the control's log path. And one more thing, the SSHD process was not activated in host `tiup-cluster-control`, and the scp always failed, such as:

```bash
Run docker exec tiup-cluster-control bash /tiup-cluster/tests/tiup-cluster/script/pull_log.sh /tiup-cluster/logs
172.19.0.100
ssh: connect to host 172.19.0.100 port 22: Connection refused
172.19.0.101
Warning: Permanently added '172.19.0.101' (ECDSA) to the list of known hosts.
172.19.0.102
Warning: Permanently added '172.19.0.102' (ECDSA) to the list of known hosts.
172.19.0.103
Warning: Permanently added '172.19.0.103' (ECDSA) to the list of known hosts.
172.19.0.104
Warning: Permanently added '172.19.0.104' (ECDSA) to the list of known hosts.
172.19.0.105
Warning: Permanently added '172.19.0.105' (ECDSA) to the list of known hosts.
```
Replace the `scp` with local cp for control's node.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```


.github/workflows/integrate-dm.yaml
```diff
@@ -66,6 +66,23 @@ jobs:
           # ref: https://stackoverflow.com/questions/43099116/error-the-input-device-is-not-a-tty
           docker exec tiup-cluster-control bash /tiup-cluster/tests/tiup-dm/run.sh ${{matrix.cases}}
 
+      - name: Collect component log
```
lucklove, line 69: Why upload log? I think just replacing `cat ./tests/*.log` with a correct log path will be fine.

9547: As cluster https://github.com/pingcap/tiup/blob/cc41ad5af87bde821ecde48d63dd7c1c40d72ea2/.github/workflows/integrate-cluster-cmd.yaml#L70 uploads it, so I think dm should upload it too.

lucklove: OK, How about add `cat /path/to/*.log` too? By this way we can check what happened without downloading the logs when some error occurs in test.

9547: I don't think it's a good idea, as the debug file may be very large and may not be human-readable on the terminal, BTW, I'll give it a try.

9547: I'm going to adjust my strategy to look like this:if the test passed, no log were uploaded and print out; else upload and print it.

lucklove: That's cool

.github/workflows/integrate-cluster-cmd.yaml
```diff
@@ -64,16 +64,17 @@ jobs:
...
       - name: Collect component log
         working-directory: ${{ env.working-directory }}
-        # if: steps.test.outputs.exit_code != 0
-        if: always()
+        if: steps.test.outputs.exit_code != 0
+        # if: always()
         run: |
           docker exec tiup-cluster-control bash /tiup-cluster/tests/tiup-cluster/script/pull_log.sh /tiup-cluster/logs
           ls ${{ env.working-directory }}
+          cat /tiup-cluster/logs/*.log || true
```
lucklove, line 72: Prefer to move this line to the `Output cluster debug log` step because its' name is `Output cluster debug log`...

9547: Thanks for your advice

lucklove (approved):
LGTM

PR #918, opened 2020-11-18
fix(install): install script export path

Obooman:
Add a new line for the export path of TIUP, fix the issue caused by adding the path export to the previous line.

Adding a `\n` new line in front of the inject sentence

### Check List 

Tests 
 - Integration test

Code changes

Release notes:
```release-note
NONE
```


lucklove (approved):
LGTM

PR #922, opened 2020-11-19
Fix: duplicated pd_servers.name in the topology before truly deploy

### LINKED ISSUE #764: check duplicated  pd_servers.name in the topology before truly deploy the cluster
https://github.com/pingcap/tiup/blob/41f1373cb21df335773ba5b916fc586ff34c0adb/examples/topology.example.yaml#L77
if contains duplicated pd name, we can deploy success, but failed to start pd.

We can add the duplicated name check in https://github.com/pingcap/tiup/blob/41f1373cb21df335773ba5b916fc586ff34c0adb/pkg/cluster/spec/validate.go#L695

So when a user using a topology with duplicated pd name, we will hint the user to update the topology.

anywhy:
fix #764  pd_servers.name has same name 

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```


pkg/cluster/spec/validate.go
```diff
@@ -794,6 +794,22 @@ func (s *Specification) validateTLSEnabled() error {
 	return nil
 }
 
+func (s *Specification) validatePdPodNames() error {
+	// check pdserver pod name
+	if len(s.PDServers) > 1 {
```
lucklove, line 799: This check is redundant, prefer to remove the if statement

pkg/cluster/spec/validate.go
```diff
@@ -794,6 +794,22 @@ func (s *Specification) validateTLSEnabled() error {
 	return nil
 }
 
+func (s *Specification) validatePdPodNames() error {
```
lucklove, line 797: prefer to use the name `validatePDNames` because `Pod` is a term of K8S and we can really deploy TiDB to K8S. So we'd better change it to avoid misunderstanding

pkg/cluster/spec/validate.go
```diff
@@ -794,6 +794,22 @@ func (s *Specification) validateTLSEnabled() error {
...
+	// check pdserver pod name
+	if len(s.PDServers) > 1 {
+		cnt := make(map[string]int)
```
lucklove, line 800: Prefer to use [StringSet](https://github.com/pingcap/tiup/blob/master/pkg/set/string_set.go):

```
names := set.NewStringSet()

for _, pd := range s.PDServers {
    if pd.Name == "" {
        continue
    }
    if names.Exist(pd.Name) { 
        return errors.Errorf("xxxxxx")
    }
    names.Insert(pd.Name)
}
```

lucklove (commented):
Rest LGTM

pkg/cluster/spec/validate.go
anywhy: got it

AstroProfundis (approved):

PR #926, opened 2020-11-20
playground: allow a pure TiKV cluster

sticnarf:

I usually need a TiKV environment that is not polluted by TiDB. But the TiDB number check in playground prevents me from it! 

Do not check TiDB number in playground.

### Check List 

Tests 

`tiup playground --db 0` works \o/

Release notes:

```release-note
Allow creating a pure TiKV cluster in TiUP playground.
```


sticnarf:
PTAL @lucklove

lucklove (approved):
LGTM

PR #930, opened 2020-11-23
Fix dm monitor config

lucklove:
Introduced by https://github.com/pingcap/tiup/pull/890:
After deploy/scale-out grafana/prometheus with tiup-dm, the dashborads/*.json and *.rule.yml will be the ones of tidb instead of dm.

Add deleted function [MonitorInstance.installRules](https://github.com/pingcap/tiup/pull/890/files#diff-3899b00869d9aaa51f876b71e527e043d1863cb9defa69cf0dc6a0d0b18040b8L205) and [GrafanaInstance.initDashboards](https://github.com/pingcap/tiup/pull/890/files#diff-031e0480914e60bfb94c812eaeb71bd382f3ea774dd71927dab7f33276af5812L164) back.

### Check List 

Tests 

 - Integration test

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/pkg/cluster/spec/grafana.go
+++ b/pkg/cluster/spec/grafana.go
@@ -149,6 +149,12 @@ func (i *GrafanaInstance) InitConfig(
 		return err
 	}
 
+	if i.topo.Type() == TopoTypeDM {
+		if err := i.installDashboards(e, paths.Deploy, clusterName, clusterVersion); err != nil {
+			return errors.Annotate(err, "install dashboards")
+		}
+	}
+
 	// initial dashboards/*.json
 	if err := i.initDashboards(e, i.InstanceSpec.(GrafanaSpec), paths, clusterName); err != nil {
 		return errors.Annotate(err, "initial dashboards")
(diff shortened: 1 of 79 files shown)


9547:
@lucklove Sorry for my buggy code. 😭

lucklove:
> @lucklove Sorry for my buggy code. 😭

It's OK, we cannot avoid bugs completely. The link is not to say that's you who introduce the bug, it‘s just used to mark the influence of the bug (when it start and when it's fixed). Fortunately, it's not released yet, so no one will get into trouble due to this bug.

BTW, you will be a reviewer of TiUP after this PR: https://github.com/pingcap/community/pull/351. Congratulations! Be carefully about your `LGTM` on every PR: make sure you really understand and agree with the code before you give the `LGTM` :)

lucklove:
@9547 Can you help on reviewing this? You should be able to do this now, I think.

9547:
Thanks for the invite, let me have a try.

pkg/cluster/spec/grafana.go
```diff
@@ -223,6 +229,49 @@ func (i *GrafanaInstance) initDashboards(e executor.Executor, spec GrafanaSpec,
 	return nil
 }
 
+// installDashboards is used by dm cluster
+func (i *GrafanaInstance) installDashboards(e executor.Executor, deployDir, clusterName, clusterVersion string) error {
```
9547, line 233: Could you please add some comments about why `installDashboards` is only used by dm. BTW, if this install is only used by DM, should we rename it into `installDMDashboards()` or something similar?

pkg/cluster/spec/prometheus.go
```diff
@@ -261,6 +267,50 @@ func (i *MonitorInstance) InitConfig(
 	return e.Transfer(fp, dst, false)
 }
 
+// installRules is used by dm cluster
```
9547, line 270: ditto

pkg/cluster/spec/util.go
```diff
@@ -155,3 +155,9 @@ func MultiDirAbs(user, paths string) []string {
 	}
 	return dirs
 }
+
+// PackagePath return the tar bar path
+func PackagePath(comp string, version string, os string, arch string) string {
+	fileName := fmt.Sprintf("%s-%s-%s-%s.tar.gz", comp, version, os, arch)
+	return ProfilePath(TiOpsPackageCacheDir, fileName)
```
9547, line 162: Should we rename `TiOpsPackageCacheDir` to `TiUPPackageCacheDir`?

tests/tiup-dm/test_cmd.sh
```diff
@@ -69,6 +69,18 @@ topo_worker=./topo/full_scale_in_dm-worker.yaml
...
+yes | tiup-dm scale-in $name -N $ipprefix.101:3000
+wait_instance_num_reach $name $total_sub_one
+
+echo "start scale out grafana"
+topo_grafana=./topo/full_scale_in_grafana.yaml
+sed "s/__IPPREFIX__/$ipprefix/g" $topo_grafana.tpl > $topo_grafana
+yes | tiup-dm scale-out $name $topo_grafana
+
+# test grafana config
+tiup-dm exec $name -N $ipprefix.101 --command "ls /home/tidb/deploy/grafana-3000/dashboards/dm.json && ! grep magic-string-for-test /home/tidb/deploy/grafana-3000/dashboards/dm.json"
```
9547, line 82: Should `! grep` be the opposite? Curious abot the test-case was passed, but this file contains this string 
https://github.com/pingcap/tiup/blob/7c52950c0bf42e664777a6839d60af6ad5e33dca/tests/tiup-dm/local/grafana/dm.json#L177

pkg/cluster/spec/util.go
lucklove: good idea. TiOps is the predecessor of `tiup-cluster`. Since `tiup-dm` and `tiup-cluster` both use this variable, I think it should be `TiUPPackageCacheDir`

tests/tiup-dm/test_cmd.sh
lucklove: Because in the scale-out topo ($topo_grafana), we didn't specify dashboard_dir, so it will use the built-in one which doesn't contains the magic string.

lucklove:
@9547 PTAL

pkg/cluster/spec/grafana.go
```diff
@@ -149,6 +149,12 @@ func (i *GrafanaInstance) InitConfig(
 		return err
 	}
 
+	if i.topo.Type() == TopoTypeDM {
```
9547, line 152: IMO, we can put this line inside ｀installDashboard｀

pkg/cluster/spec/prometheus.go
```diff
@@ -250,6 +250,12 @@ func (i *MonitorInstance) InitConfig(
 		}
 	}
 
+	if i.topo.Type() == TopoTypeDM {
```
9547, line 253: ditto

lucklove:
@9547 done

9547 (approved):

lucklove:
@9547 Can you use `/approve` like me?

lucklove:
@AndreMouche PTAL, the bot confused me.

lucklove:
@9547 Can you help on debug this? Try `/lgtm` this time

9547:
Seems the bot not respond

lucklove:
Ok, maybe we should fire this stupid bot and hire a smarter one next week...

9547:
@ti-srebot are you ok?

PR #935, opened 2020-11-25
Fix/GitHub action failure

9547:

This case failed https://github.com/pingcap/tiup/runs/1450410729, but the log upload steps are skiped

image

Seems we need to use `if: ${{ failure() }}` condition to check the job status, ref [Setting exit codes for actions - GitHub Docs](https://docs.github.com/en/free-pro-team@latest/actions/creating-actions/setting-exit-codes-for-actions#about-exit-codes).

After modified, the log upload steps are executed:

image

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/.github/workflows/integrate-cluster-cmd.yaml
+++ b/.github/workflows/integrate-cluster-cmd.yaml
@@ -87,7 +87,7 @@ jobs:
           docker ps
           df -h
           free -h
-          cat /tiup-cluster/logs/*.log || true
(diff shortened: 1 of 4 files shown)


9547:
@lucklove PTAL

lucklove (approved):
LGTM

PR #937, opened 2020-11-25
fix importing grafana admin user / password from ansible

### LINKED ISSUE #932: User defined Grafana username and password not imported from Ansible
## Bug Report

1. What did you do?
If the user set custom Grafana username and password in Ansible inventory file, and have not changed it after deployment, those settings are ignored during `import` and the imported Grafana falls back to default `admin / admin` credentials.
图片

The configs are kept and correctly set after import.

AstroProfundis:
Fix #932 

### Check List 

Tests 

 - Unit test

Code changes

 - Has persistent data change

Related changes

Release notes:

```release-note
fix: custom grafana admin username / password not imported from ansible deployed cluster
```


9547:
@AstroProfundis Please add a test case. :)

AstroProfundis:
This case is already added in the `test-data/inventory.ini`, it's not a whole new test, but two new fields in the test inventory.

9547:
Should we apply username/password for DM's case? https://github.com/pingcap/tiup/blob/cc41ad5af87bde821ecde48d63dd7c1c40d72ea2/components/dm/ansible/import.go#L315-L316 or https://github.com/pingcap/tiup/blob/cc41ad5af87bde821ecde48d63dd7c1c40d72ea2/components/dm/ansible/import.go#L528-L562

templates/config/grafana.ini.tpl
```diff
@@ -119,10 +119,10 @@ check_for_updates = true
 #################################### Security ####################################
 [security]
 # default admin user, created on startup
-;admin_user = admin
+admin_user = {{.Username}}
```
lucklove, line 122: What if the Username is empty?

AstroProfundis: It would be set to `admin` during unmarshal, the default value is set in spec tag

PR #938, opened 2020-11-25
Fix bug for only one directory in TiFlash storage.* configuration

JaySon-Huang:

If there is only one directory in TiFlash storage.* configuration, the TiFlashSpec DataDir become unexpected value.
Introduced by #931 

Fix a bug that if there is only one directory in TiFlash storage.* configuration

### Check List 

Tests 

 - Unit test

Code changes

 - N/A

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```


lucklove (approved):
LGTM

Issue #691, opened 2020-08-19
Add `listen_host` for all components


lonng:
Add `listen_host` for all components

PR #941, opened 2020-11-26
Fix PD --peer-urls flag to use listen_host variable

### LINKED ISSUE #337: failed to start pd with the host name in topo.yaml
topo.yaml
```
tidb_servers:
  - host: n1
  - host: n2

pd_servers:
  - host: n3
  - host: n4
  - host: n5

tikv_servers:
  - host: n3
  - host: n4
  - host: n5
```

run deploy and start, failed to start pd

pd log:
```
[2020/03/24 16:27:17.759 +00:00] [INFO] [server.go:217] ["PD Config"] [config="{\"client-urls\":\"http://n3:2379\",\"peer-urls\":\"http://n3:2380\",\"advertise-client-urls\":\"http://n3:2379\",\"advertise-peer-urls\":\"http://n3:2380\",\"name\":\"pd-n3\",\"data-dir\":\"/home/tidb/deploy/pd-2379/data\",\"force-new-cluster\":false,\"enable-grpc-gateway\":true,\"initial-cluster\":\"pd-n3=http://n3:2380,pd-n4=http://n4:2380,pd-n5=http://n5:2380\",\"initial-cluster-state\":\"new\",\"join\":\"\",\"lease\":3,\"log\":{\"level\":\"\",\"format\":\"text\",\"disable-timestamp\":false,\"file\":{\"filename\":\"logs/pd.log\",\"max-size\":300,\"max-days\":0,\"max-backups\":0},\"development\":false,\"disable-caller\":false,\"disable-stacktrace\":false,\"disable-error-verbose\":true,\"sampling\":null},\"log-file\":\"\",\"log-level\":\"\",\"tso-save-interval\":\"3s\",\"metric\":{\"job\":\"pd-n3\",\"address\":\"\",\"interval\":\"15s\"},\"schedule\":{\"max-snapshot-count\":3,\"max-pending-peer-count\":16,\"max-merge-region-size\":20,\"max-merge-region-keys\":200000,\"split-merge-interval\":\"1h0m0s\",\"enable-one-way-merge\":\"false\",\"enable-cross-table-merge\":\"false\",\"patrol-region-interval\":\"100ms\",\"max-store-down-time\":\"30m0s\",\"leader-schedule-limit\":4,\"leader-schedule-policy\":\"count\",\"region-schedule-limit\":2048,\"replica-schedule-limit\":64,\"merge-schedule-limit\":8,\"hot-region-schedule-limit\":4,\"hot-region-cache-hits-threshold\":3,\"store-balance-rate\":15,\"tolerant-size-ratio\":0,\"low-space-ratio\":0.8,\"high-space-ratio\":0.6,\"scheduler-max-waiting-operator\":5,\"enable-remove-down-replica\":\"true\",\"enable-replace-offline-replica\":\"true\",\"enable-make-up-replica\":\"true\",\"enable-remove-extra-replica\":\"true\",\"enable-location-replacement\":\"true\",\"enable-debug-metrics\":\"false\",\"schedulers-v2\":[{\"type\":\"balance-region\",\"args\":null,\"disable\":false,\"args-payload\":\"\"},{\"type\":\"balance-leader\",\"args\":null,\"disable\":false,\"args-payload\":\"\"},{\"type\":\"hot-region\",\"args\":null,\"disable\":false,\"args-payload\":\"\"},{\"type\":\"label\",\"args\":null,\"disable\":false,\"args-payload\":\"\"}],\"schedulers-payload\":null,\"store-limit-mode\":\"manual\"},\"replication\":{\"max-replicas\":3,\"location-labels\":\"\",\"strictly-match-label\":\"false\",\"enable-placement-rules\":\"false\"},\"pd-server\":{\"use-region-storage\":\"true\",\"max-reset-ts-gap\":86400000000000,\"key-type\":\"table\",\"runtime-services\":\"\",\"metric-storage\":\"\"},\"cluster-version\":\"0.0.0\",\"quota-backend-bytes\":\"8GiB\",\"auto-compaction-mode\":\"periodic\",\"auto-compaction-retention-v2\":\"1h\",\"TickInterval\":\"500ms\",\"ElectionInterval\":\"3s\",\"PreVote\":true,\"security\":{\"cacert-path\":\"\",\"cert-path\":\"\",\"key-path\":\"\",\"client-cert-auth\":false,\"cert-allowed-cn\":null},\"label-property\":null,\"WarningMsgs\":null,\"DisableStrictReconfigCheck\":false,\"HeartbeatStreamBindInterval\":\"1m0s\",\"LeaderPriorityCheckInterval\":\"1m0s\",\"enable-dynamic-config\":true,\"EnableDashboard\":true}"]
[2020/03/24 16:27:17.780 +00:00] [INFO] [server.go:190] ["register REST path"] [path=/pd/api/v1]
[2020/03/24 16:27:17.787 +00:00] [INFO] [apiserver.go:109] ["Enabled Dashboard API"] [path=/dashboard/api/]
[2020/03/24 16:27:17.787 +00:00] [INFO] [server.go:190] ["register REST path"] [path=/dashboard/api/]
[2020/03/24 16:27:17.787 +00:00] [INFO] [uiserver.go:38] ["Enabled Dashboard UI"] [path=/dashboard/]
[2020/03/24 16:27:17.787 +00:00] [INFO] [server.go:190] ["register REST path"] [path=/dashboard/]
[2020/03/24 16:27:17.787 +00:00] [FATAL] [main.go:125] ["run server failed"] [error="expected IP in URL for binding (http://n3:2380)"] [stack="github.com/pingcap/log.Fatal\n\t/home/jenkins/agent/workspace/release_tidb_4.0/go/pkg/mod/github.com/pingcap/log@v0.0.0-20200117041106-d28c14d3b1cd/global.go:59\nmain.main\n\t/home/jenkins/agent/workspace/release_tidb_4.0/go/src/github.com/pingcap/pd/cmd/pd-server/main.go:125\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:203"]
```

note the error `expected IP in URL for binding (http://n3:2380)`

fln:

This allows us to use server hostname instead of IP address. Without
this change if PD service is declared using hostname it fails to start,
reporting an error that it can not bind to given interface.

Metadata variable `listen_host` was introduced in #495 and is already used for
the --client-urls flag.

This should resolve #337 and partially implement #691 .

This commit updates run_pd.sh template to use `listen_host` for
--peer-urls flag instead of `host`.

### Check List 

Tests

 - No code

Side effects

If existing installations relied on `host` field to be used as address for accepting peer connections this change could be treated as breaking change. Default value of `listen_host` is 0.0.0.0 so it should not break existing clusters.

Some might want to use different listen addresses for peer and client connections. However I think this would add unnecessary complexity and not worth supporting.

Related changes

Not sure where metadata file is properly documented. So far I have used source code to understand what is expected in the metadata.yaml file.

Release notes:

```release-note
Support using hostname instead of IP for PD servers in host section.
```


9547:
@fln Could you please add some test case, such as the case of #337

fln:
@9547 Thanks for feedback and suggestions.

Updated the tests as requested. First attempt was to create a separate test just for host name support. It turned out the integration tests are quite convoluted and performs some tricks to actually work with IP addresses which are no longer necessary if hostnames are used.

In the end all IP addresses in tests were replaced with docker hostnames `n1` - `n5`. One could argue that it is good idea to use hostnames as a default since we have a support for it already. It will also prevent us from breaking hostname support in the future. Furthermore it is more likely that code which works with hostnames will work with IP addresses without modifications, which is not true in the opposite direction.

Key changes that were made since initial commit:

* Discovered that `run_pd_scale.sh.tpl` is used for scaling it needed same template change that was done in the initial commit. This made refactoring the tests worthwhile and saved us tripping on it at the later stage.
* Made a small change to the automated TLS certificate generator, to generate correct certificates if hostname are used - IP and hostname is goes to different x509 sections.
* Made all test topology files stable. Removed `sed` invocations that used to create actual topology files from the topology templates. Running tests no longer creates temporary topology files (it was kind of annoying as docker running with root created temp files owned by root). In the process file `full_scale_in_tidb.yaml` had to be split into two versions, apparently this file was being mutated in the middle of the tests.

9547:
Thanks, but could you please split this pr into two or three

- this pr: fix the pd url
- refact tests: use hostname instead of ip for common tests, but keep some ip cases
- fix the x509 issue 

thanks again.

fln:
PRs are split. Just noticed that I have overlooked the request to have mixed IP and host name tests.

Mixing host names and IP addresses in tests is not that trivial.

If we mix host names and IP address in a single topology - tiup-cluster will treat them as different hosts and complain about ports being already in use by external services.

Leaving some tests with host names while other test with IP addresses also causes issue because full.yaml is shared between scale and cmd tests.

Is there any chance full test suite switch to host names would be accepted :smile: ?

Alternatively would a simple test of just building a minimal cluster and tearing it down using only IP addresses suffice? My opinion is that if we have code that works with host names it should with high confidence work with IP addresses as well without explicit testing.

fln:
Closing this PR, it is replaced by:
* #948 
* #949 
* #950

Issue #742, opened 2020-09-01
Support -N config to specific node in tiup cluster clean


Win-Man:
## Feature Request

**Is your feature request related to a problem? Please describe:**

I want to clean the log on specific node.

**Describe the feature you'd like:**

Support -N config in `tiup cluster clean` command.

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

PR #946, opened 2020-11-28
Support --node/-n config to specific node in tiup cluster clean(#742)

anywhy:
(#742) Support -n config in tiup cluster clean command.

``` bash
 $ tiup cluster clean <cluster-name> --log -n 172.16.13.12
 $ tiup cluster clean <cluster-name> --all -n 172.16.13.12
 $ tiup cluster clean <cluster-name> --all --node 172.16.13.11:9000
 $ tiup cluster clean <cluster-name> --log -n 172.19.0.105 -n 172.19.0.104 --ignore-node 172.19.0.105
```

Release notes:

```release-note
NONE
```

### DIFF
--- a/components/cluster/command/clean.go
+++ b/components/cluster/command/clean.go
@@ -32,7 +32,9 @@ You can retain some nodes and roles data when cleanup the cluster, eg:
     $ tiup cluster clean <cluster-name> --data
     $ tiup cluster clean <cluster-name> --all --ignore-role prometheus
     $ tiup cluster clean <cluster-name> --all --ignore-node 172.16.13.11:9000
-    $ tiup cluster clean <cluster-name> --all --ignore-node 172.16.13.12`,
+    $ tiup cluster clean <cluster-name> --all --ignore-node 172.16.13.12
+    $ tiup cluster clean <cluster-name> --all -n 172.16.13.12
+    $ tiup cluster clean <cluster-name> --all --node 172.16.13.11:9000`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if len(args) != 1 {
 				return cmd.Help()
@@ -56,6 +58,7 @@ You can retain some nodes and roles data when cleanup the cluster, eg:
 
 	cmd.Flags().StringArrayVar(&cleanOpt.RetainDataNodes, "ignore-node", nil, "Specify the nodes or hosts whose data will be retained")
 	cmd.Flags().StringArrayVar(&cleanOpt.RetainDataRoles, "ignore-role", nil, "Specify the roles whose data will be retained")
+	cmd.Flags().StringArrayVarP(&cleanOpt.CleanDataNodes, "node", "n", nil, "Specify the nodes whose log will be clean")
 	cmd.Flags().BoolVar(&cleanOpt.CleanupData, "data", false, "Cleanup data")
 	cmd.Flags().BoolVar(&cleanOpt.CleanupLog, "log", false, "Cleanup log")
 	cmd.Flags().BoolVar(&cleanALl, "all", false, "Cleanup both log and data")
--- a/pkg/cluster/manager.go
+++ b/pkg/cluster/manager.go
@@ -307,11 +307,12 @@ func (m *Manager) CleanCluster(clusterName string, gOpt operator.Options, cleanO
 			target = "log"
 		}
 		if err := cliutil.PromptForConfirmOrAbortError(
-			"This operation will clean %s %s cluster %s's %s.\nNodes will be ignored: %s\nRoles will be ignored: %s\nDo you want to continue? [y/N]:",
+			"This operation will clean %s %s cluster %s's %s.\nNodes will be included:%s\nNodes will be ignored: %s\nRoles will be ignored: %s\nDo you want to continue? [y/N]:",
 			m.sysName,
 			color.HiYellowString(base.Version),
 			color.HiYellowString(clusterName),
 			target,
+			cleanOpt.CleanDataNodes,
 			cleanOpt.RetainDataNodes,
 			cleanOpt.RetainDataRoles); err != nil {
 			return err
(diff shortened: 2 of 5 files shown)


components/cluster/command/clean.go
```diff
@@ -56,6 +58,7 @@ You can retain some nodes and roles data when cleanup the cluster, eg:
 
 	cmd.Flags().StringArrayVar(&cleanOpt.RetainDataNodes, "ignore-node", nil, "Specify the nodes or hosts whose data will be retained")
 	cmd.Flags().StringArrayVar(&cleanOpt.RetainDataRoles, "ignore-role", nil, "Specify the roles whose data will be retained")
+	cmd.Flags().StringArrayVarP(&cleanOpt.CleanDataNodes, "node", "n", nil, "Specify the nodes whose log will be clean")
```
9547, line 61: 1. Why we need this argument to be persistent? 
2. s/Specify the nodes whose log will be clean/Specify the nodes or hosts whose data will be cleaned/

9547: One more thing, compared to `--ignore-node` and `--ignore-role`, correspondingly, we also need `--role` parameter.

pkg/cluster/manager.go
```diff
@@ -307,11 +307,12 @@ func (m *Manager) CleanCluster(clusterName string, gOpt operator.Options, cleanO
 			target = "log"
 		}
 		if err := cliutil.PromptForConfirmOrAbortError(
-			"This operation will clean %s %s cluster %s's %s.\nNodes will be ignored: %s\nRoles will be ignored: %s\nDo you want to continue? [y/N]:",
+			"This operation will clean %s %s cluster %s's %s.\nNodes will be included:%s\nNodes will be ignored: %s\nRoles will be ignored: %s\nDo you want to continue? [y/N]:",
```
9547, line 310: IMO, if one specified `-n`, then ONLY those nodes will be cleaned, this is the opposite condition to `--ignore-node`,  so this prompt message should be an independent one, such as:

```bash
This operation will clean the following nodes from cluster xxx:
172.10.1.0:4000
172.10.1.1:4000
Do you want to continue? [y/N]:"
```

pkg/cluster/operation/operation.go
```diff
@@ -34,8 +34,9 @@ type Options struct {
 	SSHType           executor.SSHType // the ssh type: 'builtin', 'system', 'none'
 
 	// What type of things should we cleanup in clean command
-	CleanupData bool // should we cleanup data
-	CleanupLog  bool // should we clenaup log
+	CleanupData    bool     // should we cleanup data
+	CleanupLog     bool     // should we clenaup log
+	CleanDataNodes []string // clean data on specific nodes.
```
9547, line 39: Please put this field after `RetainDataNodes []string`

pkg/cluster/operation/destroy.go
```diff
@@ -237,10 +237,20 @@ func DestroyMonitored(getter ExecutorGetter, inst spec.Instance, options *spec.M
 func CleanupComponent(getter ExecutorGetter, instances []spec.Instance, cls spec.Topology, options Options) error {
 	retainDataRoles := set.NewStringSet(options.RetainDataRoles...)
 	retainDataNodes := set.NewStringSet(options.RetainDataNodes...)
+	cleanDataNodes := set.NewStringSet(options.CleanDataNodes...)
 
 	for _, ins := range instances {
+		dataRetained := false
```
9547, line 243: Could you please move those lines into a small function, such as: 

```go
checkInstanceRetained(instance spec.Instance, options Options) bool 
```

then we'll add more unit tests on this function.

9547, line 243: this declare is not used.

tests/tiup-cluster/script/cmd_subtest.sh
```diff
@@ -94,6 +94,12 @@ function cmd_subtest() {
     tiup-cluster $client exec $name -R tidb --command="systemctl status tidb-4000|grep 'enabled;'"
     tiup-cluster $client exec $name -R pd --command="systemctl status pd-2379|grep 'enabled;'"
 
+    tiup-cluster $client --yes clean $name --log -n $ipprefix.102:9090
+
+    echo "checking clean specific node log"
+    tiup-cluster $client exec $name $ipprefix.101 --command "ls /home/tidb/deploy/prometheus-9090/log/prometheus.log"
```
9547, line 100: Could you please add more test cases, such as:

- [ ] --node x --ignore-node x:4000
- [ ] --node x:4000 --ignore-node x 
...

components/cluster/command/clean.go
lucklove: Prefer to keep consistent with existed code:https://github.com/pingcap/tiup/blob/master/components/cluster/command/start.go#L49

```
	cmd.Flags().StringSliceVarP(&gOpt.Roles, "role", "R", nil, "Only clean specified roles")
	cmd.Flags().StringSliceVarP(&gOpt.Nodes, "node", "N", nil, "Only clean specified nodes")
```

pkg/cluster/manager.go
lucklove: @anywhy @9547 I think we can refactor this message as:

```
This operation will clean the following directories from cluster xxx:
172.10.1.0:/path/to/dir1
172.10.1.0:/path/to/dir2
172.10.1.1:/path/to/dir
```

No matter the user specify `-N` or `--ignore-node`, we can use the same message.

This will make the clean action more transparent.

pkg/cluster/operation/operation.go
lucklove: We don't need this field, in fact. Just use `.Roles` and `.Nodes`

pkg/cluster/manager.go
AstroProfundis: I think this could be a separate PR, otherwise this one would be too large

pkg/cluster/operation/destroy.go
```diff
@@ -237,10 +237,20 @@ func DestroyMonitored(getter ExecutorGetter, inst spec.Instance, options *spec.M
 func CleanupComponent(getter ExecutorGetter, instances []spec.Instance, cls spec.Topology, options Options) error {
 	retainDataRoles := set.NewStringSet(options.RetainDataRoles...)
 	retainDataNodes := set.NewStringSet(options.RetainDataNodes...)
+	cleanDataNodes := set.NewStringSet(options.CleanDataNodes...)
```
AstroProfundis, line 240: What's the purpose? `destroy` is intended to delete *all nodes* of a cluster, it make no sense to "clean data and configs for one node", that's what `scale-in` would do.

Issue #892, opened 2020-11-09
Utilize TiKV --advertise-status-addr


breeswish:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Now `--advertise-status-addr` is not set, so that TiKV reports a wrong advertise address (which comes from `--status-addr`) when `--status-addr` is set to a wildcard address like `0.0.0.0`.

**Describe the feature you'd like:**

For [TiKV >= 4.0.1](https://github.com/tikv/tikv/commit/511c346a4654f166845ec23fdffc70a94885d360), add `--advertise-status-addr` parameter.

**Describe alternatives you've considered:**

As an alternative, TiKV can infer a correct advertise status address from advertise address. However I have no idea why it is not considered in the [original proposal](https://github.com/tikv/tikv/issues/7920#issuecomment-638579276).

**Teachability, Documentation, Adoption, Migration Strategy:**

lucklove:
Related change: https://github.com/tikv/tikv/issues/7920

PR #951, opened 2020-11-30
Utilize TiKV --advertise-status-addr

lucklove:
Fix https://github.com/pingcap/tiup/issues/892

When init script for TiKV, check if current cluster version >= v4.0.1, if so, use `--advertise-status-addr`

### Check List 

Tests 

 - Integration test

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```


lucklove:
@breeswish PTAL

breeswish (approved):

PR #953, opened 2020-12-01
Fix the dependencies field to match the specification

lucklove:
As https://github.com/pingcap/tiup/blame/master/doc/design/manifest.md#L256 specified, the dependencies filed should be `map[string]string`. But in the current implementation, it's `[]string`, fortunately, this field is not used up until now and it's aways a `null` in the manifest json file, which adopted both `[]string` and `map[string]string`. But in the future, we may use this field, so it's better to correct it as early as possible.

- change `[]string` to `map[string]string`
- upgrade the spec version from `v0.1.0` to `v0.1.1`

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/pkg/repository/v1manifest/manifest.go
+++ b/pkg/repository/v1manifest/manifest.go
@@ -44,7 +44,7 @@ const (
 	ManifestFilenameTimestamp = "timestamp.json"
 
 	// SpecVersion of current, maybe we could expand it later
-	CurrentSpecVersion = "0.1.1"
+	CurrentSpecVersion = "0.1.0"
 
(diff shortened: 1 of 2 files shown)


9547 (approved):
/lgtm

lucklove:
@9547 I found a bug ):

lucklove:
We can't upgrade spec_version at this time, until https://github.com/pingcap/tiup/issues/960 resolved.

PR #958, opened 2020-12-01
Fix/scalein pump monitor notreload

### LINKED ISSUE #936: scale-in pump/drainer action，prometheus do not update target
## Bug Report

1. What did you do?

使用 tiup cluster scale-in tc1 -N  172.16.4.36:18247 / 172.16.4.36:18259 命令缩容 drainer/pump 节点

  172.16.4.36:18247   drainer       172.16.4.36   18247
  172.16.4.36:18259   pump          172.16.4.36   18259

-----

缩容操作成功后，prometheus 配置文件中未关联更新 drainer/pump target 信息

tiup --version
v1.2.3 tiup
Go Version: go1.13
Git Branch: release-1.2
GitHash: df7e28a

9547:
Fix #936 

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/pkg/cluster/manager.go
+++ b/pkg/cluster/manager.go
@@ -1459,7 +1459,6 @@ func (m *Manager) ScaleIn(
 			m.specManager.Path(clusterName, "ssh", "id_rsa.pub")).
 		ClusterSSH(topo, base.User, sshTimeout, sshType, metadata.GetTopology().BaseTopo().GlobalOptions.SSHType)
 
-	// TODO: support command scale in operation.
 	scale(b, metadata, tlsCfg)
 
 	t := b.Parallel(force, regenConfigTasks...).Parallel(force, buildDynReloadProm(metadata.GetTopology())...).Build()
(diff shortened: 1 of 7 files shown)


pkg/cluster/manager.go
```diff
@@ -1583,6 +1580,126 @@ func (m *Manager) ScaleOut(
...
+			log.Infof("Start destroy Tombstone nodes: %v ...", nodes)
+			return err
+		}).
+		ClusterOperate(cluster, operator.DestroyTombstoneOperation, gOpt, tlsCfg).
+		UpdateMeta(clusterName, clusterMeta, nodes).
+		UpdateTopology(clusterName, m.specManager.Path(clusterName), clusterMeta, nodes)
+
+	var regenConfigTasks []task.Task
+	deletedNodes := set.NewStringSet(nodes...)
+	topo.IterInstance(func(instance spec.Instance) {
```
lucklove, line 1644: Can we add a function such as `RefreshConfig` to build the `regenConfigTasks`? I found that there is a lot of redundant code: https://github.com/pingcap/tiup/pull/958/files#diff-5609f7b865f14ce911938ab0a8f21caa07446244d9c106a5df27ba9c08cf80e6R1386

Notice: there are many other places use the same logic, we can replace that logic with just a function call: 
- https://github.com/pingcap/tiup/blob/master/pkg/cluster/manager.go#L702
- https://github.com/pingcap/tiup/blob/master/pkg/cluster/manager.go#L2073

9547: Good idea

9547: As this PR contains 200+ lines modified, IMO, I'll trigger another PR to do the refactor work. BTW the file `manager.go`  is too large(~ 3000 loc),  I want to split this file into some small files If you agree with it. :)

lucklove: It's very nice if you can split the `manager.go`, I want to do that days ago.

I think we can add a package (eg. `tiup/pkg/cluster/manager`) which contains several .go files, each of them has different function (eg. split by `Deploy`, `Start`, 'ScaleOut` .etc.)

Issue #798, opened 2020-09-21
Support THP in `tiup cluster check`


youjiali1995:
## Feature Request

**Is your feature request related to a problem? Please describe:**

`THP` can result in high memory usage in the TiDB cluster. We recommend to disable it. https://github.com/pingcap/tidb-ansible/issues/1078

**Describe the feature you'd like:**

Support THP in `tiup cluster check` and disable it when applies the check.

**Describe alternatives you've considered:**

No

**Teachability, Documentation, Adoption, Migration Strategy:**

PR #964, opened 2020-12-03
Support THP in tiup cluster check

anywhy:
try implement #798 

### Check List 

Tests 
 - Integration test

Release notes:

```release-note
NONE
```


pkg/cluster/operation/check.go
```diff
@@ -58,6 +58,7 @@ var (
 	CheckNameSELinux     = "selinux"
 	CheckNameCommand     = "command"
 	CheckNameFio         = "fio"
+	CheckNameTHP         = "THP"
```
9547, line 61: To keep up with the previous code style, use lowercase

pkg/cluster/task/check.go
```diff
@@ -33,6 +33,7 @@ var (
 	CheckTypePackage      = "package"
 	CheckTypePartitions   = "partitions"
 	CheckTypeFIO          = "fio"
+	CheckTypeTHP          = "THP"
```
9547, line 36: IMO, THP should be part of `CheckTypeSystemConfig`

pkg/cluster/operation/check.go
```diff
@@ -651,3 +652,20 @@ func CheckFIOResult(rr, rw, lat []byte) []*CheckResult {
 
 	return results
 }
+
+// CheckTHPEnabled checks THP in /sys/kernel/mm/transparent_hugepage/enabled,/sys/kernel/mm/transparent_hugepage/defrag
+func CheckTHPEnabled(opt *CheckOptions, l []byte) []*CheckResult {
```
9547, line 657: s/CheckTHPEnabled/CheckTHP/

pkg/cluster/operation/check.go
```diff
@@ -651,3 +652,20 @@ func CheckFIOResult(rr, rw, lat []byte) []*CheckResult {
...
+	var results []*CheckResult
+
+	for _, line := range strings.Split(string(l), "\n") {
+		if !strings.Contains(line, "[never]") {
```
9547, line 661: IMO, if THP is disabled, we should also add a `Msg` result

tests/tiup-cluster/script/cmd_subtest.sh
```diff
@@ -31,6 +31,7 @@ function cmd_subtest() {
     echo $check_result | grep "os-version"
     echo $check_result | grep "selinux"
     echo $check_result | grep "service"
+    echo $check_result | grep "THP"
```
9547, line 34: ditto

components/cluster/command/check.go
```diff
@@ -261,6 +261,24 @@ func checkSystemInfo(s *cliutil.SSHConnectionProps, topo *spec.Specification, op
...
+				).
+				// check for THP
+				Shell(
+					inst.GetHost(),
+					"cat /sys/kernel/mm/transparent_hugepage/enabled",
+					false,
+				).
+				Shell(
+					inst.GetHost(),
+					"cat /sys/kernel/mm/transparent_hugepage/defrag",
```
9547, line 273: 1. Maybe you can merge the two command into one:

```bash
cat /sys/kernel/mm/transparent_hugepage/{enabled,defrag}
```

2. IMO, maybe you can implement the THP Check like `CheckSELinux()`

pkg/cluster/operation/check.go
anywhy: thx, got it

components/cluster/command/check.go
```diff
@@ -442,6 +442,11 @@ func fixFailedChecks(ctx *task.Context, host string, res *operator.CheckResult,
 			),
 			true)
 		msg = fmt.Sprintf("will try to %s, reboot might be needed", color.HiBlueString("disable SELinux"))
+	case operator.CheckNameTHP:
+		t.Shell(host,
+			"echo never > /sys/kernel/mm/transparent_hugepage/enabled && echo never > /sys/kernel/mm/transparent_hugepage/defrag",
```
9547, line 447: This will disable THP in runtime, if the system rebooted, the THP's status maybe enabled again? Should we disable it permanently? ref https://access.redhat.com/solutions/46111

lucklove: Since the origin issue is reference to tidb-ansible, maybe just keep the same behavior as [tidb-ansible](https://github.com/pingcap/tidb-ansible/pull/1079) is enough

AstroProfundis: It seems this config does not have an entry in `sysctl`, so the only way to make it consist after reboot would be adding a kernel parameter at boot. ref https://www.kernel.org/doc/html/v5.4/admin-guide/mm/transhuge.html#boot-parameter

lucklove: It's  unpractical to add a kernel parameter, just remind the user to run `tiup cluster check` again after reboot.

9547 (approved):
/lgtm

lucklove:
> /lgtm

The bot will ignore the `lgtm` in this case.

PR #966, opened 2020-12-03
Refine the error message of monitor port conflict

### LINKED ISSUE #965: The error message of monitor port conflict is misleading
## Bug Report

1. What did you do?

I try to deploy a cluster that has a monitor port conflict with a previous cluster. TiUP tells me that there exists a port conflict between pd and tiflash:
```
>  tiup cluster deploy j2 v4.0.8 topo.yaml 
Error: Deploy port conflicts to an existing cluster (spec.deploy.port_conflict)

The port you specified in the topology file is:
  Port:      9110
  Component: pd 172.16.5.81

It conflicts to a port in the existing cluster:
  Existing Cluster Name: xxx
  Existing Port:         9110
  Existing Component:    tiflash 172.16.5.81

Please change to use another port or another host.
```
But the port 9110 is the port of monitored.node_exporter_port

TiUP tells me that there exists a port conflict of monitor port

3. What did you see instead?

TiUP tells me that there exists a port conflict between pd and tiflash

v1.2.5 tiup
Go Version: go1.13
Git Branch: release-1.2
GitHash: 5b9cf5d

JaySon-Huang:

Fix #965. Refine the error message of monitor port conflict.

Use `spec.RoleMonitor` as the component name instead of the instance name.

### Check List 

Tests 

 - Unit test

Code changes

 - N/A

Side effects

 - N/A

Related changes

Release notes:

```release-note
Refine the error message of monitor port conflict
```


pkg/cluster/spec/validate.go
```diff
@@ -222,23 +223,26 @@ func CheckClusterPortConflict(clusterList map[string]Metadata, clusterName strin
 			blackboxExporterPort := mOpt.BlackboxExporterPort
 			for _, port := range inst.UsedPorts() {
 				existingEntries = append(existingEntries, Entry{
-					clusterName: name,
-					instance:    inst,
-					port:        port,
+					clusterName:   name,
+					componentName: inst.ComponentName(),
+					host:          inst.GetHost(),
```
AstroProfundis, line 228: Um, I didn't quite get the point of adding a new field, we could use `inst.GetHost()` to get the value anytime.

pkg/cluster/spec/validate_test.go
```diff
@@ -428,12 +428,12 @@ tidb_servers:
 	c.Assert(ok, IsTrue)
 	c.Assert(suggestion, Equals, `The port you specified in the topology file is:
   Port:      9100
-  Component: tidb 172.16.5.138
+  Component: monitor 172.16.5.138
```
AstroProfundis, line 431: Did you tested if the conflict is between a tidb node and a pd node but not with their monitoring agent's ports?

pkg/cluster/spec/validate.go
JaySon-Huang: I try to make the message contains in `Entry` is clear and eliminate the ambiguous. If there exist fields `port`, `componentName` along with `inst`, should I get the port from `inst.GetPort()` or `inst.ComponentName()`?
However, keeping `inst` instead of adding `host` is more flexible.

I can keep `inst` instead of adding `host` if you insist.

pkg/cluster/spec/validate_test.go
JaySon-Huang: The 9100 is the default port of the monitoring agent.
And I debug the code step by step to know that the conflict is monitoring agent after seeing the misleading error message https://github.com/pingcap/tiup/issues/965#issuecomment-738515276

For the test case that conflicts between tidb node and pd node but not their monitoring agent's, there is a test case between pump_server and pd in the following lines.

pkg/cluster/spec/validate.go
AstroProfundis: Yes I would still suggest using `inst` as it carries more information.

AstroProfundis (approved):

PR #967, opened 2020-12-03
Support rotate root manifests

lucklove:
In TiUP's repository, there is a [root.json](https://tiup-mirrors.pingcap.com/root.json)(the root manifest) file, which records the keys we used for the whole repository. It's the foundation of the trust chain. Related document of the manifest mechanism: https://github.com/pingcap/tiup/blob/master/doc/design/manifest.md

The root.json is signed by 5 different people (with 5 different keys) and it has an `expires` field, so the 5 people must re-sign the root.json before it really `expires`. We should do the follow things:

- Update the expires and version field in root.json
- The 5 people who has root key signs the modified root.json
- Update snapshot.json to record the new version of root.json
- Re-sign snapshot.json because it's content has been changed
- Update timestamp.json since the snapshot.json has been changed
- Re-sign timestamp.json since it's content has been changed

Do these things manually will be boring and it's easy to make mistake (what if some one is publishing component to the repository while you modify timestamp.json and snapshot.json manually?). So we'd better make it automated, at least semi-automated.

- Enhance the `tiup mirror sign` command, make it possible to sign the manifest via network by `tiup mirror sign http:/xxx/yyy.json`, in a similar way as signing a manifest file in local filesystem.
- Implement a temp server, it will open an editor to let you edit root.json and start to serve the edited root.json via the internet
- Implement a `Rotate` method on `Mirror` interface, which will update snapshot.json and timestamp.json automatically according to the updated root.json

By this way, we can update the root.json with these steps:

- The one who response to update root.json execute the command `tiup mirror rotate` and edit the root.json (update expires field .etc.), after saving the file, a server will be started, and the address to sign the manifest will be printed on the console.
- The 5 people who have the key use the command `tiup mirror sign http://rotate-address` to sign the modified root.json
- After all people signed, the temp-server will close automatically and send the signed root.json to the real tiup-server (if you use remote mirror) or just push it to local mirror (if you use offline mirror)
- The mirror will check the permission and update snapshot and timestamp

By this way, we can do things easier and avoid transfering private key via the internet (we should never do this)

### Check List 

Tests 

The manual test steps:

In terminal 1:

- `tiup mirror init /tmp/test-mirror`
- `tiup mirror set /tmp/test-mirror`
- `tiup mirror rotate`: in the eidtor, increase the manifest version by one and save, this should start a server

In terminal 2:

- `tiup mirror sign http://<my-ip>:8080/rotate/root.json --key /tmp/test-mirror/keys/<hash1>-root.json`
- `tiup mirror sign http://<my-ip>:8080/rotate/root.json --key /tmp/test-mirror/keys/<hash2>-root.json`
- `tiup mirror sign http://<my-ip>:8080/rotate/root.json --key /tmp/test-mirror/keys/<hash3>-root.json`

After these steps, the server of terminal 1 will closed automatically and the `/tmp/test-mirror/2.root.json` should be produced.

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
Support rotating the root manifest
```


cmd/mirror.go
```diff
@@ -81,12 +90,41 @@ func newMirrorSignCmd() *cobra.Command {
 				return cmd.Help()
 			}
 
-			if len(args) == 1 {
-				return v1manifest.SignManifestFile(args[0], env.Profile().Path(localdata.KeyInfoParentDir, "private.json"))
+			if privPath == "" {
+				privPath = env.Profile().Path(localdata.KeyInfoParentDir, "private.json")
```
AstroProfundis, line 94: `private.json` has been used multiple times in the package, I think we should make it a constant value.

cmd/mirror.go
```diff
@@ -237,6 +275,71 @@ func newMirrorModifyCmd() *cobra.Command {
...
+
+			manifest, err := rotate.Serve(addr, root)
+			if err != nil {
+				return err
+			}
+
+			return environment.GlobalEnv().V1Repository().Mirror().Rotate(manifest)
+		},
+	}
+	cmd.Flags().StringVarP(&addr, "addr", "", addr, "listen address of rotate server")
```
AstroProfundis, line 300:
```suggestion
	cmd.Flags().StringVarP(&addr, "addr", "", addr, "listen address:port when starting the temp server for rotating")
```

pkg/repository/mirror.go
```diff
@@ -341,6 +356,39 @@ func (l *httpMirror) Grant(id, name string, key *v1manifest.KeyInfo) error {
...
+	defer resp.Body.Close()
+
+	if resp.StatusCode < 300 {
+		return nil
+	}
+	switch resp.StatusCode {
+	case http.StatusConflict:
+		return errors.Errorf("The manifest has been modified after you fetch it, please try again")
+	case http.StatusBadRequest:
+		return errors.Errorf("The server refused the manifest, please check if it's a valid root manifest")
```
AstroProfundis, line 381:
```suggestion
		return errors.Errorf("The server rejected the manifest, please check if it's a valid root manifest")
```

pkg/repository/model/model.go
```diff
@@ -300,9 +353,21 @@ func (m *model) readSnapshotManifest() (*v1manifest.Manifest, error) {
 	return m.txn.ReadManifest(v1manifest.ManifestFilenameSnapshot, &v1manifest.Snapshot{})
 }
 
-// readRootManifest returns root.json
+// readRootManifest returns newest root.json
```
AstroProfundis, line 356:
```suggestion
// readRootManifest returns the latest root.json
```

pkg/repository/model/model.go
```diff
@@ -300,9 +353,21 @@ func (m *model) readSnapshotManifest() (*v1manifest.Manifest, error) {
...
 func (m *model) readRootManifest() (*v1manifest.Manifest, error) {
-	return m.txn.ReadManifest(v1manifest.ManifestFilenameRoot, &v1manifest.Root{})
+	root, err := m.txn.ReadManifest(v1manifest.ManifestFilenameRoot, &v1manifest.Root{})
+	if err != nil {
+		return root, err
+	}
+
+	for {
+		file := fmt.Sprintf("%d.%s", root.Signed.Base().Version+1, v1manifest.ManifestFilenameRoot)
```
AstroProfundis, line 364: Use the `Filename()` function instead.

server/rotate/rotate_server.go
```diff
@@ -0,0 +1,123 @@
...
+	manifest := &v1manifest.Manifest{Signed: root}
+	status := newStatusRender(manifest, addr)
+	defer status.stop()
+
+SIGLOOP:
+	for sig := range sigCh {
+		k := root.Roles[v1manifest.ManifestTypeRoot].Keys[sig.KeyID]
+		if k == nil {
+			// Received a signature signed by an invalid key
+			continue
```
AstroProfundis, line 99: Is there any message returned to client to let them know they used a wrong key?

lucklove: Added

AstroProfundis (approved):

PR #968, opened 2020-12-03
Playground: support setting custom timeout for waiting instances up

### LINKED ISSUE #962: Playground incorrectly detected that tidb failed to start
## Bug Report

1. What did you do?

```
tiup playground --host 0.0.0.0 --tiflash 1
```

outputs
```
To connect TiDB: mysql --host %s --port %s -u root
```
and start tiflash

3. What did you see instead?

tidb start successfully but outputs

```
Playground Bootstrapping...
Start pd instance
Start tikv instance
Start tidb instance
Waiting for tidb localhost:4000 ready  ... Error
To view the dashboard: http://localhost:2379/dashboard
To view the Prometheus: http://0.0.0.0:9090
To view the Grafana: http://0.0.0.0:3000
```

```
v1.2.0 tiup
Go Version: go1.13
Git Branch: bump
GitHash: 7e35be8
```

--- 

I use tiup in Ubuntu 18 for WSL. And this happens occasionally.

unbyte:

#962 

Currently tiup playground will wait for tidb up in 60 seconds, and if it fails, tiflash instance will be ignored. However, sometimes tidb instances start normally but only the startup time exceeds the timeout time. So maybe support setting custom timeout for waiting instances up can help.

1. add two flag
```
--db.timeout int           TiDB max wait time in seconds for staring, 0 means no limit (default 60)
--tiflash.timeout int      TiFlash max wait time in seconds for staring, 0 means no limit (default 120)
```

2. remove waiting for tikv

3. check tidb and tiflash status in parallel

image

### Check List 

Tests 

Related changes

Release notes:

```release-note
NONE
```

--- 

close #962


unbyte:
image

Noticed that even if a process such as tiflash exited with error, the function to check status is still polling (this problem exists in both the master branch and this PR). Maybe need to find a way to notify check functions when the process exits with error?

components/playground/playground.go
```diff
@@ -785,68 +786,82 @@ func (p *Playground) bootCluster(ctx context.Context, env *environment.Environme
...
-		startTiFlash := func() error {
+		var started []*instance.TiFlashInstance
+		for _, flash := range p.tiflashs {
+			if err := p.startInstance(ctx, flash); err != nil {
+				fmt.Println(color.RedString("TiFlash %s failed to start: %s", flash.Addr(), err))
+			} else {
+				started = append(started, flash)
+			}
+		}
+		p.tiflashs = started
```
unbyte, line 827: ignore tiflash instances that fail to start like how master branch works. 
however, I think it's better to exit and throw error when fail to start tiflash just like other components.

lucklove: I think it's because TiFlash failed to start is not a critical issue, the cluster can still bootstrap. So it's better to ignore it and just print out a log to let the user know what happened.

unbyte: yeah it's reasonable. But IMO it is based on the fact that the number of tiflash instances defaults to 1 rather than 0 while some user actually don't intend to use tiflash. 
BTW, some days ago, I've found that both en and zh documentations say the number of tiflash instances defaults to 0. I wonder if it's neccessary to set the default number to 1.

lucklove: I think we should update the document. In early releases, it's 0. But we changed it to 1 in recent releases.

components/playground/main.go
```diff
@@ -246,47 +250,43 @@ func tryConnect(dsn string) error {
...
 			return true
 		}
+		time.Sleep(time.Second)
 	}
-	return false
 }
 
-func checkStoreStatus(pdClient *api.PDClient, typ, storeAddr string) error {
-	prefix := color.YellowString("Waiting for %s %s ready ", typ, storeAddr)
-	bar := progress.NewSingleBar(prefix)
```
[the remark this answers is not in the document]
unbyte: moved to bootCluster()

components/playground/playground.go
```diff
@@ -785,68 +786,82 @@ func (p *Playground) bootCluster(ctx context.Context, env *environment.Environme
...
+		var wg sync.WaitGroup
+		bars := progress.NewMultiBar(color.YellowString("Waiting for tidb instances ready\n"))
+		for _, db := range p.tidbs {
+			wg.Add(1)
+			prefix := color.YellowString(db.Addr())
+			bar := bars.AddBar(prefix)
+			go func(dbInst *instance.TiDBInstance) {
+				defer wg.Done()
+				if s := checkDB(dbInst.Addr(), options.tidb.UpTimeout); s {
+					succ = append(succ, dbInst.Addr())
```
unbyte, line 799: may lose some successfully started instances, because slice is not thread safe. it's better to use channel so.

lucklove: Oh, yes. It's better to fix that...

unbyte:
solved

PR #970, opened 2020-12-03
typo(*): TiOps -> TiUp

9547:

s/TiOps/TiUP/g

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```


unbyte:
image

it seems there are still a lot of typos

AstroProfundis:
The `cluster` component was named as `tiops` before the initial GA release so...

lucklove:
Can you use `TiUP` instead of `TiUp`? And some other places is still using `TiUp`, I think it should be `TiUP` because we have use it for a long time in the community....

9547:
> Can you use `TiUP` instead of `TiUp`? And some other places is still using `TiUp`, I think it should be `TiUP` because we have use it for a long time in the community....


1. tiops -> tiup
2. Ti[UuPp] -> TiUP

PTAL

PR #979, opened 2020-12-09
Feature/split cluster into small files

9547:

ref https://github.com/pingcap/tiup/pull/958#discussion_r538976861

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```


components/cluster/command/deploy.go
```diff
@@ -77,17 +77,7 @@ func newDeploy() *cobra.Command {
...
-				version,
-				topoFile,
-				opt,
-				postDeployHook,
-				skipConfirm,
-				gOpt.OptTimeout,
-				gOpt.SSHTimeout,
-				gOpt.SSHType,
-			)
+			return cm.Deploy(clusterName, version, topoFile, opt, postDeployHook, skipConfirm, gOpt)
```
lucklove, line 80: It's better to keep the default manager defined in `manager` package:

```
var clusterManager *clusterManager

func Deploy(...) {
    return clusterManager.Deploy(...)
}
```

9547: I'm sorry, but I'm curious about the advantage of this design pattern.  We used a singleton instance to handle all `Manager` related works, the former declares the instance in user-space, while the latter in the package-space. But the later one seems duplicated, do we need to declare a one-one proxy method for the public methods?

Issue #947, opened 2020-11-29
Make sure `log_dir` is not the same with `data_dir` for TiKV cluster


shenli:
## Feature Request

**Is your feature request related to a problem? Please describe:**

If the `log_dir` is set as the same dir with `data_dir`, `clean --log` will remove the data in `data_dir`

**Describe the feature you'd like:**

Add a pre-flight check to make sure the two dirs are not the same one.

**Describe alternatives you've considered:**

None

**Teachability, Documentation, Adoption, Migration Strategy:**

None

3pointer:
I think the directory is not the key problem. and we should not delete directory directly. if user choose to cleanup log. we need to make sure all files to be deleted have a suffix `.log`.

AstroProfundis:
There are several improves we would do:

 - [ ] Not allowing deployment of cluster if the log_dir is the same with data_dir or deploy_dir
 - [ ] Not deleting anything if the log_dir is the same as data_dir on `clean` command
 - [ ] Make experimental features with the “experimental” mark more visible
 - [ ] Display list of files / directories to be deleted before the confirmation prompt
 - [ ] Only delete files with .log extension on `clean --log`

PR #981, opened 2020-12-10
cluster: refine log_dir handling on `clean` (part 1)

AstroProfundis:
Part of #947 

 - Mark experimental features more visible
 - Check for paths to avoid deleting files if data dir is a sub path of log dir
 - Only delete `*.log` when deleting log files

### Check List 

Tests 

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
cluster: adjust checking process when cleaning log files
```

### DIFF
--- a/pkg/cluster/operation/destroy.go
+++ b/pkg/cluster/operation/destroy.go
@@ -308,18 +308,22 @@ func CleanupComponent(getter ExecutorGetter, instances []spec.Instance, cls spec
 		log.Infof("Cleanup instance %s", ins.GetHost())
 
 		delFiles := set.NewStringSet()
+		dataPaths := set.NewStringSet()
+		logPaths := set.NewStringSet()
 
 		if options.CleanupData && len(ins.DataDir()) > 0 {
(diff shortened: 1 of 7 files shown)


pkg/cluster/operation/destroy.go
```diff
@@ -308,19 +308,33 @@ func CleanupComponent(getter ExecutorGetter, instances []spec.Instance, cls spec
...
 			for _, logDir := range strings.Split(ins.LogDir(), ",") {
-				delFiles.Insert(path.Join(logDir, "*"))
+				logPaths.Insert(path.Join(logDir, "*.log"))
 			}
 		}
 
+		for _, lp := range logPaths.Slice() {
+			for _, dp := range dataPaths.Slice() {
+				// remove path of log dir from list if any data dir is under it
+				if strings.HasPrefix(dp, lp) {
```
lucklove, line 329: I don't think this will work because the logPaths is like:

```
/deploy/log/*.log
```

And the data dir could be:

```
/deploy/log/data
```

In this case, the data dir is under log dir, but it's not detected.

Issue #642, opened 2020-07-29
部署 tiflash 缺少 numactl 出现 log 目录为空


zhouyueyue1115:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?

2. What did you expect to see?

3. What did you see instead?

4. What version of TiUP are you using (`tiup --version`)?

lucklove:
Cause:
```
#!/bin/bash
set -e

# WARNING: This file was auto-generated. Do not edit!
#          All your edit might be overwritten!
cd "{{.DeployDir}}" || exit 1

export RUST_BACKTRACE=1

export TZ=${TZ:-/etc/localtime}
export LD_LIBRARY_PATH={{.DeployDir}}/bin/tiflash:$LD_LIBRARY_PATH

echo -n 'sync ... '
stat=$(time sync)
echo ok
echo $stat

{{- if .NumaNode}}
exec numactl --cpunodebind={{.NumaNode}} --membind={{.NumaNode}}  \
{{- else}}
exec \
{{- end}}
    bin/tiflash/tiflash server --config-file conf/tiflash.toml
```

We should append something like:
```
2>> "{{.LogDir}}/tiflash_stderr.log"
```

PR #984, opened 2020-12-11
Fix tiflash start script

lucklove:
Fix https://github.com/pingcap/tiup/issues/642


9547 (approved):
BTW, `>> log/tiflash_stderr.log` will only append the stderr log into file, but issue #642 seems to say the `tiflash/logs` not created?

lucklove:
> BTW, `>> log/tiflash_stderr.log` will only append the stderr log into file, but issue #642 seems to say the `tiflash/logs` not created?

Good question. The log directory is empty is because the tiflash didn't start at all (it should be start by numactl but there is no numactl found), so there is no opportunity for it to create a log file.

9547:
> > BTW, `>> log/tiflash_stderr.log` will only append the stderr log into file, but issue #642 seems to say the `tiflash/logs` not created?
> 
> Good question. The log directory is empty is because the tiflash didn't start at all (it should be start by numactl but there is no numactl found), so there is no opportunity for it to create a log file.

I got it. I previously thought that issue want to say the log dir is missing, not the log files under log dir is missing :bear:

glkappe:
> > BTW, `>> log/tiflash_stderr.log` will only append the stderr log into file, but issue #642 seems to say the `tiflash/logs` not created?
> 
> Good question. The log directory is empty is because the tiflash didn't start at all (it should be start by numactl but there is no numactl found), so there is no opportunity for it to create a log file.

So in files changed, does the final execution of the command change?
Newly added command line parameter `2>> "{{.LogDir}}/tiflash_stderr.log`, the output content should be printed in cmd.
So we can still locate the cause of the tiflash startup failure before `v1.3.0` ?

9547:
this behavior is the same as other components, we use systemctl to start/stop those components, if start failed,the error message was omited, and you can check the error log for more detail.

PR #987, opened 2020-12-13
Feature/remove not used file

9547:

[pkg/cluster/template/config/dm/prometheus.go](pkg/cluster/template/config/dm/prometheus.go) is not used anymore, remove it

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```


pkg/cluster/flags/debug.go
```diff
@@ -24,7 +24,6 @@ var (
 )
 
 func init() {
-	if s := os.Getenv("TIUP_CLUSTER_DEBUG"); s == "enable" {
-		DebugMode = true
-	}
+	ShowBacktrace = len(os.Getenv("TIUP_BACKTRACE")) > 0
```
lucklove, line 27: For the env which can be modify by user, better to define it in [localdata](https://github.com/pingcap/tiup/blob/master/pkg/localdata/constant.go#L26) and list it in [envList](https://github.com/pingcap/tiup/blob/master/cmd/env.go#L24)

pkg/cluster/flags/debug.go
```diff
@@ -24,7 +24,6 @@ var (
...
+	DebugMode = os.Getenv("TIUP_CLUSTER_DEBUG") == "enable"
```
lucklove, line 28: ditto

pkg/cluster/flags/debug.go
9547: I got it

9547: BTW, `ShowBacktrace` was not used, I'll remove it.

9547: one more thing, `tiup env` currently only displays the env with  `key=val`, should we extend it with `key=val description`, the description will tell users how to set the env, or the usage of this env?

lucklove: Maybe the  current style (same with go env) is more friendly for script?

PR #989, opened 2020-12-13
Feature: don't reload prometheus if the scaled comp is itself

### LINKED ISSUE #944: Don't need to reload Prometheus if the scale-in component is Prometheus
## Bug Report

1. What did you do?

```bash
tiup cluster scale-in xxx 172.19.0.101:9090
```

The Prometheus node scale-in and no error occured.

3. What did you see instead?

```bash
+ [ Serial ] - SystemCtl: host=172.19.0.101 action=reload prometheus-9090.service

Error: stdout: , stderr:Failed to reload prometheus-9090.service: Unit prometheus-9090.service not found.
: executor.ssh.execute_failed: Failed to execute command over SSH for 'tidb@172.19.0.101:22' {ssh_stderr: Failed to reload prometheus-9090.service: Unit prometheus-9090.service not found.
, ssh_stdout: , ssh_command: export LANG=C; PATH=$PATH:/usr/bin:/usr/sbin sudo -H bash -c "systemctl daemon-reload && systemctl reload prometheus-9090.service"}, cause: Process exited with status 5
```

master

9547:

fix #944 

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```


AstroProfundis (approved):

lucklove:
There are two test cases failed, both on `scale-out grafana` and waiting the instance number. I retried the test and it still failed, I think it's not a coincidence.

tests/tiup-cluster/topo/full_scale_in_prometheus.yaml
```diff
@@ -0,0 +1,3 @@
+monitoring_servers:
+  - host: n2
+    rule_dir: /tmp/local/prometheus
```
lucklove, line 3: I think this is the root cause that the test failed: you don't have this path on n2. So just delete this line, and the default rules will be used.

lucklove: And BTW, why we should scale-out a prometheus and scale-in? there is already a prometheus in the cluster. 

Maybe we can just find the existed (not in delete set) prometheus and reload them only.

tests/tiup-cluster/script/scale_tools.sh
```diff
@@ -70,6 +74,14 @@ function scale_tools() {
     topo=./topo/full_scale_in_grafana.yaml
     tiup-cluster $client --yes scale-out $name $topo
 
+    echo "start scale out prometheus"
+    topo=./topo/full_scale_in_prometheus.yaml
+    wait_instance_num_reach $name $total_add_one $native_ssh
```
lucklove, line 79: This line should be moved below the scale-out operation

9547: Sorry，fix it ASAP.

PR #1016, opened 2020-12-25
cluster: refine the process of getting latest store info from pd

AstroProfundis:
This is a following up of #1011 

Some more functions than checking store state are querying for store info of the latest / current store from PD that matching a specific host, so use a dedicated function to archive that purpose and implement the workaround of non-monotonic ID issue of PD there.

### Check List 

Tests 

Code changes

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/pkg/cluster/api/pdapi.go
+++ b/pkg/cluster/api/pdapi.go
@@ -21,7 +21,6 @@ import (
 	"io"
 	"net/http"
 	"net/url"
-	"sort"
 	"time"
 
 	"github.com/jeremywohl/flatten"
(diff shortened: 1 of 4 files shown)


pkg/cluster/api/pdapi.go
```diff
@@ -530,23 +545,19 @@ func (pc *PDClient) DelPD(name string, retryOpt *utils.RetryOption) error {
 
 func (pc *PDClient) isSameState(host string, state metapb.StoreState) (bool, error) {
 	// get info of current stores
-	stores, err := pc.GetStores()
+	storeInfo, err := pc.GetCurrentStore(host)
 	if err != nil {
 		return false, err
 	}
+	if storeInfo == nil {
```
lucklove, line 552: If storeInfo == nil, the err will not be nil. So I think this code is not necessary。

pkg/cluster/api/pdapi.go
```diff
@@ -568,25 +579,17 @@ var ErrStoreNotExists = errors.New("store not exists")
...
-	for _, storeInfo := range stores.Stores {
-		if storeInfo.Store.Address != host {
-			continue
-		}
-		storeID = storeInfo.Store.Id
-		break
-	}
-
-	if storeID == 0 {
+	if storeInfo == nil {
```
lucklove, line 586: ditto

pkg/cluster/api/pdapi.go
```diff
@@ -614,23 +617,24 @@ func (pc *PDClient) DelStore(host string, retryOpt *utils.RetryOption) error {
 		}
 	}
 	if err := utils.Retry(func() error {
-		currStores, err := pc.GetStores()
+		currStore, err := pc.GetCurrentStore(host)
```
lucklove, line 620: The err will not be nil if we don't found target store, and the Retry will retry the action and fail again

pkg/cluster/api/pdapi.go
```diff
@@ -614,23 +617,24 @@ func (pc *PDClient) DelStore(host string, retryOpt *utils.RetryOption) error {
...
 		if err != nil {
 			return err
 		}
+		if currStore == nil {
```
lucklove, line 624: If we should check if currStore ==nil, I think it's better not return err in GetCurrentStore(host) while not found store

pkg/cluster/spec/tikv.go
```diff
@@ -66,45 +64,15 @@ func checkStoreStatus(storeAddr string, tlsCfg *tls.Config, pdList ...string) st
...
-			}
-
-			if store.Store.Id > latestStore.Store.Id {
-				latestStore = store
-			}
-		}
-	}
-	if latestStore != nil {
-		return latestStore.Store.StateName
+	if store == nil {
```
lucklove, line 72: The store will not be nil. Maybe is NoStoreErr?

Issue #1026, opened 2020-12-29
support break off restart upgrade


King-Dylan:
## Feature Request
tiup support break off restart upgrade
**Is your feature request related to a problem? Please describe:**

In many cases, rolling update a large cluster  may be interrupted. Now if the tiup is  break off ,
components that have been upgraded before may be restarted again. Because we need a break off restart upgrade, we can ensure that the upgrade can be initiated after the upgrade is interrupted.
**Describe the feature you'd like:**

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

PR #1069, opened 2021-01-14
Implement checkpoint

lucklove:
Close https://github.com/pingcap/tiup/issues/1026

Implement a cache struct (`CheckPointExecutor`) overlay `Executor`, which writes the results of `Executor` to auth log and read them back when in checkpoint mode:

If user specify the `--checkpoint` flag，the `CheckPointExecutor` will read all checkpoints from the specified auth log and use them as the cache. After that, when there is a command should be execute in remote, it will check if it's exists in the cache, if so, use the cached result directly, otherwise, call real `Executor` and records it's result to current auth log.

### Check List 

Tests 

 - Unit test

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```


pkg/cluster/operation/action.go
```diff
@@ -595,13 +600,14 @@ func PrintClusterStatus(getter ExecutorGetter, cluster *spec.Specification) (hea
 		}
 
 		log.Infof("Checking service state of %s", com.Name())
-		errg, _ := errgroup.WithContext(context.Background())
+		errg, _ := errgroup.WithContext(ctx)
 		for _, ins := range com.Instances() {
 			ins := ins
 
+			nctx := checkpoint.NewContext(ctx)
```
AstroProfundis, line 607: Better add comment describe why a new context is needed.

pkg/cluster/operation/operation.go
```diff
@@ -29,6 +29,7 @@ type Options struct {
 	SSHTimeout        uint64           // timeout in seconds when connecting an SSH server
 	OptTimeout        uint64           // timeout in seconds for operations that support it, not to confuse with SSH timeout
 	APITimeout        uint64           // timeout in seconds for API operations that support it, like transferring store leader
+	CheckPoint        string           // the auth log id where should recover from, this is useful when an action failed and we want to continue that action
```
AstroProfundis, line 32:
```suggestion
	CheckPoint        string           // the audit log ID where we should recover from, this is useful when an action failed and we want to continue that action
```

pkg/cluster/operation/telemetry.go
```diff
@@ -68,11 +69,12 @@ func GetNodeInfo(
 		}
 		foundHosts[host] = struct{}{}
 
+		nctx := checkpoint.NewContext(ctx)
```
AstroProfundis, line 72: ditto

pkg/queue/any_queue.go
```diff
@@ -0,0 +1,41 @@
...
+	eq    func(a interface{}, b interface{}) bool
+	slice []interface{}
+}
+
+// NewAnyQueue builds a AnyQueue
+func NewAnyQueue(eq func(a interface{}, b interface{}) bool, aa ...interface{}) *AnyQueue {
+	return &AnyQueue{eq, aa}
+}
+
+// Get returns previews stored val that equals to val and remove it from queue, if not found, nil
```
AstroProfundis, line 27:
```suggestion
// Get returns previous stored value that equals to val and remove it from the queue, if not found, return nil
```

pkg/queue/any_queue.go
```diff
@@ -0,0 +1,41 @@
...
+func (q *AnyQueue) Get(val interface{}) interface{} {
+	for i, a := range q.slice {
+		if q.eq(a, val) {
+			q.slice = append(q.slice[:i], q.slice[i+1:]...)
```
AstroProfundis, line 31: So this is not a FIFO queue?

PR #1072, opened 2021-01-15
Feature/validate port user group

9547:

- check port: should between 0 and 65535
- check username/groupname

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/pkg/cluster/spec/spec.go
+++ b/pkg/cluster/spec/spec.go
@@ -58,7 +58,7 @@ type (
 	// specification in topology.yaml
 	GlobalOptions struct {
 		User            string               `yaml:"user,omitempty" default:"tidb"`
-		Group           string               `yaml:"group,omitempty"`
+		Group           string               `yaml:"group,omitempty" default:"pingcap"`
 		SSHPort         int                  `yaml:"ssh_port,omitempty" default:"22" validate:"ssh_port:editable"`
 		SSHType         executor.SSHType     `yaml:"ssh_type,omitempty" default:"builtin"`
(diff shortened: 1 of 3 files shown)


pkg/cluster/spec/spec.go
```diff
@@ -58,7 +58,7 @@ type (
 	// specification in topology.yaml
 	GlobalOptions struct {
 		User            string               `yaml:"user,omitempty" default:"tidb"`
-		Group           string               `yaml:"group,omitempty"`
+		Group           string               `yaml:"group,omitempty" default:"pingcap"`
```
lucklove, line 61: It's not a good idea to set a default value for `Group`, because by default, it should be same with `User`. 
https://github.com/pingcap/tiup/blob/master/pkg/cluster/module/user.go#L86

pkg/cluster/spec/validate.go
```diff
@@ -460,6 +474,61 @@ func (s *Specification) platformConflictsDetect() error {
 	return nil
 }
 
+func (s *Specification) portInvalidDetect() error {
+	portTypes := []string{
```
lucklove, line 478: How about just iterate every reflected field and apply check process if it's name like `XXXPort`

9547: Thanks for the advice

pkg/cluster/spec/spec.go
9547: OK, I'll revert it. I thought this was a forgotten place.

PR #1073, opened 2021-01-16
Fix/prometheus rules missing

9547:

I'm deploying a cluster with `rule_dir` set in topology.yaml:

```yaml
  monitoring_servers:
    - host: 172.19.0.101
      rule_dir: /tmp/local/prometheus
```

After deploy, found the `nodes.rules.yml`, `tikv.rules.yml` and other rules under `/home/tidb/deploy/prometheus/conf/` were missing. 

Should be merged after #1075 

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```


lucklove:
I'm a little confused, it seems what you changed has nothing to do with `RuleDir`, it's `DashboardDir` which is used to config grafana instead of prometheus.

Could you explain more about how it works?

9547:
> I'm a little confused, it seems what you changed has nothing to do with `RuleDir`, it's `DashboardDir` which is used to config grafana instead of prometheus.
> 
> Could you explain more about how it works?

Sorry for the confusing, It's both fixed for `RuleDir` and `DashboardDir`.

First of all, let's talk about the usage of `rule_dir` for Prometheus and `dashboard_dir` for Grafana. IMO, those fields are an addition to the package provided rules and dashboards, because the provided ones are sufficient for the normal use case.

But in the current implementation, if those fields were set, then It will only transfer the local files into the remote, without copy the provided ones too. https://github.com/pingcap/tiup/blob/4ece39a150f284f309ed4118df2c1e3ab536ed28/pkg/cluster/spec/prometheus.go#L315-L319 and this way the testcases failed for #1074

lucklove:
> IMO, those fields are an addition to the package provided rules and dashboards, because the provided ones are sufficient for the normal use case.
>
> But in the current implementation, if those fields were set, then It will only transfer the local files into the remote, without copy the provided ones too.

This is [by designed](https://docs.pingcap.com/zh/tidb/stable/tiup-cluster#%E6%9B%B4%E6%96%B0%E9%85%8D%E7%BD%AE), because the `modify` also includes `remove` some dashboards that not applicable for some users' case (for example, one user may not deploy tidb server at all, so he don't want see the tidb dashboard).

If we implement the dir as additional, the user can only add files and once the file added, he can never remove it with TiUP, it's not self-consistent, I think.

9547:
If the Grafana or Prometheus was deployed independently(without tidb or dm), then the default dashboards or rules seems redundant. But I 🤔 if someone uses TiUP, he'd prefer to deploy Prometheus with the full tidb  or dm stack.

If the user want to delete the dashboards or rules, maybe he can replace it with an empty file(not tested)

lucklove:
> If the Grafana or Prometheus was deployed independently(without tidb or dm), then the default dashboards or rules seems redundant. But I 🤔 if someone uses TiUP, he'd prefer to deploy Prometheus with the full tidb or dm stack.
> 
> If the user want to delete the dashboards or rules, maybe he can replace it with an empty file(not tested)

I have tried replacing node.json with empty file, the grafana can start, but with error logs:
屏幕快照 2021-01-20 下午12 32 09

9547:
A file with empty JSON seems worked

image

9547:
Let me see if there's a better solution or tradeoff.

9547:
@lucklove Let's put aside the Grafana dashboards, and please look at the Prometheus issue. 
Currently, the _prometheus.yml_ template will always include those rules if any host is monitored: https://github.com/pingcap/tiup/blob/d4881bacd219e4c802dea0fa83909051cc094496/templates/config/prometheus.yml.tpl#L12-L16 but if the user specifies the `rule_dir` but without those rules, then the _prometheus.yml_ is actually problematic and the rules specified by the user are not working. 

For this bug, I'll try to introduce a new member `CustomRules`, and iterator the `rule_dir` and append `CustomRules`:

```yaml
 rule_files:
{{- if .MonitoredServers}}
  - 'node.rules.yml'
  - 'blacker.rules.yml'
  - 'bypass.rules.yml'
{{- end}}
{{- range .CustomRules}}
  - '{{.}}'
{{- end}}
```

lucklove:
> @lucklove Let's put aside the Grafana dashboards, and please look at the Prometheus issue.
> Currently, the _prometheus.yml_ template will always include those rules if any host is monitored:
> 
> https://github.com/pingcap/tiup/blob/d4881bacd219e4c802dea0fa83909051cc094496/templates/config/prometheus.yml.tpl#L12-L16
> 
> but if the user specifies the `rule_dir` but without those rules, then the _prometheus.yml_ is actually problematic and the rules specified by the user are not working.
> For this bug, I'll try to introduce a new member `CustomRules`, and iterator the `rule_dir` and append `CustomRules`:
> 
> ```yaml
>  rule_files:
> {{- if .MonitoredServers}}
>   - 'node.rules.yml'
>   - 'blacker.rules.yml'
>   - 'bypass.rules.yml'
> {{- end}}
> {{- range .CustomRules}}
>   - '{{.}}'
> {{- end}}
> ```

In this way, there may be two `node.rules.yaml` if the user provide it in his directory:

```
- node.rules.yml # one for {{- if .MonitoredServers}}
- node.rules.yml # one for {{- range .CustomRules}}
```

9547:
Those rules will be ignored if duplicated:

  - 'node.rules.yml'
  - 'blacker.rules.yml'
  - 'bypass.rules.yml'

The user given ones has higher priority

lucklove:
> Those rules will be ignored if duplicated

If duplicated rules will not trigger any error, I think it's OK

lucklove:
Sorry that there is something conflict after refactoring the context: https://github.com/pingcap/tiup/pull/1069

9547:
Seems the bot can't respond to the `/retest` command

9547:
@lucklove PTAL, the current implementation will append the Grafana Dashboards or Prometheus ruls if `dashbaoard_url` or `rule_dir` specified.

lucklove:
> @lucklove PTAL, the current implementation will append the Grafana Dashboards or Prometheus ruls if `dashbaoard_url` or `rule_dir` specified.

It seems we didn't solve the deletion problem yet: the user can only append rule/dashboard (we can use an empty file, but it will print error logs and it's not graceful)

lucklove:
IMO, it's not a good idea to change current behavior: 

if some user already deployed a cluster with local directory and delete some rule/dashboard from that directory, after he upgrade tiup-cluster, the removed rule/dashboard will be back again. And the removed rule will fire some alert that he didn't care at all. This is unexpected.

9547:
> IMO, it's not a good idea to change current behavior since it's documented and released long ago:
> 
> if some user already deployed a cluster with local directory and delete some rule/dashboard from that directory, after he upgrade tiup-cluster, the removed rule/dashboard will be back again. And the removed rule will fire some alert that he didn't care at all. This is unexpected.

I will hold on to this PR for the moment and consider it when I have a suitable opportunity next time

9547:
I'm just going to close this first.

PR #1078, opened 2021-01-19
*: remove v0manifest

### LINKED ISSUE #906: Remove v0manifest
The `v0manifest` was our initial repository manifest implementation, as we have switched to TUF implementation `v1manifest` since `v1.0.0`, the `v0manifest` is deprecated and only kept for backward compatibility.

The `v1.0.0` was released at May 2020, and it would be reasonable to remove the legacy `v0manifest` codes from `v1.4` and later releases.

Some helper utils and tools used for v0 -> v1 migration should also be removed.

AstroProfundis:
Close #906 

Remove `v0manifest` and legacy codes using it.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Release notes:

```release-note
Remove `v0manifest` as it is not longer been used.
```


pkg/environment/env.go
```diff
@@ -368,28 +256,6 @@ func ParseCompVersion(spec string) (string, pkgver.Version) {
 
 // IsSupportedComponent return true if support if platform support the component.
 func (env *Environment) IsSupportedComponent(component string) bool {
```
lucklove, line 258: Should we delete this function?

pkg/environment/env.go
```diff
@@ -99,38 +96,19 @@ func InitEnv(options repository.Options) (*Environment, error) {
 		return nil, err
 	}
 
-	var repo *repository.Repository
 	var v1repo *repository.V1Repository
```
nrc, line 99: should we rename `v1repo` to `repo`? And likewise `v1Repo` and `V1Repository`

nrc: and V1Manifest, etc.

nrc (approved):
lgtm

pkg/environment/env.go
AstroProfundis: I think it's OK, as we don't guarantee there won't be `v2repo` in the future. The only misleading thing might be that we still use `v0.1.0` as `spec_version` in `v1Manifest`.

Issue #849, opened 2020-10-19
tiup supports config.ini for configuring grafana and prometheus


sultan8252:
## Feature Request

**Is your feature request related to a problem? Please describe:**
Manually adjust the grafana related configuration after each reload operation or cluster upgrade. Similar problems and prometheus

**Describe the feature you'd like:**
The configuration in config.ini related to grafana and prometheus is supported in tiup.

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

lucklove:
Dose [this](https://docs.pingcap.com/zh/tidb/stable/tiup-cluster#%E6%9B%B4%E6%96%B0%E9%85%8D%E7%BD%AE) help on this issue?

sultan8252:
The grafana address is accessed through the nginx reverse proxy and needs to be manually specified after each reload

lucklove:
> The grafana address is accessed through the nginx reverse proxy and needs to be manually specified after each reload

So do you want to config the [domain](https://github.com/pingcap/tiup/blob/master/templates/config/grafana.ini.tpl#L42) filed in topology file?

mianhk:
Configuring Grafana to run behind a reverse proxy need root_url defined refrenced to : https://grafana.com/tutorials/run-grafana-behind-a-proxy/#1
But now when we use reload/scale-in/scale-out command ，the grafana.ini will be refreshed and need to be edit manually.

PR #1085, opened 2021-01-21
fix: tiup supports define root_url and enable anonymous in grafana.ini

mianhk:
fix: #849 

1. You can configure Grafana to run behind a reverse proxy.
   reference: https://grafana.com/tutorials/run-grafana-behind-a-proxy/#1
2. You can set grafana access anonymous

### Check List 

Tests 

 - Unit test
 - Integration test
```
tiup cluster edit-config test_cluster

support config:
- host: 127.0.0.2
  ssh_port: 22
  arch: amd64
  os: linux
  username: admin
  password: admin
  anonymous_enable: true
  root_url: /tidb/grafana/test-cluster
  domain: example.com
```

 Then, you can get a grafana.ini like:
```
[auth.anonymous]
enabled = true
# The full public facing url
[server]
domain = example.com
...
root_url = /tidb/grafana/test-cluster
server_from_sub_path = true
```
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```


lucklove:
In https://grafana.com/tutorials/run-grafana-behind-a-proxy/#1

It says:

```
When running Grafana behind a proxy, you need to configure the domain name to let Grafana know how to render links and redirects correctly.
```

And the final example is like:

```
[server]
domain = example.com
root_url = %(protocol)s://%(domain)s:%(http_port)s/grafana/
serve_from_sub_path = true
```

So I think maybe you forget to set the domain (by default, it's the IP address).

I suggest that you can add a `Domain` field and parse the domain from the root_url, if the root_url is not set, set `Domain` to the IP address.

mianhk:
> In https://grafana.com/tutorials/run-grafana-behind-a-proxy/#1
> 
> It says:
> 
> ```
> When running Grafana behind a proxy, you need to configure the domain name to let Grafana know how to render links and redirects correctly.
> ```
> 
> And the final example is like:
> 
> ```
> [server]
> domain = example.com
> root_url = %(protocol)s://%(domain)s:%(http_port)s/grafana/
> serve_from_sub_path = true
> ```
> 
> So I think maybe you forget to set the domain (by default, it's the IP address).
> 
> I suggest that you can add a `Domain` field and parse the domain from the root_url, if the root_url is not set, set `Domain` to the IP address.

OK,I add domain field parsed from domain if root_url is set or not cause someone may only set domain.

Issue #846, opened 2020-10-16
Patch command corner case when using a wrong tar.gz file.


wshwsh12:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?

Step 1: Patch a tar.gz file build by Jenkins.
             1.   tar tidb-server.tar.gz, it will remove tidb-server binary and create tidb-server directory (There is a directory named tidb-server in tidb repo).
             2.  tidb-server directory can't execute so that tidb will restart failed.

Step 2: Patch a tar.gz file right.
             1. tar can't replace the directory to the right binary file.

2. What did you expect to see?

In Step 1, the bin directory should be polluted if the tar.gz file is invalid.
In Step 2, the file/directory should be replaced successfully.

3. What did you see instead?

Get some tar error.  And it can't back to normal by tiup itself.

4. What version of TiUP are you using (`tiup --version`)?

lucklove:
We should unzip the tar file to a temporary dir, and check if it has correct entry that satisfies TiUP, and then replace the current package with it.

PR #1091, opened 2021-01-22
Fix the issue that a dummy entry may cheat patch command

lucklove:
Fix https://github.com/pingcap/tiup/issues/846

### Check List 

Tests 

 - Integration test

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```


pkg/cluster/manager/patch.go
```diff
@@ -118,8 +119,18 @@ func checkPackage(bindVersion spec.BindVersion, specManager *spec.SpecManager, n
...
+	if err != nil {
+		if os.IsNotExist(err) {
+			return errors.Errorf("entry %s not found in package %s", entry, packagePath)
+		}
+		return errors.AddStack(err)
+	}
+	if !fi.Mode().IsRegular() {
+		return errors.Errorf("entry %s in package %s is not a regular file", entry, packagePath)
+	}
+	if fi.Mode()&0500 != 0500 {
```
AstroProfundis, line 132: There is `os.Executable()` that works cross-platfrom.

AstroProfundis: My fault, this is not for checking file permission.

AstroProfundis (approved):

PR #1103, opened 2021-01-27
Add debug mode for checkpoint

lucklove:
Don't panic in prod mode

tiup-cluster or tiup-dm will panic if the context passed to checkpoint is not generated by checkpoin.NewContext. It hurts the user experience when user use tiup in prod mode.

Add `DEBUG_CHECKPOIN` env which should be used in CI. Only in debug mode, tiup-cluster can panic.

### Check List 

Tests 

 - Integration test

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```


pkg/checkpoint/checkpoint.go
```diff
@@ -59,9 +64,13 @@ func SetCheckPoint(file string) error {
 
 // Acquire wraps CheckPoint.Acquire
 func Acquire(ctx context.Context, point map[string]interface{}) *Point {
-	if ctx.Value(goroutineKey) == nil {
-		panic("the value of the key " + goroutineKey + " not found in context, use the ctx from checkpoint.NewContext please")
+	if ctx.Value(goroutineKey) == nil || ctx.Value(semKey) == nil {
+		if DebugCheckpoint {
+			panic("the context passed to checkpoint.Acquire is not generated by checkpoint.NewContext")
+		}
+		ctx = NewContext(ctx)
```
AstroProfundis, line 71: I think we should print a debug log here

AstroProfundis (approved):

Issue #1111, opened 2021-01-29
tiup clone mirror --full break when not found resource


mianhk:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?

As I want to use a full offline mirror which convenients for me to update it by command like:
```
tiup mirror clone path/to/dir --full --os=linux --arch=amd64
```
and it may output and break:
```
Error: download resource: cluster: url https://tiup-mirrors.pingcap.com/cluster-v0.7.0-linux-amd64.tar.gz: not found
```
The problem is it won't continue even the `not found` component can be skipped.  
  

2. What did you expect to see?

Continue when `not found` component appears.

3. What did you see instead?

4. What version of TiUP are you using (`tiup --version`)?

lucklove:
I have fixed this by removing the 0.x versions (which are out of date). I think it should success when you retry.

mianhk:
OK，thanks!

PR #1112, opened 2021-01-29
fix:command 'tiup mirror clone --full' break by NOT_FOUND component

mianhk:
fix: #1111 

### Check List 

Tests 

 - Unit test
 - Integration test
 exec command:
 ```
tiup mirror clone path/to/dir --full --os=linux --arch=amd64
```
output:
```
Skip exists file: /home/work/opbin/ftp/tiup_mirror/v4.2/cluster-v0.4.9-linux-amd64.tar.gz
download https://tiup-mirrors.pingcap.com/cluster-v0.7.0-linux-amd64.tar.gz 0 B ?% ? p/s
download resource: cluster.err:url https://tiup-mirrors.pingcap.com/cluster-v0.7.0-linux-amd64.tar.gz: not found
Skip exists file: /home/work/opbin/ftp/tiup_mirror/v4.2/cluster-v0.0.4-linux-amd64.tar.gz
Skip exists file: /home/work/opbin/ftp/tiup_mirror/v4.2/cluster-v1.0.2-linux-amd64.tar.gz
```
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/pkg/repository/clone_mirror.go
+++ b/pkg/repository/clone_mirror.go
@@ -15,6 +15,7 @@ package repository
 
 import (
 	"fmt"
+	"github.com/pingcap/tiup/pkg/colorutil"
 	"os"
 	"path"
 	"path/filepath"
@@ -309,7 +310,11 @@ func cloneComponents(repo *V1Repository,
 					continue
 				}
 				if err := download(targetDir, tmpDir, repo, &versionItem); err != nil {
-					return nil, errors.Annotatef(err, "download resource: %s", name)
+					if strings.Contains(err.Error(), "not found") {
+						_, _ = colorutil.ColorErrorMsg.Printf("download resource: %s.err:%s\n", name, err)
+					} else {
+						return nil, errors.Annotatef(err, "download resource: %s", name)
+					}
 				}
 			}
 		}


pkg/repository/clone_mirror.go
```diff
@@ -15,6 +15,7 @@ package repository
 
 import (
 	"fmt"
+	"github.com/pingcap/tiup/pkg/colorutil"
```
AstroProfundis, line 18: Please run `make check` before commit so that the `go fmt` is invoked automatically to format the code.

pkg/repository/clone_mirror.go
```diff
@@ -309,7 +310,11 @@ func cloneComponents(repo *V1Repository,
 					continue
 				}
 				if err := download(targetDir, tmpDir, repo, &versionItem); err != nil {
-					return nil, errors.Annotatef(err, "download resource: %s", name)
+					if strings.Contains(err.Error(), "not found") {
```
AstroProfundis, line 313: I think it's better check for HTTP status code, that might need to be implemented in underlying functions.

AstroProfundis (commented):
Thanks for your contribution!

There's a question that we'd actually want to be aware of download failures so that we don't accidentally produce an offline package with something missing. So that we may need a better way than just ignore the 404 error.

I'd say that either optimize the error information a little, or maybe we could add a switch to control the error handling logic.

@lucklove do you have any idea? I don't really think it's reasonable to download cluster v0.7.0 though.

lucklove:
It's not a good idea to ignore this error because it's critical.

I've checked the mirror the tiup-cluster-v0.7.0 is missing, it's a problem we should fix:

```sh
[root@ip-172-16-5-140 ~]# tiup install cluster:v0.7.0
download https://tiup-mirrors.pingcap.com/cluster-v0.7.0-linux-amd64.tar.gz 0 B ?% ? p/s               
Error: url https://tiup-mirrors.pingcap.com/cluster-v0.7.0-linux-amd64.tar.gz: not found
```

mianhk:
> It's not a good idea to ignore this error because it's critical.
> 
> I've checked the mirror and found that the tiup-cluster-v0.7.0 is missing, it's a problem we should fix instead of just ignore:
> 
> ```shell
> [root@ip-172-16-5-140 ~]# tiup install cluster:v0.7.0
> download https://tiup-mirrors.pingcap.com/cluster-v0.7.0-linux-amd64.tar.gz 0 B ?% ? p/s               
> Error: url https://tiup-mirrors.pingcap.com/cluster-v0.7.0-linux-amd64.tar.gz: not found
> ```

Yes, this is not a good way actually,but  i can't download full mirror before . Now problem has been solved.

Issue #1122, opened 2021-02-02
Add support for darwin/arm64 (Apple M1)


terasum:
## General Question

I'm using the new Apple Macbook Pro with M1 chip, and when I'm running this command
```
 curl --proto '=https' --tlsv1.2 -sSf https://tiup-mirrors.pingcap.com/install.sh | sh 
```

Error occured:

```
tar: Error opening archive: Unrecognized archive format
Failed to download and/or extract tiup archive.
```

and I printed out the download URL:
```
https://tiup-mirrors.pingcap.com/tiup-darwin-arm64.tar.gz?20210202131357 -o /tmp/tiup-darwin-arm64.tar.gz
```
opened the URL, and It returns:

```
{
"error": "Document not found"
}
```

So, will you support the new Apple Mac with M1 chip?

AstroProfundis:
Apple Silicon is supported since Go 1.16, but we build and release TiUP and buildin components with 1.13, so there would be some works to do:

 - [ ] Wait until Go 1.16 is released
 - [ ] Switch our release CI piplines to use Go 1.16, this may need rebuilding of the base image of packaging container
 - [ ] Add a new release target for `darwin/arm64`
 - [ ] Remove the error message in installer scripts
 - [ ] Testing

terasum:
@AstroProfundis is there any resolution with Rosetta?

AstroProfundis:
I don't use apple, have you tried to run current TiUP binaries for `darwin/amd64` with Rosetta? Is there any particular error?

AstroProfundis:
I think we need some contributors holding M1 devices to help with the basic tests... @baurine are you using an M1 Mac?

baurine:
> I think we need some contributors holding M1 devices to help with the basic tests... @baurine are you using an M1 Mac?

Our new colleague @shhdgit uses it and he bases Hangzhou office.

shhdgit:
@AstroProfundis I used darwin/amd64 binary with Rosetta yesterday, but tikv does not support apple silicon yet.
[tikv/rocksdb#219](https://github.com/tikv/rocksdb/pull/219)
[tikv/tikv#9192](https://github.com/tikv/tikv/issues/9192)

And I can help test tiup.

terasum:
Hi @AstroProfundis, according to @shhdgit, until the tikv and the rust-rocksdb fixes the Apple Silicon related issues, we can't resolve this problem.

AstroProfundis:
[...]

PR #1123, opened 2021-02-02
Add darwin-arm64 not support hint for install.sh

terasum:
The darwin-arm64 is not supported currently, and add some friendly hint.

temporary fix for #1122 

Add some hints 

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -27,6 +27,11 @@ if [ -z "$arch" ]; then
     exit 1
 fi
 
+if [ "$os-$arch" == "darwin-arm64" ]; then
+    echo "Architecture darwin-arm64 not supported." >&2
+    exit 1
+fi
+
 if [ -z "$TIUP_HOME" ]; then
     TIUP_HOME=$HOME/.tiup
 fi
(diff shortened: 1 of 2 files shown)


lucklove:
Could you please modify [local_install.sh](https://github.com/pingcap/tiup/blob/master/pkg/cluster/template/install/local_install.sh.go#L46) too? This is used for offline  installation

terasum:
@lucklove sure! I will fix local_install.sh too!

terasum:
@lucklove Hi, please take a look again, thanks.

baurine:
Will we support it and when? it seems some guys start to use M1.

lucklove:
@baurine We need wait TiDB/TiKV/PD support that first, otherwise the user only have a bare TiUP

baurine:
> @baurine We need wait TiDB/TiKV/PD support that first, otherwise the user only have a bare TiUP

make sense, now the TiKV can't run on M1.

PR #1144, opened 2021-02-21
cluster: updateTopology maybe hang forever

### LINKED ISSUE #333: Add timeout when updatetopology
`tiup cluster start` will statll at `updatetopology` stage and don't exit when server can't connect to pd server.
Suggest to add timeout at `updatetopology`.

9547:

fix #333 

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
fixed an issue with hang forever when updating topology
```

### DIFF
--- a/pkg/cluster/task/update_topology.go
+++ b/pkg/cluster/task/update_topology.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"path/filepath"
+	"time"
 
 	"github.com/pingcap/errors"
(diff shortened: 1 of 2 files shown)


AstroProfundis (approved):

AstroProfundis:
Timeout is good, but I think we should also refine the process of checking PD status and fail there. (I believe we should have timeout set for HTTP clients)

Issue #1033, opened 2020-12-31
Reduct jitter feature


King-Dylan:
## Feature Request
Reduct jitter feature
**Is your feature request related to a problem? Please describe:**

 - [ ] before upgrade check pd miss-peer pending-peer and give a suggestion :solve miss-peer or ignore . Check the status of the region before the upgrade to reduce upgrade jitter or failure caused by the abnormal region before the upgrade.
 - [ ] before eviting tikv leader configure leader-scheduler-limit store-limit region-scheduler-limit.speed up leader transfer, after upgrade change to default value。can take the original value and add a number, and make a judgment of not greater than leader-scheduler-limit(64) store-limit(1024) region-scheduler-limit(1024) example:
 1.tiup cluster  upgrade xxxx 
 2.background(get default value) 
      tiup ctl pd -u xxx.xxx.xxx.xxx:2379 config show |grep  leader-schedule-limit 
      tiup ctl pd -u xxx.xxx.xxx.xxx:2379 config show |grep  region-schedule-limit
      tiup ctl pd -u xxx.xxx.xxx.xxx:2379 store limit
3.background
      tiup ctl pd -u xxx.xxx.xxx.xxx:2379 config set  leader-schedule-limit {default value}+32
      tiup ctl pd -u xxx.xxx.xxx.xxx:2379 config set  region-schedule-limit  {default value}+512
      tiup ctl pd -u xxx.xxx.xxx.xxx:2379 store limit all  {default value}+512
4.tikv rolling upgrade end
5.background(get value and check default >=0) 
      tiup ctl pd -u xxx.xxx.xxx.xxx:2379 config show |grep  leader-schedule-limit 
      tiup ctl pd -u xxx.xxx.xxx.xxx:2379 config show |grep  region-schedule-limit
      tiup ctl pd -u xxx.xxx.xxx.xxx:2379 store limit
6.background (setback default value)
      tiup ctl pd -u xxx.xxx.xxx.xxx:2379 config set  leader-schedule-limit {value}-32
      tiup ctl pd -u xxx.xxx.xxx.xxx:2379 config set  region-schedule-limit {value}-512
      tiup ctl pd -u xxx.xxx.xxx.xxx:2379 store limit all {value}-512

**Describe the feature you'd like:**

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

King-Dylan:
[...]

PR #1161, opened 2021-02-26
cluster: increase schedule limit during upgrade of tikv

AstroProfundis:
Implement task 2 in #1033 

Increase schedule limits before upgrade of TiKV, and set them back after.

### Check List 

Tests 

 - Integration test

Code changes

Side effects

 - Increased code complexity

Release notes:

```release-note
NONE
```

### DIFF
--- a/pkg/cluster/operation/upgrade.go
+++ b/pkg/cluster/operation/upgrade.go
@@ -18,8 +18,11 @@ import (
 	"crypto/tls"
 	"reflect"
 	"strconv"
+	"time"
 
+	perrs "github.com/pingcap/errors"
 	"github.com/pingcap/tiup/pkg/checkpoint"
+	"github.com/pingcap/tiup/pkg/cluster/api"
 	"github.com/pingcap/tiup/pkg/cluster/spec"
 	"github.com/pingcap/tiup/pkg/logger/log"
 	"github.com/pingcap/tiup/pkg/set"
@@ -31,6 +34,10 @@ func init() {
 	checkpoint.RegisterField(
 		checkpoint.Field("operation", reflect.DeepEqual),
 		checkpoint.Field("instance", reflect.DeepEqual),
+	)
+	checkpoint.RegisterField(
+		checkpoint.Field("operation", reflect.DeepEqual),
+		checkpoint.Field("func", reflect.DeepEqual),
 	)
 }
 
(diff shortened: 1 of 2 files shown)


pkg/cluster/operation/upgrade.go
```diff
@@ -46,13 +53,33 @@ func Upgrade(
 	components := topo.ComponentsByUpdateOrder()
 	components = FilterComponent(components, roleFilter)
 
+	pdList := make([]string, 0)
```
lucklove, line 56: I think we can move L56-L61 to L76 and L120 and change it as

```go
pdClient := api.NewPDClient(topo.(*spec.Specification).GetPDList(), 10*time.Second, tlsCfg)
```

One line is better than 6 lines

pkg/cluster/operation/upgrade.go
```diff
@@ -142,3 +179,86 @@ func Addr(ins spec.Instance) string {
...
+		point.Release(err,
+			zap.String("operation", "upgrade"),
+			zap.String("func", "increaseScheduleLimit"),
+			zap.Int("currLeaderScheduleLimit", currLeaderScheduleLimit),
+			zap.Int("currRegionScheduleLimit", currRegionScheduleLimit),
+		)
+	}()
+
+	if data := point.Hit(); data != nil {
+		currLeaderScheduleLimit = data["currLeaderScheduleLimit"].(int)
```
lucklove, line 213: I think this may panic because `currLeaderScheduleLimit` and `currRegionScheduleLimit` will be stored as float64 in json.

pkg/cluster/operation/upgrade.go
```diff
@@ -142,3 +179,86 @@ func Addr(ins spec.Instance) string {
...
+			return currLeaderScheduleLimit, currRegionScheduleLimit, err
+		}
+	}
+	if currRegionScheduleLimit < regionScheduleLimitThreshold {
+		newLimit := currRegionScheduleLimit + regionScheduleLimitOffset
+		if newLimit > regionScheduleLimitThreshold {
+			newLimit = regionScheduleLimitThreshold
+		}
+		if err := pc.SetReplicationConfig("region-schedule-limit", newLimit); err != nil {
+			return currLeaderScheduleLimit, currRegionScheduleLimit, err
```
lucklove, line 250: do we need rollback `leader-schedule-limit` here?

Maybe changes L197 `err error` -> `retErr err` a defer on L 242:

```go
defer func() {
  if retErr != nil {
    pc.SetReplicationConfig("leader-schedule-limit", currLeaderScheduleLimit)
  }
}()
```

pkg/cluster/operation/upgrade.go
AstroProfundis: In that way, we'll be creating PD client multiple times during the upgrade process.

pkg/cluster/operation/upgrade.go
AstroProfundis: If setting it to a new value fails, I don't think we need to try to set it to the old value again.

lucklove: I mean we should revert `leader-schedule-limit` when setting `region-schedule-limit` faild

PR #1191, opened 2021-03-05
Fix the issue that tiup may choose yanked version if it's already installed

lucklove:
### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/pkg/environment/env.go
+++ b/pkg/environment/env.go
@@ -187,24 +187,24 @@ func (env *Environment) downloadComponent(component string, version pkgver.Versi
 func (env *Environment) SelectInstalledVersion(component string, ver pkgver.Version) (pkgver.Version, error) {
 	installed, err := env.Profile().InstalledVersions(component)
 	if err != nil {
-		return "", err
-	}
-
-	errInstallFirst := errors.Annotatef(ErrInstallFirst, "use `tiup install %s` to install component `%s` first", component, component)
-	if len(installed) < 1 {
-		return "", errInstallFirst
+		return ver, err
 	}
 
 	versions := []string{}
 	for _, v := range installed {
 		vi, err := env.v1Repo.ComponentVersion(component, v, true)
 		if err != nil {
-			return "", err
+			return ver, err
 		}
 		if vi.Yanked {
 			continue
 		}
 		versions = append(versions, v)
+	}
+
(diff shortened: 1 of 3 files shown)


pkg/environment/env.go
```diff
@@ -36,6 +36,11 @@ const (
 	tiupName = "tiup"
 )
 
+var (
+	// ErrInstallFirst indicates that a component/version is not installed
+	ErrInstallFirst = errors.New("component not install")
```
AstroProfundis, line 41:
```suggestion
	ErrInstallFirst = errors.New("component not installed")
```

pkg/environment/env.go
```diff
@@ -179,17 +184,64 @@ func (env *Environment) downloadComponent(component string, version pkgver.Versi
...
-// DownloadComponentIfMissing downloads the specific version of a component if it is missing
-func (env *Environment) DownloadComponentIfMissing(component string, ver pkgver.Version) (pkgver.Version, error) {
-	versions, err := env.profile.InstalledVersions(component)
+func (env *Environment) SelectInstalledVersion(component string, ver pkgver.Version) (pkgver.Version, error) {
+	installed, err := env.Profile().InstalledVersions(component)
 	if err != nil {
 		return "", err
 	}
 
+	errInstallFirst := errors.Annotatef(ErrInstallFirst, "use `tiup install %s` to install component `%s` first", component, component)
```
AstroProfundis, line 193: why not use `ErrInstallFirst` here?

pkg/environment/env.go
```diff
@@ -200,23 +252,13 @@ func (env *Environment) DownloadComponentIfMissing(component string, ver pkgver.
...
-		for _, v := range versions {
-			if pkgver.Version(v) == ver {
-				installed = true
-				break
-			}
+	needDownload := false
+	if ver.IsEmpty() {
+		ver, err = env.SelectInstalledVersion(component, ver)
+		needDownload = errors.Cause(err) == ErrInstallFirst
+		if err != nil && errors.Cause(err) != ErrInstallFirst {
```
AstroProfundis, line 259:
```suggestion
		if err != nil && !needDownload {
```

pkg/environment/env.go
lucklove: Because this error will be displayed to the user

PR #1196, opened 2021-03-08
grafana: fix lightning's name not changed when cluster renamed

### LINKED ISSUE #1166: exec `tiup rename` ; Grafana dashboard lightning new and old documents
## Bug Report

1. What did you do?

3. What did you see instead?

9547:
fix #1166 

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/embed/templates/scripts/run_grafana.sh.tpl
+++ b/embed/templates/scripts/run_grafana.sh.tpl
@@ -6,6 +6,15 @@ set -e
 DEPLOY_DIR={{.DeployDir}}
 cd "${DEPLOY_DIR}" || exit 1
 
+# try to compatible with tidb-lightning's dashboard,
+# the tidb-lightning dashboard's uid is not set,
+# after the cluster renamed, the old dashboard will not be replaced with the newly one.
+# and Grafana use the unique uid as indepent dashaboard.
+# this script was stolen from https://stackoverflow.com/a/13844351
+if [[ -f dashboards/lightning.json ]]; then
+    grep -qxF uid dashboards/lightning.json || sed -i '$s/}/,\n"uid":"t5PjsKUGz"}/' dashboards/lightning.json
+fi
+
 LANG=en_US.UTF-8 \
 {{- if .NumaNode}}
 exec numactl --cpunodebind={{.NumaNode}} --membind={{.NumaNode}} bin/bin/grafana-server \
--- a/embed/templates/scripts/run_grafana.sh.tpl
+++ b/embed/templates/scripts/run_grafana.sh.tpl
@@ -6,15 +6,6 @@ set -e
 DEPLOY_DIR={{.DeployDir}}
 cd "${DEPLOY_DIR}" || exit 1
 
-# try to compatible with tidb-lightning's dashboard,
-# the tidb-lightning dashboard's uid is not set,
-# after the cluster renamed, the old dashboard will not be replaced with the newly one.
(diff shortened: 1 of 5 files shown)


embed/templates/scripts/run_grafana.sh.tpl
```diff
@@ -6,6 +6,15 @@ set -e
 DEPLOY_DIR={{.DeployDir}}
 cd "${DEPLOY_DIR}" || exit 1
 
+# try to compatible with tidb-lightning's dashboard,
+# the tidb-lightning dashboard's uid is not set,
+# after the cluster renamed, the old dashboard will not be replaced with the newly one.
```
lucklove, line 11: I checked old `dashboards/lightning.json` but there is no `uid` in this file. Is there something wrong?

embed/templates/scripts/run_grafana.sh.tpl
```diff
@@ -6,6 +6,15 @@ set -e
 DEPLOY_DIR={{.DeployDir}}
 cd "${DEPLOY_DIR}" || exit 1
 
+# try to compatible with tidb-lightning's dashboard,
```
lucklove, line 9: We need set this in https://github.com/pingcap/tidb-lightning/blob/master/metrics/grafana/lightning.json instead of set it with script in tiup-cluster. Because there are other tools depends on it, too.

9547: Good idea, the _lightning.json_ needs to be modified,  and for the existing offline tarballs, we need to compatible and fix it.

embed/templates/scripts/run_grafana.sh.tpl
```diff
@@ -6,6 +6,15 @@ set -e
...
+# the tidb-lightning dashboard's uid is not set,
+# after the cluster renamed, the old dashboard will not be replaced with the newly one.
+# and Grafana use the unique uid as indepent dashaboard.
+# this script was stolen from https://stackoverflow.com/a/13844351
+if [[ -f dashboards/lightning.json ]]; then
+    grep -qxF uid dashboards/lightning.json || sed -i '$s/}/,\n"uid":"t5PjsKUGz"}/' dashboards/lightning.json
```
lucklove, line 15: So could you remove these lines and open a PR on https://github.com/pingcap/tidb-lightning/blob/master/metrics/grafana/lightning.json?

9547: So for the Grafana package that has been released, are we not going to fix it, to keep the code simple?

lucklove: Sorry for my mistake: https://github.com/PingCAP-QE/ci/blob/jenkins-pipelines/jenkins/pipelines/cd/tiup/grafana-tiup-mirror-update-test.groovy#L95

We should change BR instead...

lucklove: > So for the Grafana package that has been released, are we not going to fix it, to keep the code simple?

No, we shouldn't change codesthat not belong to TiUP. (I think the xxx.json in other repos can be treated as code). The user can upgrade cluster if he really want to fix this.

9547:
@lucklove PTAL

PR #1200, opened 2021-03-09
*: remove direct depends of dm and pd

AstroProfundis:
DM and PD are large repos depending on many other packages, directly depending on them introduces much complexity to us while we only use some struct from them to unmarshal API responses.

This nature also make us difficult to upgrade our dependencies if any package is depended by one of DM or PD (or worse, by both DM and PD) but with different version with us, and may even lead to a dependency version dead lock in some circumstances.

Remove direct depends of DM and PD by copying the structs we need from their repos.

This doesn't introduce more compatibility issue (if any) as if they changed the struct definitions, we need to update depends anyway.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Release notes:

```release-note
NONE
```


AstroProfundis:
And... This change reduces the binary size of related components by about 7MB :rofl: 

master:
```
$ env LANG=C ls -l bin
total 153000
-rwxr-xr-x 1 allen users 26284032 Mar  9 18:19 tiup
-rwxr-xr-x 1 allen users 22872064 Mar  9 18:19 tiup-bench
-rwxr-xr-x 1 allen users 14041088 Mar  9 18:19 tiup-client
-rwxr-xr-x 1 allen users 29593600 Mar  9 18:19 tiup-cluster
-rwxr-xr-x 1 allen users 28798976 Mar  9 18:19 tiup-dm
-rwxr-xr-x 1 allen users 26394624 Mar  9 18:19 tiup-playground
-rwxr-xr-x 1 allen users  8687616 Mar  9 18:19 tiup-server
```

PR:
```
$ env LANG=C ls -l bin
total 124740
-rwxr-xr-x 1 allen users 18661376 Mar  9 18:21 tiup
-rwxr-xr-x 1 allen users 22908928 Mar  9 18:21 tiup-bench
-rwxr-xr-x 1 allen users 14057472 Mar  9 18:21 tiup-client
-rwxr-xr-x 1 allen users 22355968 Mar  9 18:21 tiup-cluster
-rwxr-xr-x 1 allen users 21553152 Mar  9 18:21 tiup-dm
-rwxr-xr-x 1 allen users 19505152 Mar  9 18:21 tiup-playground
-rwxr-xr-x 1 allen users  8691712 Mar  9 18:21 tiup-server
```

PR #1211, opened 2021-03-15
Omitting deprecated `"users.*"` settings for TiFlash

JaySon-Huang:

A part of https://github.com/pingcap/tics/issues/1497.
TiFlash PRs on 5.0.x and 4.0.x: https://github.com/pingcap/tics/pull/1526, https://github.com/pingcap/tics/pull/1535

Some access control settings are inherited from Clickhouse and useless for TiFlash, we make these settings to be the default value if empty since v4.0.12 and v5.0.0. Some settings (like `password = ""`) make trouble for TiFlash users when their configuration files are scanned by a security scanner. 

This PR make TiUP don't generate those deprecated settings for cluster version >= v4.0.12 and not euqals to v5.0.0-rc

Ignore the `'users.*'` and `'quotas.*'` settings in TiFlash if the cluster version >= v4.0.12 and not equals to v5.0.0-rc.
But keep those settings for version <= v4.0.11

### Check List 

Tests 

 - Unit test

Code changes

 - Has persistent data change

Side effects

 - N/A

Related changes

Release notes:

```release-note
Omitting deprecated `"users.*"` settings in the configuration file of TiFlash.
```


JaySon-Huang:
Do not merge until we confirm that v4.0.12 will contain that commit.

JaySon-Huang:
@AstroProfundis PTAL

pkg/cluster/spec/tiflash.go
```diff
@@ -367,6 +368,32 @@ func (i *TiFlashInstance) initTiFlashConfig(cfg *scripts.TiFlashScript, clusterV
 		pathConfig = fmt.Sprintf(`path: "%s"`, cfg.DataDir)
 	}
 
+	if (semver.Compare(clusterVersion, "v4.0.12") >= 0 && semver.Compare(clusterVersion, "v5.0.0-rc") != 0) || clusterVersion == "nightly" {
```
AstroProfundis, line 371: It's better to use `utils.Version(clusterVersion).IsNightly()` to check for nightly versions, the actual version in topo will be expanded from `nightly` to a specific version at that date.

JaySon-Huang: I push a commit to use `utils.Version(clusterVersion).IsNightly()` for TiFlash.

But I'm wondering should I also change this code in `playground.go`?
https://github.com/pingcap/tiup/blob/bd3cf97ef84f62be369b1c4af95b5901872c2803/components/playground/playground.go#L688

lucklove: I think so, it's better to change playground.go too.

JaySon-Huang: Addressed.

AstroProfundis (approved):

PR #1212, opened 2021-03-15
Feature: cache timestamp.json to accelerate

lucklove:
We implemented a base function, `ensureManifests()`, which will fetch `n.root.json` and `timestamp.json`, and if the timestamp.json is a new one, fetch `snapshot.json` and `{component}.json`.

However, there are many places invoke `ensureManifests()` and TiUP will fetch `timestamp.json` and `root.json` many times even though there is no change on them, which cause TiUP run slowly.

Cache `timestamp.json` in memory so that TiUP will only fetch it once every time the user execute TiUP command.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/pkg/repository/v1_repository.go
+++ b/pkg/repository/v1_repository.go
@@ -48,8 +48,9 @@ var ErrUnknownVersion = errors.New("unknown version")
 // V1Repository represents a remote repository viewed with the v1 manifest design.
 type V1Repository struct {
 	Options
-	mirror Mirror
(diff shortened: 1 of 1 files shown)


AstroProfundis (approved):

AstroProfundis:
/hold

unit tests are failing

Issue #1218, opened 2021-03-16
Make advertise urls configurable


Connor1996:
## Feature Request

**Is your feature request related to a problem? Please describe:**

TiKV and PD have some `advertise-xxx-urls` to indicate the remote accessible address for clients under NAT environment. We are expected to set it individually, but tiup generates the run script with fixed advertised address(host:ip) 
https://docs.pingcap.com/zh/tidb/stable/command-line-flags-for-tikv-configuration#--advertise-addr and https://docs.pingcap.com/zh/tidb/stable/command-line-flags-for-pd-configuration#--advertise-client-urls 

**Describe the feature you'd like:**

In yaml file, add config for these `advertise-xxx-urls` and set the corresponding field of the run script(if one is not set, use default value as before).
```
tikv_servers:
   - host: xxxxxx
     port: xxxxx
     advertise_addr: xxxx
     advertise_status_addr: xxxx
pd_servers:
    - host: xxxx
      port: xxxx
      advertise_client_urls: xxx
      advertise_peer_urls: xxx
```

**Describe alternatives you've considered:**

None

**Teachability, Documentation, Adoption, Migration Strategy:**

None

PR #1219, opened 2021-03-16
[DNM] cluster: add advertise addr for tidb,tikv,pd

9547:

implement https://github.com/pingcap/tiup/issues/1218

- [ ] Need more test case
- [ ] Add example config inside templates

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/pkg/cluster/template/scripts/pd.go
+++ b/pkg/cluster/template/scripts/pd.go
@@ -16,6 +16,9 @@ package scripts
 import (
 	"bytes"
 	"errors"
+	"fmt"
+	"net"
+	"net/url"
 	"os"
 	"path"
 	"text/template"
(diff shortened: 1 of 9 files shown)


lucklove:
Hold on please, this is a critical change on product level... we are talking about this now

AstroProfundis (commented):
We should also add default values to these fields in `spec`, as most users don't need to customize them at all. And maybe some helper functions to get the address from an instance object.

pkg/cluster/template/scripts/pd.go
```diff
@@ -77,6 +84,36 @@ func (c *PDScript) WithPeerPort(port int) *PDScript {
 	return c
 }
 
+// WithAdvertisWithAdvertiseClientAddr set AdvertiseClientAddr field of PDScript
```
lucklove, line 87: The comment is not correct

lucklove:
Support in https://github.com/pingcap/tiup/pull/1224

PR #1224, opened 2021-03-18
Add advertise addr for TiKV and PD

lucklove:
Fix https://github.com/pingcap/tiup/issues/1218

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/embed/templates/scripts/run_pd.sh.tpl
+++ b/embed/templates/scripts/run_pd.sh.tpl
@@ -24,9 +24,9 @@ exec bin/pd-server \
 {{- end}}
     --name="{{.Name}}" \
     --client-urls="{{.Scheme}}://{{.ListenHost}}:{{.ClientPort}}" \
-    --advertise-client-urls="{{.Scheme}}://{{.IP}}:{{.ClientPort}}" \
+    --advertise-client-urls="{{.AdvertiseClientAddr}}" \
     --peer-urls="{{.Scheme}}://{{.ListenHost}}:{{.PeerPort}}" \
-    --advertise-peer-urls="{{.Scheme}}://{{.IP}}:{{.PeerPort}}" \
+    --advertise-peer-urls="{{.AdvertisePeerAddr}}" \
     --data-dir="{{.DataDir}}" \
     --initial-cluster="{{template "PDList" .Endpoints}}" \
     --config=conf/pd.toml \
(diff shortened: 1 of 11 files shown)


lucklove:
This is a hidden feature, no release not and user document will be provided.

9547:
Also need to add `WithXXXAddr` in pd's `ScaleConfig` https://github.com/pingcap/tiup/pull/1224/files?file-filters%5B%5D=.go#diff-52ed75eb339663407c56ff6cf1b6bdf3a0995e7e9da391e4d28f113c9f14eb39R260-R270

lucklove:
> Also need to add `WithXXXAddr` in pd's `ScaleConfig` https://github.com/pingcap/tiup/pull/1224/files?file-filters%5B%5D=.go#diff-52ed75eb339663407c56ff6cf1b6bdf3a0995e7e9da391e4d28f113c9f14eb39R260-R270

Oh, true

AstroProfundis:
Do we need to also implement custom default values for these fields in `spec`? Or just leaving them empty is enough?

9547:
> Do we need to also implement custom default values for these fields in `spec`? Or just leaving them empty is enough?
 
Seems it's hard to set the default value which is based on other field in the `spec`, the `advertise-xxx` address are based on the `host` and `port`(the other fields). 

But we can always set a default value when the `pkg/cluster/template/scripts.NewXXXScript` is first invoked.

AstroProfundis:
> based on other field in the `spec`

Actually we do have such logic as in https://github.com/pingcap/tiup/blob/197a8a3a0c0cab28116c1a122aa53d268d77b98f/pkg/cluster/spec/spec.go#L500

And yes that's complex so not necessarily a must, I'm wondering if it's possible for the filed to be uninitialized if we don't set it a default value during unmarshalling.

lucklove:
> Do we need to also implement custom default values for these fields in `spec`? Or just leaving them empty is enough?

It's not a good idea to fill these fields with default value when it's empty, because it's a hidden feature and we don't wish users change it.

AstroProfundis:
That sounds reasonable to me.

AstroProfundis (approved):

PR #1227, opened 2021-03-18
cluster: fix the issue when endpoints are empty in scale-out

### LINKED ISSUE #1226: scale-out action doesn't start tiflash success
## Bug Report

1. What did you do?

I deployed a cluster which doesn't has any TiFlash node, and then scale-out one TiFlash with the command `tiup cluster scale-out test tiflash.yaml` and execute `tiup cluster display test`.

The status of TiFlash node should be `UP`

3. What did you see instead?
The status of TiFlash node is `N/A`. And after I execute `tiup cluster start test -R tiflash`, It's `UP`.

And I tried to debug why this happens, I found the [guide to scale-out tiflash](https://docs.pingcap.com/zh/tidb/stable/scale-tidb-using-tiup#%E6%89%A9%E5%AE%B9-tiflash-%E8%8A%82%E7%82%B9) which says you should `tiup ctl pd -u <pd-host>:<pd-port> config set enable-placement-rules true` before scale-out TiFlash. 

But I found we have do this in the scale-out stage: https://github.com/pingcap/tiup/blob/master/pkg/cluster/spec/tiflash.go#L657

Then I debug the result of `pdClient.UpdateReplicateConfig(bytes.NewBuffer(enablePlacementRules))`, I found that it returns nil error in both `scale-out` and `start` action, but only changes `enable-placement-rules` (`curl pd:port/pd/api/v1/config|grep enable-placement-rules`) to `true` in `start` action. 

In another word, the `pdClient.UpdateReplicateConfig` not work when you `scale-out` tiflash but work when you start `tiflash`. That's not reasonable.

v1.3.5 and master

9547:
Fix #1226 

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/pkg/cluster/operation/action.go
+++ b/pkg/cluster/operation/action.go
@@ -76,7 +76,6 @@ func Start(
 	cluster spec.Topology,
 	options Options,
 	tlsCfg *tls.Config,
-	finalCluster spec.Topology,
 ) error {
 	uniqueHosts := set.NewStringSet()
 	roleFilter := set.NewStringSet(options.Roles...)
@@ -87,7 +86,7 @@ func Start(
 
 	for _, comp := range components {
 		insts := FilterInstance(comp.Instances(), nodeFilter)
-		err := StartComponent(ctx, insts, options, tlsCfg, finalCluster)
+		err := StartComponent(ctx, insts, options, tlsCfg)
 		if err != nil {
 			return errors.Annotatef(err, "failed to start %s", comp.Name())
 		}
@@ -183,7 +182,7 @@ func Restart(
 		return errors.Annotatef(err, "failed to stop")
 	}
 
-	err = Start(ctx, cluster, options, tlsCfg, cluster)
+	err = Start(ctx, cluster, options, tlsCfg)
 	if err != nil {
 		return errors.Annotatef(err, "failed to start")
 	}
@@ -453,13 +452,7 @@ func EnableMonitored(
 }
 
 // StartComponent start the instances.
-func StartComponent(
-	ctx context.Context,
-	instances []spec.Instance,
-	options Options,
-	tlsCfg *tls.Config,
-	finalCluster spec.Topology,
-) error {
+func StartComponent(ctx context.Context, instances []spec.Instance, options Options, tlsCfg *tls.Config) error {
 	if len(instances) == 0 {
 		return nil
 	}
@@ -477,7 +470,7 @@ func StartComponent(
 		// of checkpoint context every time put it into a new goroutine.
 		nctx := checkpoint.NewContext(ctx)
 		errg.Go(func() error {
-			if err := ins.PrepareStart(finalCluster, tlsCfg); err != nil {
+			if err := ins.PrepareStart(tlsCfg); err != nil {
 				return err
 			}
 			err := startInstance(nctx, ins, options.OptTimeout)
--- a/pkg/cluster/operation/action.go
+++ b/pkg/cluster/operation/action.go
@@ -470,7 +470,7 @@ func StartComponent(ctx context.Context, instances []spec.Instance, options Opti
 		// of checkpoint context every time put it into a new goroutine.
 		nctx := checkpoint.NewContext(ctx)
 		errg.Go(func() error {
-			if err := ins.PrepareStart(tlsCfg); err != nil {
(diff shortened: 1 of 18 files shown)


9547:
/cc @lucklove 

PTAL

pkg/cluster/operation/action.go
```diff
@@ -452,7 +453,13 @@ func EnableMonitored(
 }
 
 // StartComponent start the instances.
-func StartComponent(ctx context.Context, instances []spec.Instance, options Options, tlsCfg *tls.Config) error {
+func StartComponent(
+	ctx context.Context,
+	instances []spec.Instance,
+	options Options,
+	tlsCfg *tls.Config,
+	finalCluster spec.Topology,
```
AstroProfundis, line 461: How about passing a PD client object to the function, I think the `tlsCfg` could also be removed that way.

9547: Since `Instance` is an abstraction interface used by both TiDB and DM, while PD only exists in the TiDB scenario. IMHO it is not very good to pass PD client as parameters, or we can abstract a `MetaClient` interface, whose responsibility is for these metadata operations by the client, but the kinds of metadata operation in TIDB and DM are so different, seems hard to abstract.

lucklove: Then what about put the base topology into ctx since we may access it anywhere? Insert it at here: https://github.com/pingcap/tiup/blob/master/pkg/cluster/manager/scale_out.go#L151

9547: sounds good to me.

9547:
@AstroProfundis @lucklove  PTAL

pkg/cluster/spec/tiflash.go
```diff
@@ -640,25 +641,36 @@ type replicateConfig struct {
...
+	if topo == nil {
+		topo = i.topo
+	}
+
+	topo, ok := topo.(Topology)
+	if !ok {
+		return perrs.New("base topology in context is invalid")
 	}
 
+	endpoints := i.getEndpoints(topo.(Topology))
```
AstroProfundis, line 673:
```suggestion
	endpoints := i.getEndpoints(topo)
```

9547: Sorry, the variable `topo` was firstly introduced in line 663, that is the value of context, its type is `interface{}`,  the line 668 short variable declaration doesn't change its type.

I'll change the code style above to make it more maintainable.

pkg/cluster/spec/tiflash.go
```diff
@@ -652,25 +653,37 @@ type replicateConfig struct {
...
+func (i *TiFlashInstance) PrepareStart(ctx context.Context, tlsCfg *tls.Config) error {
 	// set enable-placement-rules to true via PDClient
-	pdClient := api.NewPDClient(endPoints, 10*time.Second, tlsCfg)
 	enablePlacementRules, err := json.Marshal(replicateConfig{
 		EnablePlacementRules: "true",
 	})
+	// this should not failed, else exit
 	if err != nil {
-		return nil
+		return err
```
lucklove, line 672:
```suggestion
		return errors.Annotate(err, "failed to marshal replicate config")
```

pkg/cluster/operation/action.go
```diff
@@ -470,7 +470,7 @@ func StartComponent(ctx context.Context, instances []spec.Instance, options Opti
 		// of checkpoint context every time put it into a new goroutine.
 		nctx := checkpoint.NewContext(ctx)
 		errg.Go(func() error {
-			if err := ins.PrepareStart(tlsCfg); err != nil {
+			if err := ins.PrepareStart(ctx, tlsCfg); err != nil {
```
lucklove, line 473: Use `nctx` in case we add checkpoint in `PrepareStart` in the future.

pkg/cluster/spec/tiflash.go
lucklove: We should give an annotate when received an error from third-party lib or standard lib.

pkg/cluster/operation/action.go
9547: +1

Issue #1223, opened 2021-03-17
display component uptime


9547:
## Feature Request

**Is your feature request related to a problem? Please describe:**

**Describe the feature you'd like:**

I want to display the uptime time for each component in `tiup {cluster|dm} display xxx`

**Describe alternatives you've considered:**

As some components' Prometheus metric API returns `process_start_time_seconds` metric, which we can use directly to represent the process's start timestamp, and use `time.Now() - start_timestamp` as uptime, those components as below: 

- pd
- tidb
- tikv
- ticdc
- drainer
- pump
- alertmanager, grafana, prometheus

Those components doesn't contain the `process_start_time_seconds`:

- tiflash
- tispark-{master,worker}

For those components that do not include this metric, especially TIFlash, we can wait until the product side is integrated before adding it. During this transition, we can use `ssh then ps` to see how long the process lives.

**Teachability, Documentation, Adoption, Migration Strategy:**

9547:
@lucklove @AstroProfundis PTAL

lucklove:
> As some components' Prometheus metric API returns process_start_time_seconds metric, which we can use directly to represent the process's start timestamp, and use time.Now() - start_timestamp as uptime

That's a good idea, but what if the user didn't deploy prometheus? Can we handle that case?

9547:
> > As some components' Prometheus metric API returns process_start_time_seconds metric, which we can use directly to represent the process's start timestamp, and use time.Now() - start_timestamp as uptime
> 
> 
> 
> That's a good idea, but what if the user didn't deploy prometheus? Can we handle that case?

I'm sorry, I didn't make myself clear, we are not using Prometheus, but use the component's metric API. Btw, if use Prometheus to query the latest uptime, this is not correct, maybe fetched the staled data, which is the last message before the process dies.

lucklove:
I've checked the metrics components returned, but I didn't find any metric that records the `start_timestamp`...

9547:
> I've checked the metrics components returned, but I didn't find any metric that records the `start_timestamp`...

Sorry for the spelling error, is `process_start_time_seconds`, not `start_timestamp`. BTW, we can use `systemctl status xxx | grep -Po ".*; \K(.*)(?= ago)"` to get the process's uptime  for those doesn't has metric.

lucklove:
> > I've checked the metrics components returned, but I didn't find any metric that records the `start_timestamp`...
> 
[...]

PR #1231, opened 2021-03-21
dm,cluster: display with uptime returned

9547:

implement #1223 

display the uptime of the instance, 

if this instance is up, then `uptime` means the alive time, else is the inactive time. 

see #1223 for more detail 

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code
image

image

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/components/cluster/command/display.go
+++ b/components/cluster/command/display.go
@@ -80,6 +80,7 @@ func newDisplayCmd() *cobra.Command {
 
 	cmd.Flags().StringSliceVarP(&gOpt.Roles, "role", "R", nil, "Only display specified roles")
 	cmd.Flags().StringSliceVarP(&gOpt.Nodes, "node", "N", nil, "Only display specified nodes")
+	cmd.Flags().BoolVar(&gOpt.ShowUptime, "uptime", false, "Display with uptime")
 	cmd.Flags().BoolVar(&showDashboardOnly, "dashboard", false, "Only display TiDB Dashboard information")
 	cmd.Flags().BoolVar(&showVersionOnly, "version", false, "Only display TiDB cluster version")
 
--- a/components/dm/command/display.go
+++ b/components/dm/command/display.go
@@ -54,6 +54,7 @@ func newDisplayCmd() *cobra.Command {
 	cmd.Flags().StringSliceVarP(&gOpt.Roles, "role", "R", nil, "Only display specified roles")
 	cmd.Flags().StringSliceVarP(&gOpt.Nodes, "node", "N", nil, "Only display specified nodes")
 	cmd.Flags().BoolVar(&showVersionOnly, "version", false, "Only display DM cluster version")
+	cmd.Flags().BoolVar(&gOpt.ShowUptime, "uptime", false, "Display DM with uptime")
 
 	return cmd
 }
--- a/components/dm/spec/logic.go
+++ b/components/dm/spec/logic.go
@@ -15,9 +15,11 @@ package spec
 
 import (
 	"context"
+	"crypto/tls"
 	"fmt"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/pingcap/tiup/pkg/logger/log"
 	"github.com/pingcap/tiup/pkg/meta"
@@ -88,6 +90,9 @@ func (c *DMMasterComponent) Instances() []Instance {
 					s.DataDir,
 				},
 				StatusFn: s.Status,
+				UptimeFn: func(tlsCfg *tls.Config) time.Duration {
+					return spec.UptimeByHost(s.Host, s.Port, tlsCfg)
+				},
 			},
 			topo: c.Topology,
 		})
@@ -216,6 +221,9 @@ func (c *DMWorkerComponent) Instances() []Instance {
 					s.DataDir,
 				},
 				StatusFn: s.Status,
+				UptimeFn: func(tlsCfg *tls.Config) time.Duration {
+					return spec.UptimeByHost(s.Host, s.Port, tlsCfg)
+				},
 			},
 			topo: c.Topology,
 		})
--- a/pkg/cluster/manager/display.go
+++ b/pkg/cluster/manager/display.go
@@ -17,6 +17,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math"
 	"sort"
 	"strings"
 	"time"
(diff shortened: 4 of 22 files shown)


components/dm/spec/logic.go
```diff
@@ -216,6 +221,9 @@ func (c *DMWorkerComponent) Instances() []Instance {
 					s.DataDir,
 				},
 				StatusFn: s.Status,
+				UptimeFn: func(tlsCfg *tls.Config) time.Duration {
```
AstroProfundis, line 224: This is actually "Last State Time"

pkg/cluster/manager/display.go
```diff
@@ -44,6 +45,7 @@ type InstInfo struct {
 	Ports     string `json:"ports"`
 	OsArch    string `json:"os_arch"`
 	Status    string `json:"status"`
+	Uptime    string `json:"uptime"`
```
AstroProfundis, line 48: I think it's better be "Since" in the table header, so no extra documents about what does it mean if the node is down will be ever needed.

pkg/cluster/spec/instance.go
```diff
@@ -95,6 +95,7 @@ type Instance interface {
 	UsedPorts() []int
 	UsedDirs() []string
 	Status(tlsCfg *tls.Config, pdList ...string) string
+	Uptime(tlsCfg *tls.Config) time.Duration
```
AstroProfundis, line 98: ditto

components/dm/spec/logic.go
9547: IMO, the Last State Time is the last timestamp we saw the state , but what I want to point is the alive time duration of the process.

pkg/cluster/manager/display.go
9547: Thanks for the advice, `Status` + `Since` forms the component's full information.

9547: I've added an argument `--uptime`, not `--since`, the latter seems weird. 🤔

components/cluster/command/display.go
```diff
@@ -80,6 +80,7 @@ func newDisplayCmd() *cobra.Command {
 
 	cmd.Flags().StringSliceVarP(&gOpt.Roles, "role", "R", nil, "Only display specified roles")
 	cmd.Flags().StringSliceVarP(&gOpt.Nodes, "node", "N", nil, "Only display specified nodes")
+	cmd.Flags().BoolVar(&gOpt.ShowUptime, "uptime", false, "Display with uptime")
```
AstroProfundis, line 83:
```suggestion
	cmd.Flags().BoolVar(&gOpt.ShowUptime, "uptime", false, "Show process uptime")
```

components/dm/command/display.go
```diff
@@ -54,6 +54,7 @@ func newDisplayCmd() *cobra.Command {
 	cmd.Flags().StringSliceVarP(&gOpt.Roles, "role", "R", nil, "Only display specified roles")
 	cmd.Flags().StringSliceVarP(&gOpt.Nodes, "node", "N", nil, "Only display specified nodes")
 	cmd.Flags().BoolVar(&showVersionOnly, "version", false, "Only display DM cluster version")
+	cmd.Flags().BoolVar(&gOpt.ShowUptime, "uptime", false, "Display DM with uptime")
```
AstroProfundis, line 57:
```suggestion
	cmd.Flags().BoolVar(&gOpt.ShowUptime, "uptime", false, "Show process uptime")
```

9547:
@AstroProfundis  @lucklove  PTAL

pkg/cluster/manager/display.go
```diff
@@ -237,16 +240,29 @@ func (m *Manager) GetClusterTopology(name string, opt operator.Options) ([]InstI
 			status = ins.Status(tlsCfg, masterActive...)
 		}
 
-		// Query the service status
-		if status == "-" {
+		since := "-"
+		if !opt.ShowUptime {
+			since = formatInstanceSince(ins.Uptime(tlsCfg))
```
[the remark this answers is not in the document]
9547: Looks like redundant code ...

pkg/cluster/manager/display.go
```diff
@@ -237,16 +240,29 @@ func (m *Manager) GetClusterTopology(name string, opt operator.Options) ([]InstI
...
+		}
+
+		// Query the service status and uptime
+		if status == "-" || (opt.ShowUptime && since == "-") {
```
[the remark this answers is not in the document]
9547: Sorry, the previous check is reversed, so the check here is necessary, if user specifies `opt.ShowUptime` but `since` is empty(not implemented or server was down), we try to fetch it via `ssh-systemctl-status`)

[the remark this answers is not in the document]
lucklove: I means that you have checked this in latter codes:

https://github.com/pingcap/tiup/pull/1231/files#diff-89b17f92d341c20020ab4246afa2eed55ab6aae2ce14541cfd40168e39b0c014R253
https://github.com/pingcap/tiup/pull/1231/files#diff-89b17f92d341c20020ab4246afa2eed55ab6aae2ce14541cfd40168e39b0c014R262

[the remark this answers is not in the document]
9547: No, the check is necessary. 

The previous check is `A or B`, 
because they all need to invoke `operator.GetServiceStatus`,  
and later for condition A I'm trying to get the `status`(if not this condition, don't need to rewrite `status`), and for B to get the `since`.

lucklove:
It's better to hide `Since` column when the user doesn't specify `--uptime`
屏幕快照 2021-03-26 下午4 06 22

lucklove:
It seems the format is different when the data source different from metric and systemd.
屏幕快照 2021-03-26 下午4 06 33

9547:
> It seems the format is not consistent when the data source different from metric and systemd.
> 屏幕快照 2021-03-26 下午4 06 33
> 
> It's better to make it consistent:
> 
> * Maybe we can parse the string returned by systemd to int and format it again
> * Or we can use the same format with systemd
> 
> @9547 @AndreMouche What do you think about it?

Agree with it, keep the format consistent, I'll try to parse the output of systemd's output. 

+1 to hide `Since` column if `--uptime` not specified.

PR #1235, opened 2021-03-22
dm: support monitored config

### LINKED ISSUE #1233: tiup dm not support deploy “monitored”
## Bug Report

1. What did you do?

yaml 

```
monitored:   // line: 7
  node_exporter_port: 39100
  blackbox_exporter_port: 39115
```

tiup dm deploy xx.yaml

```
Error: Failed to parse topology file bigdm.yaml (topology.parse_failed)
  caused by: yaml: unmarshal errors:
  line 7: field monitored not found in type spec.topology
```

**TiDB & DM is not on the same host ，Unable to view DM host resource usage。**

```
v1.3.2 tiup
Go Version: go1.13
Git Branch: release-1.3
GitHash: 2d88460
```

9547:

fix #1233 

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/components/dm/spec/bindversion.go
+++ b/components/dm/spec/bindversion.go
@@ -8,7 +8,9 @@ func DMComponentVersion(comp, version string) string {
 	switch comp {
 	case spec.ComponentAlertmanager,
 		spec.ComponentGrafana,
-		spec.ComponentPrometheus:
+		spec.ComponentPrometheus,
+		spec.ComponentBlackboxExporter,
+		spec.ComponentNodeExporter:
 		return ""
 	default:
 		return version
--- a/components/dm/spec/logic.go
+++ b/components/dm/spec/logic.go
@@ -294,7 +294,7 @@ func (topo *Specification) GetGlobalOptions() spec.GlobalOptions {
 
 // GetMonitoredOptions returns MonitoredOptions
 func (topo *Specification) GetMonitoredOptions() *spec.MonitoredOptions {
-	return &topo.MonitoredOptions
+	return topo.MonitoredOptions
 }
 
 // ComponentsByStopOrder return component in the order need to stop.
--- a/components/dm/spec/logic.go
+++ b/components/dm/spec/logic.go
@@ -294,7 +294,7 @@ func (topo *Specification) GetGlobalOptions() spec.GlobalOptions {
 
 // GetMonitoredOptions returns MonitoredOptions
 func (topo *Specification) GetMonitoredOptions() *spec.MonitoredOptions {
-	return nil
+	return &topo.MonitoredOptions
 }
 
 // ComponentsByStopOrder return component in the order need to stop.
--- a/components/dm/spec/topology_dm.go
+++ b/components/dm/spec/topology_dm.go
@@ -35,6 +35,7 @@ const (
 
 var (
 	globalOptionTypeName  = reflect.TypeOf(GlobalOptions{}).Name()
+	monitorOptionTypeName = reflect.TypeOf(MonitoredOptions{}).Name()
 	serverConfigsTypeName = reflect.TypeOf(DMServerConfigs{}).Name()
 )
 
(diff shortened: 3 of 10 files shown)


AstroProfundis (approved):

AstroProfundis:
The tests are failing, I think we could use `v2.0.1` for DM instead of nightly.

9547:
> The tests are failing, I think we could use `v2.0.1` for DM instead of nightly.

It seems that the DM version is not well bound to black_box and node_export, even `v2.0.1`

image

lucklove:
> > The tests are failing, I think we could use `v2.0.1` for DM instead of nightly.
> 
> It seems that the DM version is not well bound to black_box and node_export, even `v2.0.1`
> 
> image

You should put node_expoter and black_expoter here: https://github.com/pingcap/tiup/blob/master/components/dm/spec/bindversion.go#L12

Like https://github.com/pingcap/tiup/blob/master/pkg/cluster/spec/bindversion.go#L23

So that tiup-dm can detect what's the correct xxx_expoter version (instead of use the cluster version)

lucklove:
Should deal with https://github.com/pingcap/tiup/blob/master/embed/templates/config/prometheus.yml.tpl#L17

DM doesn't has `node.rules.yml`, `blacker.rules.yml` and `bypass.rules.yml` in released versions. So maybe we should only config these files in TiDB context.

```
{{- if .MonitoredServers}}{{if .PDAddrs}}
  - 'node.rules.yml'
  - 'blacker.rules.yml'
  - 'bypass.rules.yml'
{{end}}{{- end}}
```

lucklove:
@buggithubs This requirement is strange, grafana dashboard for DM cluster doesn't has exporter panel at all.

lucklove:
I'm sorry, 9547. I have confirmed that we don't have a grafana panel for node exporter in DM cluster yet. It may take a little time to design one.

At this time, it's better implement this as a hidden feature: If the user doesn't configure `monitored` field, we just keep the old behavior and don't deploy node exporter and blackbox exporter. To archive this, we should implments [MonitoredOptions](https://github.com/pingcap/tiup/blob/master/components/dm/spec/topology_dm.go#L89) as a pointer so that if the user doesn't specify it, it will be nil.

The advantage of this scheme is that, some users deploy DM cluster overlaps TiDB cluster, some the the machines are shared between DM and TiDB. If DM and TiDB both deploy monitors by default, there will be conflict and the user may not be aware of that. When the user destroy one of the cluster, another cluster's exporters are destroyed, too. When I make users configure DM monitors explicitly, they will think about what these config means.

lucklove:
I try to implement this scheme but get into trouble, I'm not sure what happened at this time:

屏幕快照 2021-03-25 下午5 29 02

I use the default setting and the conflict detected. https://github.com/lucklove/tiup/commit/36a5b3b390224ab1d45536fd3042f937fbef9815

9547:
> I try to implement this scheme but get into trouble, I'm not sure what happened at this time:
> 
> 屏幕快照 2021-03-25 下午5 29 02
> 
> I use the default setting and the conflict detected. [lucklove@36a5b3b](https://github.com/lucklove/tiup/commit/36a5b3b390224ab1d45536fd3042f937fbef9815)

I've cherry-picked your commit, why the conflict detected is the default config not updated. The root cause is this line https://github.com/lucklove/tiup/blob/36a5b3b390224ab1d45536fd3042f937fbef9815/components/dm/spec/topology_dm.go#L38

```go
	monitorOptionTypeName = reflect.TypeOf(&MonitoredOptions{}).Name()
```

the ponitor's reflact-type name is empty, so all the fields will be skipped in this line https://github.com/lucklove/tiup/blob/36a5b3b390224ab1d45536fd3042f937fbef9815/components/dm/spec/topology_dm.go#L697

PR #1243, opened 2021-03-25
tiup: fix -v/--version argument

AstroProfundis:
Not able to show components' version as the `-v/--version` was always been processed by `tiup`. It was introduced in #1158 

Not using cobra's builtin `Version` field, but define a custom version argument in `tiup`.

### Check List 

Tests 

Release notes:

```release-note
NONE
```

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -33,9 +33,10 @@ func init() {
 	cobra.EnableCommandSorting = false
 
 	var (
-		binary  string
-		binPath string
-		tag     string
+		binary       string
+		binPath      string
+		tag          string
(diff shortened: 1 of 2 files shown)


9547:
Another question that's a little bit related, when I'm running `tiup dm -v`, the output doesn't show dm, does it by design or not:

image

AstroProfundis:
> does it by design or not

That's not totally by design, but is the expected behaviour. The field is actually the `args[0]` of the command, if you invoke the `tiup-dm` binary directly you'll see the difference.

We have some tricks handling `args[0]` in `cluster` to show correctly help messages, but they're generally not essential so it could be ok to just leave them alone.

PR #1252, opened 2021-03-27
cluster: Type -> Rule in confirmTopology

9547:

IN `tiup cluster display` the topology's header is like below:

image

But in `tiup cluster scale-out` the topology's header is like below:

image

Let's be consistent

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/pkg/cluster/manager/manager.go
+++ b/pkg/cluster/manager/manager.go
@@ -87,7 +87,7 @@ func (m *Manager) confirmTopology(name, version string, topo spec.Topology, patc
 
 	clusterTable := [][]string{
 		// Header
-		{"Type", "Host", "Ports", "OS/Arch", "Directories"},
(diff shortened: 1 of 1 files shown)


pkg/cluster/manager/manager.go
```diff
@@ -87,7 +87,7 @@ func (m *Manager) confirmTopology(name, version string, topo spec.Topology, patc
 
 	clusterTable := [][]string{
 		// Header
-		{"Type", "Host", "Ports", "OS/Arch", "Directories"},
+		{"Rule", "Host", "Ports", "OS/Arch", "Directories"},
```
lucklove, line 90: Maybe `Role`?

9547: Sorry :sweat_smile:

PR #1253, opened 2021-03-27
cluster: reuse cluster.DeployOption and spec.PDScaleScript

9547:
### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/pkg/cluster/template/scripts/pd.go
+++ b/pkg/cluster/template/scripts/pd.go
@@ -149,3 +149,12 @@ func NewPDScaleScript(pdScript *PDScript) *PDScaleScript {
 func (c *PDScaleScript) Config() ([]byte, error) {
 	return c.configWithScript("run_pd_scale.sh.tpl")
 }
+
+// ConfigToFile write config content to specific path
+func (c *PDScaleScript) ConfigToFile(file string) error {
+	config, err := c.Config()
+	if err != nil {
+		return err
+	}
+	return os.WriteFile(file, config, 0755)
+}
--- a/pkg/cluster/template/scripts/pd.go
+++ b/pkg/cluster/template/scripts/pd.go
@@ -21,7 +21,6 @@ import (
 	"text/template"
 
 	"github.com/pingcap/tiup/embed"
-	"github.com/pingcap/tiup/pkg/logger/log"
 )
 
 // PDScript represent the data to generate pd config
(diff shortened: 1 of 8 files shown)


pkg/cluster/template/scripts/pd.go
```diff
@@ -142,58 +145,7 @@ func NewPDScaleScript(name, ip, deployDir, dataDir, logDir string) *PDScaleScrip
...
-}
-
-// ConfigToFile write config content to specific path
-func (c *PDScaleScript) ConfigToFile(file string) error {
-	config, err := c.Config()
-	if err != nil {
-		return err
-	}
-	return os.WriteFile(file, config, 0755)
+	return c.configWithScript("run_pd_scale.sh.tpl")
```
lucklove, line 150: I think this may not work as you expected since you removed `func (c *PDScaleScript) WithXXX`.

When we call WithXXX [in scale out stage](https://github.com/pingcap/tiup/blob/master/pkg/cluster/spec/pd.go#L269), the `WithXXX` will return a PDScript pointer because you removed `WithXXX` on `PDScaleScript` (it will fallback to PDScript's WithXXX).  Then when you call [ConfigToFile](https://github.com/pingcap/tiup/blob/master/pkg/cluster/spec/pd.go#L280) or `Config`, it will call on `PDScript`, so the script generated in scale-out stage is not correct.

9547: Thanks for your professional advice 👍

9547: PTAL

lucklove: Still not work because PDScaleScript not implements [ConfigToFile](https://github.com/pingcap/tiup/pull/1253/files#diff-52ed75eb339663407c56ff6cf1b6bdf3a0995e7e9da391e4d28f113c9f14eb39R281)

9547: Sorry, please take another look.

AstroProfundis:
/hold

The github actions seems to be very slow now, we'll hold all PRs until the `v1.4.0` is successfully released.

Issue #1260, opened 2021-03-31
enable_binlog not imported from ansible inventory


lucklove:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?

Deploy an ansible cluster with [enable_binlog](https://github.com/pingcap/tidb-ansible/blob/master/inventory.ini#L85) = True

And execute `tiup cluster import` in the ansible directory.

2. What did you expect to see?
The cluster should be imported and when I execute `tiup cluster edit-config xxx`, I should see a config `binlog.enable = true` in TiDB's config.

3. What did you see instead?
The cluster import success, but the config `binlog.enable` not present.

4. What version of TiUP are you using (`tiup --version`)?
v1.4.0

PR #1261, opened 2021-03-31
Fix the issue that binlog_enable not import from ansible

lucklove:
Fix https://github.com/pingcap/tiup/issues/1260

Root cause: The [config](https://github.com/pingcap/tiup/blob/master/pkg/cluster/ansible/import.go#L110) was overwrite by latter [global compute](https://github.com/pingcap/tiup/blob/master/pkg/cluster/ansible/import.go#L196)


AstroProfundis:
liniting failed

AstroProfundis (approved):

Issue #1187, opened 2021-03-05
TiUP should serially scale in and scale out PD


HunDunDM:
## Feature Request

**Is your feature request related to a problem? Please describe:**

* Changing the number of PDs needs to follow the raft process, and only one can be changed at a time. Otherwise, the expanded PD may fail to start.
* 
![image](https://user-images.githubusercontent.com/19789302/110058407-3821aa80-7d9d-11eb-9626-f2744294bbe3.png)

**Describe the feature you'd like:**

* TiUP should serially scale in and scale out PD

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

* Go

9547:
@HunDunDM Could you please provide the instruction and failed logs in detail?

9547:
I've reproduced this issue. My originally Topology has 3 PDs(all nodes are healthy), and the scale-out topology as below:


```yaml
pd_servers:
  - host: n6
  - host: n7
  - host: n8
  - host: n9
  - host: n10
  - host: n11
  - host: n12
  - host: n13
  - host: n14
  - host: n15
  - host: n16
  - host: n17
  - host: n18
  - host: n19
  - host: n20
``` 

The `scale-out` failed, message as below:

![image](https://user-images.githubusercontent.com/29431502/112712561-6c573980-8f0b-11eb-9a50-3a805e05f5ac.png)


the error log in n9:2379's _pd.log_:

```txt
[2021/03/27 06:42:07.492 +00:00] [FATAL] [main.go:93] ["join meet error"] [error="there is a member that has not joined successfully"] [stack="github.com/pingcap/log.Fatal\n\t/home/jenkins/agent/workspace/build_pd_multi_branch_v4.0.4/go/pkg/mod/github.com/pingcap/log@v0.0.0-20200511115504-543df19646ad/global.go:59\nmain.main\n\t/home/jenkins/agent/workspace/build_pd_multi_branch_v4.0.4/go/src/github.com/pingcap/pd/cmd/pd-server/main.go:93\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:203"]
```

The `cluster display`:

![image](https://user-images.githubusercontent.com/29431502/112712581-96a8f700-8f0b-11eb-8249-c572748add99.png)

2/15 nodes scale-out succeed.

PR #1262, opened 2021-03-31
cluster: start pd,dm-master in sequentially

9547:

implement #1187 

Start instances in serial for Raft related components
eg: PD has more strict restrictions on the capacity expansion process, that is, there should be only one node in the peer-join stage at most.

Ref https://github.com/tikv/pd/blob/d38b36714ccee70480c39e07126e3456b5fb292d/server/join/join.go#L179-L191

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```


9547:
@HunDunDM  @lucklove  PTAL

Issue #1266, opened 2021-04-01
tiup-dm import failed after upgrade to 1.4.0


GMHDBJD:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?

deploy dm v1.0

inventory.ini 
```
## DM modules
[dm_master_servers]
dm_master ansible_host=172.28.0.101

[dm_worker_servers]
dm-worker1 ansible_host=172.28.0.111  server_id=101 source_id="mysql-replica-01" mysql_host=mysql1 mysql_user=root mysql_password='' mysql_port=3306

dm-worker2 ansible_host=172.28.0.112  server_id=101 source_id="mysql-replica-02" mysql_host=mysql2 mysql_user=root mysql_password='' mysql_port=3306

## Monitoring modules
[prometheus_servers]
prometheus ansible_host=172.28.0.121

[grafana_servers]
grafana ansible_host=172.28.0.121

[alertmanager_servers]
alertmanager ansible_host=172.28.0.121

## Global variables
[all:vars]
cluster_name = dm-v1

ansible_user = tidb

dm_version = v1.0.6

deploy_dir = /home/tidb/deploy

grafana_admin_user = "admin"
grafana_admin_password = "admin"
```

tiup dm import --yes --dir=/home/tidb/dm-ansible --cluster-version v2.0.1
2. What did you expect to see?
upgrade successfully

3. What did you see instead?
```
## DM modules
[dm_master_servers]
dm_master ansible_host=172.28.0.101

[dm_worker_servers]
dm-worker1 ansible_host=172.28.0.111  server_id=101 source_id="mysql-replica-01" mysql_host=mysql1 mysql_user=root mysql_password='' mysql_port=3306

dm-worker2 ansible_host=172.28.0.112  server_id=101 source_id="mysql-replica-02" mysql_host=mysql2 mysql_user=root mysql_password='' mysql_port=3306

## Monitoring modules
[prometheus_servers]
prometheus ansible_host=172.28.0.121

[grafana_servers]
grafana ansible_host=172.28.0.121

[alertmanager_servers]
alertmanager ansible_host=172.28.0.121

## Global variables
[all:vars]
cluster_name = dm-v1

ansible_user = tidb

dm_version = v1.0.6

deploy_dir = /home/tidb/deploy

grafana_admin_user = "admin"
grafana_admin_password = "admin"
```

4. What version of TiUP are you using (`tiup --version`)?

```
1.4.0 tiup
[...]

PR #1273, opened 2021-04-01
Fix the dir check logic on imported cluster

lucklove:
When scaling cluster, tiup will check if there are directories that overlaps each other to avoid conflict. The directories include both imported an new deployed ones.

However, the components deployed by ansible will use the same directory for saving log, which is by designed. At this time, TiUP will report an error if the imported cluster share the same host with different components when you try to scale the cluster.

If the directories conflict are both imported, skip instead of report an error.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```


AstroProfundis:
This seems related to #1266

AstroProfundis (approved):

PR #1285, opened 2021-04-06
cluster: speed up start/enable monitor instances

9547:

In the current implementation, node_exporter and blackbox_exporter may be invoked multiple times serially during a single `tiup cluster {start|stop|enable}` processing. 
Instead，those monitor components will be concurrently executed after all components are executed.

The process time of `tiup cluster enable` compared to old's  as follows:

```bash
# old 
Enabled cluster `c1` successfully

real    0m13.318s
user    0m2.228s
sys     0m0.319s

# new 
Enabled cluster `c1` successfully

real    0m5.687s
user    0m2.176s
sys     0m0.244s
```

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/pkg/cluster/operation/action.go
+++ b/pkg/cluster/operation/action.go
@@ -60,7 +60,7 @@ func Enable(
 		for _, inst := range insts {
 			instCount[inst.GetHost()]--
 			if instCount[inst.GetHost()] == 0 {
-				if err := EnableMonitored(ctx, inst, monitoredOptions, options.OptTimeout, isEnable); err != nil {
+				if err := EnableMonitored(ctx, inst.GetHost(), monitoredOptions, options.OptTimeout, isEnable); err != nil {
 					return err
 				}
 			}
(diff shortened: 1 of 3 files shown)


pkg/cluster/operation/action.go
```diff
@@ -191,77 +223,80 @@ func Restart(
...
+				}
 
-		// Check ready.
-		if err := spec.PortStarted(ctx, e, ports[comp], timeout); err != nil {
-			return toFailedActionError(err, "start", instance)
-		}
+				var err error
+				switch action {
+				case "start":
+					err = spec.PortStarted(ctx, e, ports[comp], timeout)
```
lucklove, line 266: Use nctx otherwise the test will fail.

See https://github.com/pingcap/tiup/blob/master/doc/design/checkpoint.md#parallel-task

9547: Thanks, PTAL

PR #1298, opened 2021-04-09
Send 'Staring component...' output to stderr

dveeden:

Some components output JSON or YAML. When the "Starting component..." message is mixed with this
the output needs to be cleaned up before it can be used.

Examples of this are:

```
tiup ctl:v5.0.0 tidb --host 192.168.122.50 schema in mysql | jq
```

```
tiup cluster template > template.yaml
```

This commit changes the output o the "Starting component..." to be send to stderr instead.

### Check List 

Tests 

Run the command with `> /dev/null`, you should now only see the `Starting component...` line.

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/pkg/exec/run.go
+++ b/pkg/exec/run.go
@@ -156,7 +156,7 @@ func PrepareCommand(p *PrepareCommandParams) (*exec.Cmd, error) {
 			return nil, err
 		}
 		if semver.Compare(selectVer.String(), latestV.String()) < 0 {
-			fmt.Println(color.YellowString(`Found %[1]s newer version:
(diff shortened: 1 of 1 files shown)


lucklove:
Could you please fix [this](https://github.com/pingcap/tiup/blob/master/pkg/exec/run.go#L159) too?

When the cluster/dm is not the newest, it will print this to stdout:

屏幕快照 2021-04-12 上午11 32 49

Which will cause problem too.

dveeden:
@lucklove that should be fixed as well now

AstroProfundis (approved):

PR #1325, opened 2021-04-21
playground: support semver constraint

unbyte:

Playground is used to quickly start up a local tidb cluster for developing or testing. If it supports version constraint, it will be more convenient. For example, it can replace the combination of `tiup list` and `tiup playground` for starting up a cluster of the latest version of a major version.

implement basic semver constraint `*` `x` `^` `~` just like how https://semver.npmjs.com/ calculates, and support short form of version

In all versions that satisfies version constraint, playground will find the latest one to start up.

*`v` can be omitted*

e.g.

`5` = `5.0` =  `5.0.0`

`5.*` = `5.x` = `5.*.*` = `5.x.x` = `^5` = `^5.0` = `^5.0.0`

`5.0.x` = `5.0.*` = `~5` = `~5.0` = `~5.0.0`

prereleases always be ignored unless use constraint like `^5.0.0-rc` or `~5.0.0-rc` or `5.0.0-rc`

preview:

image

image

image

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

Release notes:

```release-note
Support semver constraint for Playground component
```

### DIFF
--- a/cmd/mirror.go
+++ b/cmd/mirror.go
@@ -341,7 +341,7 @@ func newMirrorRotateCmd() *cobra.Command {
 }
 
 func editLatestRootManifest() (*v1manifest.Root, error) {
-	root, err := environment.GlobalEnv().V1Repository().FetchRootManfiest()
+	root, err := environment.GlobalEnv().V1Repository().FetchRootManifest()
 	if err != nil {
 		return nil, err
 	}
--- a/components/playground/playground.go
+++ b/components/playground/playground.go
@@ -37,6 +37,7 @@ import (
 	"github.com/pingcap/tiup/components/playground/instance"
 	"github.com/pingcap/tiup/pkg/cliutil/progress"
 	"github.com/pingcap/tiup/pkg/cluster/api"
+	"github.com/pingcap/tiup/pkg/cluster/spec"
 	"github.com/pingcap/tiup/pkg/environment"
 	"github.com/pingcap/tiup/pkg/utils"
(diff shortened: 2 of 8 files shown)


AstroProfundis:
Um, why not use `"golang.org/x/mod/semver"` as we do in many other files?

unbyte:
@AstroProfundis 
This PR doesn't remove `golang.org/x/mod/semver` but provide version constraint

AstroProfundis:
I'm not sure if we should add this to `playground` as this may make the version numbers confusing to users not so familiar with coding.

Another question is, do we need to add the same to `cluster` if we add it to `playground`?

unbyte:
IMO `playground` is something just like its name that we can play with (deving and testing), so semver constraint can help make this process more convenient.

Suppose I need to test whether a tool for tidb works on the new version when tidb is still on v5.0.0-20210329, I can write the test script once using `tiup playground "~5-0"` then I no more need to pay attention to what is the latest rc version of tidb.

`cluster` is imore formal than `playground`, I don't think it's appropriate to use a version range when starting a cluster

AstroProfundis:
>  `playground` is something just like its name that we can play with

Agree.

> I don't think it's appropriate to use a version range when starting a cluster

Agree.

So my concern is that we're making `playground` and `cluster` having different user experience, this might lead to some confusions. But also, it may just not any big deal at all. What do you think @lucklove ?

PR #1327, opened 2021-04-23
telemetry: fix tememetry data and update reported fields

AstroProfundis:
- Fix wrong "cluster UUID" used for telemetry reports for clusters, use hashed cluster name instead
- Add "installation UUID" of the tiup instance to identify unique users
- Add "telemetry secret" to use as salt when hashing field values before uploading telemetry info
- Include tiup and tiup-cluster build info in the telemetry report
- Hash more critical field values, hash everything with the "telemetry secret" as salt so that same values used by different users produces different hashes

### Check List 

Tests 

 - Unit test

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
telemetry: fix tememetry data and update reported fields
```


pkg/telemetry/report.proto
```diff
@@ -4,16 +4,18 @@ package telemetry;
 import "gogoproto/gogo.proto";
 
 message Report {
```
lucklove, line 6: It's not a good practice to modify existed field in protobuf. 
Make sure it's compatible with old servers/clients.

pkg/telemetry/scrub.go
```diff
@@ -58,27 +73,37 @@ func scrupMap(val interface{}, hashFieldNames map[string]struct{}, hash bool) in
...
+				scrupMap(rv.Index(i).Interface(), hashFieldNames, omitFieldNames, hash, omit, salt))
 		}
 		return ret
-	}
-
-	if rv.Kind() == reflect.String {
+	case reflect.Map,
+		reflect.Struct,
+		reflect.Interface:
+		return scrupMap(val, hashFieldNames, omitFieldNames, hash, omit, salt)
```
lucklove, line 95: Dead recursive

pkg/telemetry/report.proto
AstroProfundis: We didin't use protobuf at all, it was only used to generate the structs definitions, and the actual data send to telemetry server is using `json.Marshal()` with those structs.

PR #1348, opened 2021-05-06
install.sh: check depends before downloading package

### LINKED ISSUE #1287: tiup does not check if curl/tar is available
## Bug Report

1. What did you do?

From https://github.com/pingcap/docs/pull/5228

On manual inspection it does indeed use tar:
https://github.com/pingcap/tiup/blob/52e24c2ee45e6a2acc3bd59990a4f1ebbedcd633/install.sh#L38

It would be nice if the script checked for dependencies (curl, tar) correctly before downloading, and then if it bails print the docs URL for help. i.e. https://docs.pingcap.com/tidb/stable/quick-start-with-tidb

3. What did you see instead?

The error message is only presented after the download.

Confirmed by inspecting code against master.

AstroProfundis:
Fix #1287 

### Check List 

Tests 

Related changes

Release notes:

```release-note
install.sh: check depends before downloading package
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -43,11 +43,11 @@ install_binary() {
 check_depends() {
     pass=0
     command -v curl >/dev/null || {
-        echo "Dependency check failed: please install 'curl' before installing."
-	pass=1
+        echo "Dependency check failed: please install 'curl' before proceeding."
+        pass=1
     }
     command -v tar >/dev/null || {
-        echo "Dependency check failed: please install 'tar' before installing."
+        echo "Dependency check failed: please install 'tar' before proceeding."
         pass=1
     }
     return $pass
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -40,6 +40,23 @@ install_binary() {
     return 0
 }
 
+check_depends() {
+    pass=0
+    command -v curl >/dev/null || {
+        echo "Dependency check failed: please install 'curl' before installing."
+	pass=1
```
lucklove, line 47:
```suggestion
	  pass=1
```

install.sh
```diff
@@ -40,6 +40,23 @@ install_binary() {
     return 0
 }
```
lucklove, line 42: Add this logic to local_insta.sh too.

Issue #1355, opened 2021-05-10
Support json output


kolbe:
## Feature Request

**Is your feature request related to a problem? Please describe:**

It would be nice to be able to get information about clusters in a machine-readable format, such as json.

**Describe the feature you'd like:**

`tiup cluster display` (and maybe some other commands) should support a flag like `--format` or `--output` that can accept `json` (and maybe other options?) as an output format.

**Why the featue is needed:**

I want a machine-readable output format for cluster information so I can automatically configure my application to connect to the set of tidb instances in the cluster.

**Describe alternatives you've considered:**

Trying to parse the output of `tiup cluster display` in its current format is really unwieldy and unrealistic.

**Teachability, Documentation, Adoption, Migration Strategy:**

dveeden:
I think we can close this. We have #1208 for long term solutions. And #1358 for `tiup cluster display --json` which is for the specific usecase mentioned here.

AstroProfundis:
Thanks, please update in #1208 if any new idea comes.

Issue #1208, opened 2021-03-15
Refactor (console & log) output framework


AstroProfundis:
## Feature Request

**Is your feature request related to a problem? Please describe:**

At the moment, we use several different packages to output information (both to console and to audit log files), this makes it hard to maintain and change if we need to update the TUI.

**Describe the feature you'd like:**

Sort out all outputs we use now and migrate them to one, unified output framework. I'd prefer to refactor `pkg/logger` to a wrapped zap logger, and all outputs are ported to use the customized logger instead of pure zap logger nor `fmt`.

The refactor work itself should **NOT** change any current TUI much (except necessary minor adjustments), we'll do that in other PRs, to avoid this work being too complex.

**Teachability, Documentation, Adoption, Migration Strategy:**

Most of the codes related are in `cluster` and `dm`, and **many many** codes are using `fmt` to format console outputs. And we need to make sure things don't mess up when the messages has colored string in them.

I'd like to make `pkg/logger` the all-in-one output gateway of all components, and when the command is called, a multi target logger is initialized to output messages both to audit log and to console. They may be in different format (e.g., write JSON to audit log, and write formatted text to console), and the log level could be customized.

More details should be added once someone is willing to work on this and start to dig into the code.

PR #1358, opened 2021-05-11
cluster: add JSON output option for display

dveeden:

This adds JSON output for the `tiup cluster display` command. This makes it easier to parse the output and use it in other programs.

https://github.com/pingcap/tiup/issues/1355

This puts the info in a struct and uses encoding/json to format this as JSON

### Check List 

Tests 

`tiup cluster display mycluster`
`tiup cluster display --json mycluster`

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
Support JSON output for the `tiup cluster display` command
```

### DIFF
--- a/pkg/cluster/manager/display.go
+++ b/pkg/cluster/manager/display.go
@@ -15,6 +15,7 @@ package manager
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"math"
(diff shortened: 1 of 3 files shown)


AstroProfundis:
releated to #1208

pkg/cluster/manager/display.go
```diff
@@ -53,6 +54,18 @@ type InstInfo struct {
 	Port          int
 }
 
+type DashboardInfo struct {
```
AstroProfundis, line 57: should also have dashboard URL and TLS enabled flag.

pkg/cluster/manager/display.go
```diff
@@ -53,6 +54,23 @@ type InstInfo struct {
...
+	ClusterType    string `json:"cluster_type"`
+	ClusterName    string `json:"cluster_name"`
+	ClusterVersion string `json:"cluster_version"`
+	SSHType        string `json:"ssh_type"`
+	TLSEnabled     bool   `json:"tls_enabled"`
+	TLSCACert      string `json:"tls_ca_cert"`
+	TLSClientCert  string `json:"tls_client_cert"`
+	TLSClientKey   string `json:"tls_client_key"`
+	URL            string `json:"url"`
```
AstroProfundis, line 66: I think it's better be `dashboard_url`, as it's the URL of our builtin dashboard in PD, we may have other URLs or links to something else. (there is no such thing as a general "URL" for a TiDB cluster)

AstroProfundis (commented):
There seems to be an empty file with no filename added in the PR, and need to be removed if it's not by intend.

AstroProfundis (approved):

PR #1365, opened 2021-05-15
playground: implement `mode` option for painter algorithm populating options

### LINKED ISSUE #1269: fruitful playground - tikv-slim mode
### Feature Description

So far, `tiup playground` by default create a tidb-fullstack cluster with 1 tidb + 1 tikv + 1 pd + 1 tiflash deployment. It is good for tidb audience, but not great for tikv-only audience, who just wants tikv + pd.

I'd like to introduce a [tikv quickstart guide](https://github.com/tikv/tikv#try-tikv) and recommend tiup as the defaulut, but it is an understanding burden for new user with current tiup he must set `--db 0 --tiflash 0` and likely asks "what db / tiflash is".

Thus I propose we introduce a tikv-slim mode for playground which builds on the top of instance settings and provides fruitful feature for tikv-only audience, with which the experience becomes

```
$ tiup playground --mode tikv-slim
```

... and it starts 1 tikv + 1 pd only.

### Rough Technical Design

The interface above leads to a minimal change set of code. Generally specking tiup currently hardcode the default option as below.

https://github.com/pingcap/tiup/blob/fa0a7ce14e067b1377b58079788d74ef693f82d6/components/playground/main.go#L79-L97

We can introduce a new config named `mode`(consider it!) which has options "tidb-fullstack"(default), and "tikv-slim". 

tidb-fullatsack is the default and acts as current, while tikv-slim will populate the default config with "tikv.Num =  1" and "pd.Num = 1" only, without setting up tidb & tiflash. Other configs will be respected later one by one.

tisonkun:

This PR closes #1269 

It introduces a `mode` option to switch default config set among `tidb` or `tikv-slim` and allow user populate more options based on the default set.

Now, we directly communicate with `FlagSet` and use `Changed` to see if user explicit sets a config flag.

And I implement the painter algorithm that

1. layer 0: draw the `version` config
2. layer 2: draw the default set based on `mode` option
3. layer 3: draw the reset options configured by user

each layer overwrite options set in early layer.

### Check List

Tests

I run

- `tiup playground`
- `tiup playground --mode tidb`
- `tiup playground --mode tikv-slim`
- `tiup playground --mode unknown`
- `tiup playground --db 0 --tiflash 0`
- `tiup playground --mode tikv-slim --db 1 --tiflash 1`

and all run as expected.

I'm willing to add some automatic test but help wanted. Also I think we can add tests after this pr merged instead of be too mean to a fruitful pull request.

Side effects

 - Increased code complexity

with increasing flexibility.

Related changes

Yes, we need to update the documentation. But where?

```release-note
Implement `mode` option for painter algorithm populating options. `tidb` as the original defaults and `tikv-slim` for only pd, tikv, and monitor instances.
```


components/playground/main.go
```diff
@@ -30,6 +30,8 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/spf13/pflag"
+
```
AstroProfundis, line 34: remove this empty line and `make fmt` again

AstroProfundis (commented):
You can use `pflag.Flag.Visit()` to determine if an argument is set by user, here's an example in `tiup mirror` sub command https://github.com/pingcap/tiup/blob/dd6b81842e5caa4392ff7dc1285d5b8e2dd0a096/cmd/mirror.go#L294

components/playground/main.go
```diff
@@ -266,6 +298,148 @@ Examples:
...
+	if err = populateStringOptIfChanged(flagSet, ticdcBinpath, &options.TiCDC.BinPath); err != nil {
+		return
+	}
+	if err = populateStringOptIfChanged(flagSet, pumpBinpath, &options.Pump.BinPath); err != nil {
+		return
+	}
+	if err = populateStringOptIfChanged(flagSet, drainerBinpath, &options.Drainer.BinPath); err != nil {
+		return
+	}
+	return
```
lucklove, line 410: Can we use a loop to deal with this template code?

tisonkun: how can we achieve that? the problem is keys(`ticdcBinpath`, `pdConfig`, ..) and containers(`&options.TiCDC.BinPath`, `&options.PD.ConfigPath`) all names. And there is no generic. So we may locally do the grouping and for a local lists, I don't think it is a much more clever way.

components/playground/main.go
tisonkun: done

tisonkun:
/cc @AstroProfundis @lucklove 

I've updated the implementation as @AstroProfundis suggested, using `Visit` method. The PR is ready for review.

components/playground/main.go
tisonkun: Please see if the latest refactor as you want.

components/playground/main.go
```diff
@@ -226,38 +253,42 @@ Examples:
...
-
-	rootCmd.Flags().StringVarP(&opt.TiDB.BinPath, "db.binpath", "", opt.TiDB.BinPath, "TiDB instance binary path")
-	rootCmd.Flags().StringVarP(&opt.TiKV.BinPath, "kv.binpath", "", opt.TiKV.BinPath, "TiKV instance binary path")
-	rootCmd.Flags().StringVarP(&opt.PD.BinPath, "pd.binpath", "", opt.PD.BinPath, "PD instance binary path")
-	rootCmd.Flags().StringVarP(&opt.TiFlash.BinPath, "tiflash.binpath", "", opt.TiFlash.BinPath, "TiFlash instance binary path")
-	rootCmd.Flags().StringVarP(&opt.TiCDC.BinPath, "ticdc.binpath", "", opt.TiCDC.BinPath, "TiCDC instance binary path")
-	rootCmd.Flags().StringVarP(&opt.Pump.BinPath, "pump.binpath", "", opt.Pump.BinPath, "Pump instance binary path")
-	rootCmd.Flags().StringVarP(&opt.Drainer.BinPath, "drainer.binpath", "", opt.Drainer.BinPath, "Drainer instance binary path")
+	defaultMode := "tidb"
+	defaultOptions := &BootOptions{}
```
AstroProfundis, line 257: umm, seems `defaultOptions` is not used anywhere?

tisonkun: Just used at

https://github.com/pingcap/tiup/pull/1365/files#diff-e14447fade4bdc6657816fd192d636c69c8e40eb38a9f9c3668595b8ad40661eR262-R291

AstroProfundis: But where did those values used? I'm wondering if using `options` could do the same.

tisonkun: It is used for satisfy the default value parameter that `cobra` asks for.

Technically we can use `options` but better to avoid too much concept override. Here is a use for default value, and we use an all blank options object, it won't be used anywhere else and we prevent from error-prone cases.

components/playground/main.go
```diff
@@ -226,38 +253,42 @@ Examples:
...
+
+	rootCmd.Flags().String(mode, defaultMode, "TiUP playground mode: 'tidb', 'tikv-slim")
+	rootCmd.Flags().Bool(withMonitor, false, "Start prometheus and grafana component")
+
+	rootCmd.Flags().Int(db, defaultOptions.TiDB.Num, "TiDB instance number")
```
lucklove, line 262: What's the difference between pass a zero value and `defaultOptions.XXX` since the `defaultOptions.XXX` is a zero value?

tisonkun: There is no different but I'd like to avoid magic literal. In this case I'm ok to use zero value if you ask for it.

lucklove: Another problem is that, by default, we should bootstrap 1 TiDB node, but in this implement, it will show zero in `tiup playground --help`

tisonkun: @lucklove you're right that on the default value changed. But zero value isn't shown on usage.

```
tiup playground --help
Bootstrap a TiDB cluster in your local host, the latest release version will be chosen
if you don't specified a version.

Examples:
  $ tiup playground nightly                         # Start a TiDB nightly version local cluster
  $ tiup playground v5.0.1 --db 3 --pd 3 --kv 3     # Start a local cluster with 10 nodes
  $ tiup playground nightly --monitor=false         # Start a local cluster and disable monitor system
  $ tiup playground --pd.config ~/config/pd.toml    # Start a local cluster with specified configuration file
  $ tiup playground --db.binpath /xx/tidb-server    # Start a local cluster with component binary path
  $ tiup playground --kv-mode --pd 3 --kv 3 		# Start a local cluster in KV mode (No TiDB Available)

Usage:
  tiup playground [version] [flags]
  tiup [command]

Available Commands:
  display     
  help        Help about any command
  scale-in    
  scale-out   

Flags:
      --db int                   TiDB instance number
      --db.Host host             Playground TiDB host. If not provided, TiDB will still use host flag as its host
      --db.binpath string        TiDB instance binary path
      --db.config string         TiDB instance configuration file
      --db.timeout int           TiDB max wait time in seconds for starting, 0 means no limit
      --drainer int              Drainer instance number
      --drainer.binpath string   Drainer instance binary path
      --drainer.config string    Drainer instance configuration file
  -h, --help                     help for tiup
      --host string              Playground cluster host
      --kv int                   TiKV instance number
      --kv.binpath string        TiKV instance binary path
      --kv.config string         TiKV instance configuration file
      --mode string              TiUP playground mode: 'tidb', 'tikv-slim (default "tidb")
      --monitor                  Start prometheus and grafana component
      --pd int                   PD instance number
      --pd.Host host             Playground PD host. If not provided, PD will still use host flag as its host
      --pd.binpath string        PD instance binary path
      --pd.config string         PD instance configuration file
      --pump int                 Pump instance number
      --pump.binpath string      Pump instance binary path
      --pump.config string       Pump instance configuration file
      --ticdc int                TiCDC instance number
      --ticdc.binpath string     TiCDC instance binary path
      --ticdc.config string      TiCDC instance configuration file
      --tiflash int              TiFlash instance number
      --tiflash.binpath string   TiFlash instance binary path
      --tiflash.config string    TiFlash instance configuration file
      --tiflash.timeout int      TiFlash max wait time in seconds for starting, 0 means no limit
  -v, --version                  version for tiup

Use "tiup [command] --help" for more information about a command.
```

I think it is a draw back we cannot fully customize default set but it doesn't hurt so much.

... and I'll update the `kv-mode` information.

lucklove: lgtm

AstroProfundis (approved):

andylokandy:
Does this PR need to cherry-pick?

tisonkun:
> Does this PR need to cherry-pick?

I have no idea but it just works for me using a nightly version.

andylokandy:
I think this should better be released.

lucklove:
@andylokandy Will be released in [v1.5.0](https://github.com/pingcap/tiup/milestone/11)

andylokandy:
ok, i see. The release is spilt from master.

PR #1372, opened 2021-05-19
support ticdc data dir

JinLingChristopher:

For TiCDC, it might create a `sort-dir` to store temporary files for sort. In this PR, add `data-dir` to TiCDC, which can be used to store any temporary files generated by TiCDC instance.

* add `DataDir` field at `CDCSpec`

There are two scenarios:

1. for `version < v4.0.13`, does not show data-dir in confirm information. even user set data-dir, it does not work.
2. for `version >= v4.0.13 && version != v5.0.0-RC`, show data-dir, and create it, and start CDC instance by using it.

### Check List 

Tests 

 - No code

Code changes

Side effects

Related changes

Release notes:

```release-note
NONE
```
NONE


pkg/cluster/manager/deploy.go
```diff
@@ -245,6 +245,10 @@ func (m *Manager) Deploy(
 		deployDir := spec.Abs(globalOptions.User, inst.DeployDir())
 		// data dir would be empty for components which don't need it
 		dataDirs := spec.MultiDirAbs(globalOptions.User, inst.DataDir())
+
+		log.Debugf("inst.ComponentName=%+v, inst.DataDir=%+v, dataDirs=%+v\n",
```
AstroProfundis, line 249: Did you test how it behaviors when using in real world, would it broke the dynamic TUI output?

overvenus (commented):
Could you add some tests?

E.g., tests like 

https://github.com/pingcap/tiup/blob/2faf4c934f3390bef4b2c1ae1884ec4a8cb9dc62/pkg/cluster/spec/spec_test.go#L35-L48

Rest LGTM

components/client/go.mod
```diff
@@ -5,6 +5,7 @@ go 1.16
 require (
 	github.com/StackExchange/wmi v0.0.0-20190523213315-cbe66965904d // indirect
 	github.com/gizak/termui/v3 v3.1.0
+	github.com/google/flatbuffers v2.0.0+incompatible // indirect
```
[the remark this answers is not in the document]
JinLingChristopher: this is caused by golang/x/tools. [problem](https://github.com/google/flatbuffers/issues/6466)

pkg/cluster/spec/cdc.go
```diff
@@ -35,6 +35,7 @@ type CDCSpec struct {
 	Patched         bool                   `yaml:"patched,omitempty"`
 	Port            int                    `yaml:"port" default:"8300"`
 	DeployDir       string                 `yaml:"deploy_dir,omitempty"`
+	DataDir         string                 `yaml:"data_dir,omitempty"`
```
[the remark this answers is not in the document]
JinLingChristopher: yes, it's the same logic like `tikv` or `pd`

pkg/cluster/spec/cdc.go
```diff
@@ -152,8 +154,10 @@ func (i *CDCInstance) InitConfig(
 	instanceConfig := spec.Config
 
 	configFileSupported := false
+	dataDirSupported := false
 	if semver.Compare(clusterVersion, "v4.0.13") >= 0 && clusterVersion != "v5.0.0-rc" {
 		configFileSupported = true
+		dataDirSupported = true
```
AstroProfundis, line 160: better use a dedicated if block, so that the error message can be specific to `data_dir`.

AstroProfundis: and did you make sure v5.0 series versions except `v5.0.0-rc` support the `--sort-dir` argument?

JinLingChristopher: yes, it's checked at source code level.

JinLingChristopher: yes, it's checked at the source code level.

JinLingChristopher: > better use a dedicated if block, so that the error message can be specific to `data_dir`.

I don't think we have to throw an error message on `data-dir`, the reason is described above.

overvenus:
For version >= v4.0.13 && version != v5.0.0-RC, can it report error when cdc's data-dir is not set?

pkg/cluster/spec/spec.go
```diff
@@ -376,6 +378,20 @@ func (s *Specification) GetPDList() []string {
 	return pdList
 }
 
+// AdjustByVersion modify the spec by cluster version.
+func (s *Specification) AdjustByVersion(clusterVersion string) {
+	// CDC does not support data dir for version below v4.0.13, and also v5.0.0-rc, set it to empty.
+	if semver.Compare(clusterVersion, "v4.0.13") == -1 || clusterVersion == "v5.0.0-rc" {
+		log.Warnf(color.RedString("[WARN] data_dir is only supported with TiCDC version v4.0.13 or later, "+
```
lucklove, line 385: It seems will report this warning once the version is lower than `v4.0.13`, no matter if the user configure the data directory.

JinLingChristopher:
@AstroProfundis @lucklove  @amyangfei @overvenus 

PTAL

PR #1382, opened 2021-05-24
cluster/executor: implement native SCP download instead of cat

AstroProfundis:
Use native SCP to download file from remote, instead of `cat` the file content.

Protocol doc: https://web.archive.org/web/20170215184048/https://blogs.oracle.com/janp/entry/how_the_scp_protocol_works
The implementation is partially inspired by https://github.com/dtylman/scp

### Check List 

Tests 

 - Integration test

Code changes

Release notes:

```release-note
cluster: adjust file transferring with remote server to use native SCP
```

### DIFF
--- a/pkg/cluster/executor/checkpoint.go
+++ b/pkg/cluster/executor/checkpoint.go
@@ -46,7 +46,6 @@ var (
 		checkpoint.Field("src", reflect.DeepEqual),
 		checkpoint.Field("dst", reflect.DeepEqual),
 		checkpoint.Field("download", reflect.DeepEqual),
-		checkpoint.Field("limit", reflect.DeepEqual),
 	)
 )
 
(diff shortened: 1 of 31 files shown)


pkg/cluster/executor/checkpoint.go
```diff
@@ -46,6 +46,7 @@ var (
 		checkpoint.Field("src", reflect.DeepEqual),
 		checkpoint.Field("dst", reflect.DeepEqual),
 		checkpoint.Field("download", reflect.DeepEqual),
+		checkpoint.Field("limit", reflect.DeepEqual),
```
lucklove, line 49: This is not needed.

pkg/cluster/executor/scp.go
```diff
@@ -0,0 +1,132 @@
...
+		// parse SCP command
+		line, _, err := bufr.ReadLine()
+		if err != nil {
+			return err
+		}
+		if line[0] != byte('C') {
+			return fmt.Errorf("incorrect scp command '%b', should be 'C'", line[0])
+		}
+
+		mode, err := strconv.ParseUint(string(line[1:5]), 0, 32)
```
lucklove, line 64: Move the creating file action here.

Issue #1397, opened 2021-06-03
tiup-cluster: auto-detect architecture


kolbe:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Why do we require administrators to manually specify the architecture for each host in the cluster definition?

**Describe the feature you'd like:**

We should auto-detect the architecture by default.

If an explicit architecture is given, we should use that architecture instead of auto-detecting (in case there's some reason to want the OS to emulate/translate execution).

**Why the featue is needed:**

It is much easier for TiUP to execute `uname -m` or the equivalent after connecting to each host than for the administrator to have to manually configure the architecture.

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

9547:
IMO, currently, we need to download the binary packages first, especially in the offline scenario, if the CPU architecture is not specified, then we need to download all the CPU arch packages, maybe huge and slow? On the other hand, I'm not sure whether or not with mixed deployments of different CPUs are currently supported(not related to this issue).

kolbe:
TiUP downloads all kinds of stuff "automatically" when some local package is missing, why not just do that at the time that it connects to each node during the `cluster deploy` phase?

PR #1423, opened 2021-06-12
cluster,dm:  auto detect cpu arch

9547:

Implement #1397 

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/pkg/cluster/manager/manager.go
+++ b/pkg/cluster/manager/manager.go
@@ -168,7 +168,7 @@ func (m *Manager) fillHostArch(s *tui.SSHConnectionProps, topo spec.Topology, gO
 		return nil
 	}
 
-	ctx := ctxt.New(context.Background())
+	ctx := ctxt.New(context.Background(), gOpt.Concurrency)
 	t := task.NewBuilder().
 		ParallelStep("+ Detect CPU Arch", false, detectTasks...).
 		Build()
--- a/pkg/cluster/manager/manager.go
+++ b/pkg/cluster/manager/manager.go
@@ -136,7 +136,7 @@ func (m *Manager) sshTaskBuilder(name string, topo spec.Topology, user string, o
 		ClusterSSH(topo, user, opts.SSHTimeout, opts.SSHType, topo.BaseTopo().GlobalOptions.SSHType)
 }
 
-func (m *Manager) fillHostArch(s *uti.SSHConnectionProps, topo spec.Topology, gOpt *operator.Options, user string) error {
+func (m *Manager) fillHostArch(s *tui.SSHConnectionProps, topo spec.Topology, gOpt *operator.Options, user string) error {
 	hostArch := map[string]string{}
 	var detectTasks []*task.StepDisplay
 	topo.IterInstance(func(inst spec.Instance) {
--- a/pkg/cluster/manager/manager.go
+++ b/pkg/cluster/manager/manager.go
@@ -14,12 +14,14 @@
 package manager
 
 import (
+	"context"
 	"fmt"
 	"strings"
 
 	"github.com/fatih/color"
 	"github.com/joomcode/errorx"
 	perrs "github.com/pingcap/errors"
+	"github.com/pingcap/tiup/pkg/cluster/ctxt"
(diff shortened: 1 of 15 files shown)


pkg/cluster/manager/manager.go
```diff
@@ -134,3 +136,54 @@ func (m *Manager) sshTaskBuilder(name string, topo spec.Topology, user string, o
 		).
 		ClusterSSH(topo, user, opts.SSHTimeout, opts.SSHType, topo.BaseTopo().GlobalOptions.SSHType)
 }
+
+func (m *Manager) fillHostArch(s *cliutil.SSHConnectionProps, topo spec.Topology, gOpt *operator.Options, user string) error {
+	hostArch := map[string]string{}
```
AstroProfundis, line 141: SSH port should also be part of the unique key, the SSH target might be a NAT device.

pkg/cluster/manager/manager.go
```diff
@@ -134,3 +136,54 @@ func (m *Manager) sshTaskBuilder(name string, topo spec.Topology, user string, o
...
+	t := task.NewBuilder().
+		ParallelStep("+ Detect CPU Arch", false, detectTasks...).
+		Build()
+
+	if err := t.Execute(ctx); err != nil {
+		return perrs.Annotate(err, "failed to fetch cpu arch")
+	}
+
+	for host := range hostArch {
+		stdout, _, ok := ctxt.GetInner(ctx).GetOutputs(host)
```
AstroProfundis, line 182: ... and so the result handling should be just after the probe, as there might be multiple job for one host, it's possible to append a `Func` task after `Shell` to do so.

pkg/cluster/spec/spec.go
```diff
@@ -71,7 +71,7 @@ type (
 		LogDir          string               `yaml:"log_dir,omitempty"`
 		ResourceControl meta.ResourceControl `yaml:"resource_control,omitempty" validate:"resource_control:editable"`
 		OS              string               `yaml:"os,omitempty" default:"linux"`
-		Arch            string               `yaml:"arch,omitempty" default:"amd64"`
+		Arch            string               `yaml:"arch,omitempty" default:""`
```
AstroProfundis, line 74: just remove the `default` tag and it will be zero value of the type.

AstroProfundis (commented):
What if user specified an arch different with `uname -m` reports? e.g., user sets `amd64` in the topology file, but the probe shows as `aarch64`. Will the topology changed by the probe result?

9547:
> What if user specified an arch different with `uname -m` reports? e.g., user sets `amd64` in the topology file, but the probe shows as `aarch64`. Will the topology changed by the probe result?

We only detect those nodes whose arch was not specified.

pkg/cluster/manager/manager.go
9547: When reviewing the rest of the code, I found that for the most part, we didn't handle the SSH Port case. 

eg. Shell's execute will set output by the host, if two nodes are under the same host, the later one's output will be overwritten. 

https://github.com/pingcap/tiup/blob/342f0448f9225ce8b87238319e93a659e1e71188/pkg/cluster/task/shell.go#L34-L53

pkg/cluster/manager/manager.go
9547: What do you mean is make this task along with the later ones, such as check, deploy, scale-out? If so, IMO, it is best to have as little impact on the existing business as possible, and the CPU detection is a pre-action that needs to be done before the main logic(check, deploy, scale-out) is executed.  So, here I've implemented this action in an independent operation.

AstroProfundis (approved):

pkg/cluster/manager/manager.go
AstroProfundis: Nope, I mean appending a `Func` task just after the `Shell` task that runs `uname`, but as we don't distinguish host by SSH port, it not necessary.

PR #1428, opened 2021-06-16
fix pd and dashboard address display issue in `tiup-playground`

iosmanthus:
Signed-off-by: iosmanthus <myosmanthustree@gmail.com>

In `tidb` mode, if the `tidb` starts then we can say the PDs must start. However, when in `tikv-slim` mode, there is an issue that `tiup-playground` won't print any information about PD, because it won't wait for any instance of PDs to really start.

1. Print dashboard's address only when there are some TiDB instances.
2. Always print PD's address.

### Check List 

Tests 

```sh
Starting component `playground`: /home/iosmanthus/.tiup/components/playground/v1.5.1/tiup-playground --mode tikv-slim
Using the version v5.0.2 for version constraint "".

If you'd like to use a TiDB version other than v5.0.2, cancel and retry with the following arguments:
    Specify version manually:   tiup playground <version>
    Specify version range:      tiup playground ^5
    The nightly version:        tiup playground nightly

Playground Bootstrapping...
Start pd instance
Start tikv instance
PD client endpoints: [127.0.0.1:2379]
To view the Prometheus: http://127.0.0.1:39625
To view the Grafana: http://127.0.0.1:3000
```

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
display PD client endpoints in playground
```


iosmanthus:
@AstroProfundis PTAL

AstroProfundis (approved):

PR #1464, opened 2021-06-30
clone_mirror: refuse to clone from self

dveeden:

When running something like this:
```
tiup mirror clone /tmp/tidb_mirror -a amd64 --prefix v5.1.0 --os linux
tiup mirror set /tmp/tidb_mirror
tiup mirror clone /tmp/tidb_mirror -a amd64 --prefix v5.1.0 --os linux
```
This results in metadata being changed/corrupted.

Then the next `tiup mirror clone ...` results in:
```
Error: read manifest from mirror(/tmp/tidb_mirror) failed: not enough signatures (0) for threshold 1 in timestamp.json
```

This can easily be fixed with `tiup mirror set ...`, but it is best to
avoid this. And also cloning from self is probably not what the user
wanted in the first place. So adding the source to the output to make
this easier to spot.

Related: https://tidbcommunity.slack.com/archives/C019S7HLP61/p1625059913130200

### Check List 

Tests 

Release notes:

```release-note
A clear error is given when trying to clone a mirror from itself
```

### DIFF
--- a/pkg/repository/clone_mirror.go
+++ b/pkg/repository/clone_mirror.go
@@ -52,7 +52,10 @@ func CloneMirror(repo *V1Repository,
 	targetDir string,
 	selectedVersions []string,
 	options CloneOptions) error {
-	fmt.Printf("Start to clone mirror, targetDir is %s, selectedVersions are [%s]\n", targetDir, strings.Join(selectedVersions, ","))
+	if strings.TrimRight(targetDir, "/") == strings.TrimRight(repo.Mirror().Source(), "/") {
+		return errors.Errorf("Refusing to clone from %s to %s", targetDir, repo.Mirror().Source())
+	}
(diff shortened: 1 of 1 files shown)


AstroProfundis (approved):

pkg/repository/clone_mirror.go
```diff
@@ -52,7 +52,10 @@ func CloneMirror(repo *V1Repository,
 	targetDir string,
 	selectedVersions []string,
 	options CloneOptions) error {
-	fmt.Printf("Start to clone mirror, targetDir is %s, selectedVersions are [%s]\n", targetDir, strings.Join(selectedVersions, ","))
+	if targetDir == repo.Mirror().Source() {
```
lucklove, line 55: Should we trim suffix `/` before compare? The targetDir may be `/path/to/dir/` and the `repo.Mirror().Source()` may be `/path/to/dir`.

dveeden: Yes indeed.

Issue #1465, opened 2021-07-01
Not accessing network when listing installed components


AstroProfundis:
## Feature Request

**Is your feature request related to a problem? Please describe:**
When listing installed components with `tiup list --installed`, the list is refreshed from repository and then filtering with installed ones, which is not quite reasonable, as installed components could actually be listed only with local cached manifests.

**Describe the feature you'd like:**
Do not access network when listing components with `--installed` argument set.

**Why the featue is needed:**
Reduce network access.

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

AstroProfundis:
Done in #1553

PR #1466, opened 2021-07-01
Update manifest errors in list component should not be returned, change to warning

c4pt0r:
Update manifest errors in `list` command should not be returned, change to warning

Reason:
1. If the manifest is already available locally, the local component
   should be listed, but output warning

2. If there is no local manifest, in case of update manifest error,
   init manifest will also fail and will not continue

https://github.com/pingcap/tiup/issues/1465

### DIFF
--- a/cmd/list.go
+++ b/cmd/list.go
@@ -15,6 +15,7 @@ package cmd
 
 import (
 	"fmt"
+	"os"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


AstroProfundis (approved):

AstroProfundis:
@c4pt0r Please sign the CLA

AstroProfundis (approved):

PR #1471, opened 2021-07-05
Fix TiKV config check false alarm when data_dir is depended on

tabokie:

When TiKV config check requires correct data_dir value, tiup config check will fail because it doesn't feed that information to TiKV.

Add extra data path parameter when doing TiKV config check.

### Check List 

Tests 

 - Unit test

Related changes

Release notes:

```release-note
NONE
```


pkg/cluster/spec/server_config.go
```diff
@@ -300,7 +300,7 @@ func checkConfig(ctx context.Context, e ctxt.Executor, componentName, clusterVer
 		if componentName == ComponentTiKV {
 			extra = `--pd=""`
 		}
-		cmd = fmt.Sprintf("%s --config-check --config=%s %s", binPath, configPath, extra)
+		cmd = fmt.Sprintf("%s --config-check --config=%s --data-dir=%s %s", binPath, binPath, paths.Data, extra)
```
lucklove: Please add this in extra because not every component has `data-dir`

AstroProfundis:
Is this argument working for all versions of TiKV on config check?

tabokie:
> Is this argument working for all versions of TiKV on config check?

I think so, we are using it here as well: https://github.com/pingcap/tiup/blob/2036470398dfe8f3d2158c7d372f6b5637a1646e/components/playground/instance/tikv.go#L69

AstroProfundis (approved):

tabokie:
Are the test failures related? I don't understand them😢

AstroProfundis:
> Are the test failures related? I don't understand themcry

Umm no, just retry and see...

PR #1477, opened 2021-07-08
cluster: print more clear warning when scale in with `--force`

### LINKED ISSUE #1453: Display should show the exact store state after executing scale in --force
## Bug Report

1. What did you do?

Run `scale in --force` for a TiKV

Show the real store state of that TiKV.

3. What did you see instead?

The TiKV disappears in the display list which may mislead the user to do some dangerous operation like clean up the store data. It will cause the unavailable of some regions.
Or we should warn users to prevent using --force AMAP.

master

AstroProfundis:
Close #1453 

Related changes

Release notes:

```release-note
NONE
```


lucklove (approved):
LGTM

PR #1483, opened 2021-07-11
Add display tikv labels feature

wentaojin:
tiup cluster display support display tikv labels features，improve yaml meta file TiKV Labels has poor readability and checkability

if -R/-N not set Or -R tikv，display all tikv labels
if -N set, display current machine deploy all tikv instance labels

image

image


pkg/cluster/api/pdapi.go
```diff
@@ -732,6 +733,42 @@ func (pc *PDClient) GetTiKVLabels() (map[string]map[string]string, error) {
 	return locationLabels, nil
 }
 
+// GetTiKVStoreStateAndLabels implements TiKVLabelProvider
+func (pc *PDClient) GetTiKVStoreStateAndLabels() ([]map[string]string, error) {
```
AstroProfundis, line 737: There is already a `GetLocationLabels()` function to query TiKV labels, and store statuses can be fetched with instance's `Status()` method. What's the purpose of adding this new function? Is it necessary?

pkg/cluster/manager/display.go
```diff
@@ -225,6 +225,124 @@ func (m *Manager) Display(name string, opt operator.Options) error {
 	return nil
 }
 
+// DisplayTiKVLabels display cluster tikv labels
+func (m *Manager) DisplayTiKVLabels(name string, opt operator.Options) error {
```
AstroProfundis, line 229: Seems you are referring a pretty old code base, the latest one on master has a JSON output mode available. Please rebase your changes onto master.

AstroProfundis (changes requested):

pkg/cluster/manager/display.go
wentaojin: Master JSON is other feature function，The two do not conflict

pkg/cluster/api/pdapi.go
wentaojin: origin Display function isn‘t very easy to change，so add new function and flag

pkg/cluster/manager/display.go
AstroProfundis: It is, the label displaying is not compatible with JSON output, so the `display --json` will be broken in some situation.

pkg/cluster/api/pdapi.go
AstroProfundis: So why adding a whole new display mode from the fist place? Why not adding them to the current output?

wentaojin: `GetLocationLabels()` function isn't macth my demand，I want get all machine【machine ip，not with port】 labels and region、leader etc info...

pkg/cluster/manager/display.go
wentaojin: em...，you meaning `display --labels --json` situation , It‘s broken ,  but it output json ，not very good output，because machine ip row is exist null and exist multi row

pkg/cluster/api/pdapi.go
AstroProfundis: But... again... why? Why would adding a whole new table with such information be necessary, and why adding them to the current display table would not work? What's the usage senario of this new panel?

pkg/cluster/manager/display.go
AstroProfundis: Yes, any new features should not break existing ones unless absolutely necessary.

wentaojin: It’s necessary. because yamlmeta file TiKV Labels has poor readability and checkability，I haven’t thought of a better processing method for the time being. The current look and output are friendly to operation and maintenance

pkg/cluster/api/pdapi.go
wentaojin: em... GetLocationLabels() state only not up and I do not confirm whether there are special usages elsewhere. If GetLocationLabels() is not special, can consider combining the two into one

AstroProfundis: `GetLocationLabels()` is used in `display.go` and `scale_out.go` in `pkg/cluster/manager`, they all used to check for labels. So I think it's safe to adjust it to fit the display usage. It could also avoid redundant requests to PD.

pkg/cluster/manager/display.go
AstroProfundis: I mean as `display` now supports JSON output, any change made to it should also support JSON output, otherwise the compatibility is compromised.

PR #1492, opened 2021-07-20
cluster: add ability to ignore monitor agent for instances

AstroProfundis:
If user wants to deploy instances from multiple clusters onto one host, the monitor agents from those clusters may conflict. This PR add an optional flag in instance's spec to indicate that monitor agent is not needed for this cluster on the host. Typically this means there is already an agent deployed by another cluster.

The rendering process of Prometheus config is not changed, so that if there is already an agent, monitoring metrics in Prometheus should be fine.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

 - Increased code complexity

Release notes:

```release-note
NONE
```

### DIFF
--- a/pkg/cluster/operation/destroy.go
+++ b/pkg/cluster/operation/destroy.go
@@ -92,9 +92,16 @@ func Destroy(
 func StopAndDestroyInstance(ctx context.Context, cluster spec.Topology, instance spec.Instance, options Options, destroyNode bool) error {
 	ignoreErr := options.Force
 	compName := instance.ComponentName()
+	noAgentHosts := set.NewStringSet()
+
+	cluster.IterInstance(func(inst spec.Instance) {
+		if inst.IgnoreMonitorAgent() {
+			noAgentHosts.Insert(inst.GetHost())
+		}
+	})
 
 	// just try to stop and destroy
-	if err := StopComponent(ctx, []spec.Instance{instance}, options.OptTimeout); err != nil {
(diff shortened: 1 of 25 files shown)


lucklove:
What if:

```yaml
pd_servers:
  - host: 172.16.5.138
    client_port: 2234
    peer_port: 2235
  - host: 172.16.5.138
    client_port: 2236
    peer_port: 2237
    ignore_exporter: true
```

pkg/cluster/operation/destroy.go
```diff
@@ -111,7 +111,7 @@ func StopAndDestroyInstance(ctx context.Context, cluster spec.Topology, instance
 		// monitoredOptions for dm cluster is nil
 		monitoredOptions := cluster.GetMonitoredOptions()
 
-		if monitoredOptions != nil {
+		if monitoredOptions != nil && !instance.IgnoreMonitorAgent() {
```
lucklove, line 114: What if:

```yaml
pd_servers:
  - host: 172.16.5.138
    client_port: 2234
    peer_port: 2235
  - host: 172.16.5.138
    client_port: 2236
    peer_port: 2237
    ignore_exporter: true
```

AstroProfundis: I've added a validator in `spec` to deny such configs.

AstroProfundis:
/hold

The default behavior of deploying monitoring agents in `tiup-dm` need more tests

lucklove (approved):
LGTM

PR #1495, opened 2021-07-21
playground: ignore version check error in some cases

### LINKED ISSUE #1481: It would be better bypass platform check when user provides binary
```
╭─dongxu at dongxu’s MacBook Air (2) in ~/pd on master✔ 21-07-09 - 15:56:22
╰─⠠⠵ tiup playground --db.binpath ~/tidb/bin/tidb-server --kv.binpath ~/tikv/target/debug/tikv-server --pd.binpath ~/pd/bin/pd-server --tiflash 0
Starting component `playground`: /Users/dongxu/.tiup/components/playground/v1.5.2/tiup-playground --db.binpath /Users/dongxu/tidb/bin/tidb-server --kv.binpath /Users/dongxu/tikv/target/debug/tikv-server --pd.binpath /Users/dongxu/pd/bin/pd-server --tiflash 0
Error: Playground bootstrapping failed: component tidb doesn't support platform darwin/arm64
Error: run `/Users/dongxu/.tiup/components/playground/v1.5.2/tiup-playground` (wd:/Users/dongxu/.tiup/data/SceS3nO) failed: exit status 1
```

Bypass platform check when user provides binary, I think it's more reasonable.

AstroProfundis:
Introduced in #1325, hopefully fix #1481

If the user has any of the `binpath` arguments set, we assume the user is a developer testing their latest code, so use the latest possible version for other components if needed. The version is resolved with `linux-amd64` as platform as it's the platform with every released versions available, and ignore the real platform the user is using. This, however, may cause issue if the user didn't specify `binpath` for **all components** used, and the latest version is not available for those components in the user's platform. In such cases, the user should either specify `binpath` for all the remaining components as well, or set a version that available for those components.

### Check List 

Tests 

I have **not** tested this PR as I don't have access to any Apple M1 device, on which the issue is reproducible.

Code changes

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/components/playground/main.go
+++ b/components/playground/main.go
@@ -36,6 +36,7 @@ import (
 	"github.com/pingcap/errors"
 	"github.com/pingcap/tiup/components/playground/instance"
 	"github.com/pingcap/tiup/pkg/cluster/api"
+	"github.com/pingcap/tiup/pkg/cluster/spec"
 	"github.com/pingcap/tiup/pkg/environment"
 	"github.com/pingcap/tiup/pkg/localdata"
 	"github.com/pingcap/tiup/pkg/logger/log"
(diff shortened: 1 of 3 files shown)


lucklove (approved):
LGTM

PR #1500, opened 2021-07-27
Fix telemetry timeout

AstroProfundis:
The timeout setting (hard coded 2s) was not taking affect. It was introduced in #1327 

- Disable telemetry in `local_install.sh` as offline installations does not need telemetry at all (they typically don't have access to the Internet)
- Add support of context in the `HTTPClient` in `utils` to handle timeout correctly.

### Check List 

Tests 

 - Unit test

Code changes

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/pkg/utils/http_client.go
+++ b/pkg/utils/http_client.go
@@ -28,11 +28,10 @@ import (
 // HTTPClient is a wrap of http.Client
 type HTTPClient struct {
 	client *http.Client
-	ctx    context.Context
 }
 
 // NewHTTPClient returns a new HTTP client with timeout and HTTPS support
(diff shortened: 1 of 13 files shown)


pkg/utils/http_client.go
```diff
@@ -27,16 +28,17 @@ import (
 // HTTPClient is a wrap of http.Client
 type HTTPClient struct {
 	client *http.Client
+	ctx    context.Context
```
lucklove, line 31: It's not a good idea to save context

lucklove (approved):
LGTM

Issue #1501, opened 2021-07-27
Server resource monitoring for dm components


glkappe:
We need to add server monitoring to dm-worker and dm master to check whether there are server resource bottlenecks in each component of dm in high concurrency scenarios

AstroProfundis:
It's possible to deploy monitoring agents with `tiup-dm`, they are disabled and will not be deployed by default. However, setting an `monitored` section in the topology on deploy time with ports of both `node_exporter` and `blackbox_exporter` will enable the deployment of them.

PR #1502, opened 2021-07-28
Update embed examples

AstroProfundis:
Deploying monitor agents in `tiup-dm` is supported in a slightly different way as in `tiup-cluster`, but it was not documented anywhere. This causes issues like #1501 

Add example of specifying monitor agent configs in the full topology example of `tiup-dm`, which can be printed with `tiup-dm template --full`. The section is commented as the default behavior is not to deploy those agents to avoid conflict with tidb clusters deployed by `tiup-cluster`.

Also adjusted the directory layout of `embed`, move examples to a dedicated path.

### Check List 

Tests 

 - Unit test

Release notes:

```release-note
NONE
```


lucklove (approved):
LGTM

PR #1506, opened 2021-07-30
playground: Set terminal title

dveeden:

This sets the terminal title to "TiUP Playground: <name>" where name
is the last part of the datadir. With `tiup -T foobar playground` etc.
this name can be set. Also `tiup client` uses the same names.

### Check List 

Tests 

Release notes:

```release-note
TiUP Playground now sets the terminal title to the name of the playground
```

### DIFF
--- a/components/playground/main.go
+++ b/components/playground/main.go
@@ -595,6 +595,9 @@ func newEtcdClient(endpoint string) (*clientv3.Client, error) {
 }
 
 func main() {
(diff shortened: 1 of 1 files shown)


dveeden:
This needs to be tested with a few terminal emulators on macOS and Linux for compatibility issues.

AstroProfundis (approved):
image
Works for me on ArchLinux connecting to an Debian VM via SSH.

PR #1510, opened 2021-08-05
ignore exporters when check data_dir overlap

AstroProfundis:
When user scale out new instance on a host where an imported instance already exist, and the global options set monitor paths to where the imported instance is located (and the log dir overlaps with data dir), the dir overlap check fails.

Ignore log dir overlapping if one of the compared dir is data dir of a monitor instance, as node_exporter and blackbox_exporter don't need data dir at all, the field is only kept for compatibility with tidb-ansible as the field was available there.

### Check List 

Tests 

 - Unit test

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
cluster: ignore exporters when check data_dir overlap
```


lucklove (approved):
LGTM

PR #1511, opened 2021-08-08
playground: fix bug which can't specify db port

hecomlilong:
The command "tiup playground" can't specify db port even though customizing --db.config

When db port is specified, it will be passed by parameter -P in function of "NewTiDBInstance"

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

NONE

Side effects
NONE

Related changes

Release notes:

```release-note
NONE
```


components/playground/instance/tidb_config_test.go
```diff
@@ -0,0 +1,33 @@
+package instance
+
+import (
+	"bytes"
+	"os"
+	"testing"
+
+	"github.com/BurntSushi/toml"
+	"github.com/pingcap/tidb/config"
```
AstroProfundis, line 9: We are avoiding dependency upon `tidb` and other components by intend, please use general interface objects (`map[string]interface{}` with type assertion) to decode the config.

Or simply provide a command line argument in `playground` to specify TiDB port and overwrite the default.

AstroProfundis (approved):

PR #1512, opened 2021-08-09
playground: deprecate bootstrap option `--monitor` and add a new option `--without-monitor` to disable monitors.

LittleFall:

What `--monitor                  Start prometheus and grafana component` means is unclear, `tiup playground nightly` and `tiup playground nightly --monitor` has no difference.

1. deprecate option `--monitor`, after this pr, if you add this, it still work but will report `Flag --monitor has been deprecated, Please use --without-monitor to control whether to disable monitor.
`
2. add option `--without-monitor`, in help messagae: `--without-monitor          Don't start prometheus and grafana component
`. by the way, help message won't show `--mointor` anymore.
3. when the two option occurs at the same time, the last occured will work.

### Check List 

Tests 

 - Integration test

Related changes

Release notes:

```release-note
deprecate bootstrap option --monitor and add a new option --without-monitor to disable monitors.
```


components/playground/main.go
```diff
@@ -296,7 +296,7 @@ If you'd like to use a TiDB version other than %s, cancel and retry with the fol
 	defaultOptions := &BootOptions{}
 
 	rootCmd.Flags().String(mode, defaultMode, "TiUP playground mode: 'tidb', 'tikv-slim'")
-	rootCmd.Flags().Bool(withMonitor, false, "Start prometheus and grafana component")
+	rootCmd.Flags().Bool(withMonitor, true, "Whether to start prometheus and grafana component, default value is `--monitor=true`")
```
tisonkun, line 299:
```suggestion
	rootCmd.Flags().Bool(withMonitor, true, "Whether to start prometheus and grafana component. (default: true)")
```

tend to be the same style of default message generated by `flags` and it is neat.

tisonkun (commented):
Rest LGTM.

components/playground/main.go
LittleFall: maybe we need to tell users how to set it? actually, `tiup playground nightly --monitor false` is useless.

tisonkun: In my opinion it is go flags responsibility. But it is not a strong requirement from my side. May @lucklove & @lilyjazz have other suggestion.

AstroProfundis:
I think it's better figure out why https://github.com/pingcap/tiup/blob/9a1ad59928c00b88c8844ca6aa76ad139be7e6d9/components/playground/main.go#L367 is not working in this scenario and fix it.

LittleFall:
hi @tisonkun @AstroProfundis, I've investigated why `--monitor false` is useless, and the conclusion is: 

```golang
rootCmd.Flags().Bool(withMonitor, false, "Start prometheus and grafana component")
```
will create a `pflag.Flag` with `NoOptDefVal` is true. [here is the code](https://github.com/spf13/pflag/blob/v1.0.5/bool.go#L56):
```golang
flag.NoOptDefVal = "true"
```

and when `pflag` parsing out command arguments, it won't accept pattern `--flag arg` when `flag` has NoOptDefVal. [here is the code](https://github.com/spf13/pflag/blob/v1.0.5/flag.go#L984-L998):
```golang
	if len(split) == 2 {
		// '--flag=arg'
		value = split[1]
	} else if flag.NoOptDefVal != "" {
		// '--flag' (arg was optional)
		value = flag.NoOptDefVal
	} else if len(a) > 0 {
		// '--flag arg'
		value = a[0]
		a = a[1:]
	} else {
		// '--flag' (arg was required)
		err = f.failf("flag needs an argument: %s", s)
		return
	}
```

LittleFall:
A simple fix is set `NoOptDefVal` of flag Monitor to empty like this:
```
rootCmd.Flags().Bool(withMonitor, false, "Start prometheus and grafana component")
rootCmd.Flags().Lookup(withMonitor).NoOptDefVal = ""
```
but this will introduce a compatible issue:
|                 | Before patch (NoOptDefVal="true") | After patch (NoOptDefVal="")                 |
| --------------- | --------------------------------- | -------------------------------------------- |
| (Default)       | enable monitor                    | enable monitor                               |
| --monitor=true  | enable monitor                    | enable monitor                               |
| --monitor true  | enable monitor                    | enable monitor                               |
| --monitor=false | Disable monitor                   | Disable monitor                              |
| --monitor false | **enable monitor**                | **Disable monitor**                          |
| --monitor       | **enable monitor**                | **Error: flag needs an argument: --monitor** |

what do you think about this?

LittleFall:
another thing is that after this patch, the help message will be like 
`--monitor                  Whether to start prometheus and grafana component (default true)`
we can't make it to be 
`--monitor bool                 Whether to start prometheus and grafana component (default true)`

because it's written in `pflag` code [here](https://github.com/spf13/pflag/blob/v1.0.5/flag.go#L582-L600)
```golang
	name = flag.Value.Type()
	switch name {
	case "bool":
		name = ""
	case "float64":
		name = "float"
	case "int64":
		name = "int"
	case "uint64":
		name = "uint"
	case "stringSlice":
		name = "strings"
	case "intSlice":
		name = "ints"
	case "uintSlice":
		name = "uints"
	case "boolSlice":
		name = "bools"
	}
```

tisonkun (commented):
I think it is as complex as before...

AstroProfundis:
Ummmmmm, I'd agree that change this argument to `--disable-monitor` and set it default to true might be easier.

tisonkun (commented):
I'd prefer `disable-` or `without-`.

BTW, what about user facing interface changes? Previous settings based on --monitor will break.

AstroProfundis (approved):

PR #1517, opened 2021-08-12
*: use the rand source initialized by ourself

AstroProfundis:
As we have a global random source initialized in our own code, there is no need to use the default source from `crypto/rand`, so:
- Move the package `pkg/utils/rand` to `pkg/crypto/rand`.
- Use the random source initialized by ourselves in `pkg/crypto/rand` for all usages

Release notes:

```release-note
NONE
```


lucklove (approved):
LGTM

PR #1522, opened 2021-08-18
Adjust --comments position

glkappe:
Adjust --comments position，simple delete `（no password）`
The reason is that more commands can be copied and pasted

### Check List 

Tests 

 - No code

Code changes

 - Has persistent data change

Side effects

Related changes

Release notes:

```release-note
NONE
```


AstroProfundis:
There is also one at L887

glkappe:
> There is also one at L887

done

AstroProfundis (approved):

PR #1545, opened 2021-09-01
fix binpath not work

### LINKED ISSUE #1534: TiUP shouldn't download component binaries when binpath is provided
## Bug Report

1. What did you do?
run `tiup playground --mode=tikv-slim --kv.binpath=/xxx/tikv-server` without an installed tikv component.

Since I'm using a local binary built from source, TiUP does not need to download the component and would launch with the local binary.

3. What did you see instead?
Instead of launching with the local binary, TiUP starts downloading the tikv component which is not needed
image (2)

```
1.5.5 tiup
Go Version: go1.16.7
Git Ref: v1.5.5
GitHash: 01df6de49dbf23af1a3e9d6f71d383c407cdb81
```

nexustar:

when binpath is provided, tiup still download unesserary component, fix #1534 

don't download when binpath is provided,and fix a parameter

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```


pkg/exec/run.go
```diff
@@ -176,36 +176,32 @@ type PrepareCommandParams struct {
...
+		if (p.Component == "playground" || p.Component == "cluster") && semver.Compare(selectVer.String(), "v1.0.0") < 0 {
+			return nil, errors.Errorf("incompatible component version, please use `tiup update %s` to upgrade to the latest version", p.Component)
+		}
+
 		binPath, err = env.BinaryPath(p.Component, selectVer)
 		if err != nil {
 			return nil, err
 		}
 	}
+	installPath, _ := filepath.Split(binPath)
```
AstroProfundis, line 204: Better use `filepath.Dir()` for this usage.

AstroProfundis (approved):

PR #1554, opened 2021-09-09
update root.json only when necessary

nexustar:

No need to update root.json every time update something

remote update of root.json before get timestamp.json, add it after get snapshot.json

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/pkg/repository/v1_repository.go
+++ b/pkg/repository/v1_repository.go
@@ -219,7 +219,7 @@ func (r *V1Repository) updateLocalSnapshot() (*v1manifest.Snapshot, error) {
 	}(time.Now())
 
 	timestampChanged, tsManifest, err := r.fetchTimestamp()
-	if v1manifest.IsSignatureError(errors.Cause(err)) {
+	if v1manifest.IsSignatureError(errors.Cause(err)) || v1manifest.IsExpirationError(errors.Cause(err)) {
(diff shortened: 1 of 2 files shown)


nexustar:
I also removed root version match check because it is not mentioned on TUF and the signature was already checked

pkg/repository/v1_repository.go
```diff
@@ -281,6 +266,22 @@ func (r *V1Repository) updateLocalSnapshot() (*v1manifest.Snapshot, error) {
...
 			return nil, err
 		}
+
+		// Update root if needed and restart the update process
+		var oldRoot v1manifest.Root
+		_, _, err = r.local.LoadManifest(&oldRoot)
+		if err != nil {
+			return nil, err
+		}
+		newRootVersion := snapshot.Meta[v1manifest.ManifestURLRoot].Version
```
AstroProfundis, line 276: What if the latest `snapshot.json` is signed with a new key only available in the (remote) new `root.json`?

nexustar: fixed as new commit, it will update root.json when SignatureError or ExpirationError

AstroProfundis (approved):

PR #1556, opened 2021-09-15
remove TIUP_WORK_DIR

nexustar:

it is unnecessary to set TIUP_WORK_DIR

remove it

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```


AstroProfundis (approved):
This would be a breaking change that require upgrade of at least `playground` alone with `tiup` itself together.

PR #1593, opened 2021-10-21
playground: ensure unique sockets

dveeden:

Setting `socket = "/tmp/tidb.sock"` in `tidb.toml` and then running `tiup playground --db 2 --db.config tidb.toml` causes issues as the socket path needs to be unique per instance. With https://github.com/pingcap/tidb/pull/28486 this might also cause issues with the default config.

### Check List 

Tests 

Release notes:

```release-note
A unique socket file is set for each TiDB instance
```

### DIFF
--- a/components/playground/instance/instance.go
+++ b/components/playground/instance/instance.go
@@ -39,6 +39,7 @@ type instance struct {
 	StatusPort int // client port for PD
 	ConfigPath string
 	BinPath    string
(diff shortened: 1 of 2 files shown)


dveeden:
@morgo

dveeden:
As https://github.com/pingcap/tidb/pull/29332 is merged this is not needed anymore.

PR #1596, opened 2021-10-25
support rendering template when set "--local" option

makocchi-git:
`tiup cluster template` prints example yaml file.
I think that it is more useful if users can specify global values or servers in topology.yaml.
This PR enables generating sample topology.yaml with specify variables.

When set `--local`, it renders a template with some overridden variables such as `global.user`, `global.group`, etc by setting `--user`, `--group` and so on together.

This is sample output:
```bash
$ tiup cluster template --local --user ubuntu
...
global:
  # # The OS user who runs the tidb cluster.
  user: "ubuntu"
...

=======

$ tiup cluster template --local --user ubuntu --pd-servers 127.0.0.1,127.0.0.2
...
global:
  # # The OS user who runs the tidb cluster.
  user: "ubuntu"
...
pd_servers:
  - host: 127.0.0.1
  - host: 127.0.0.2
...
```

And also support `tiup dm template --local`.

```bash
$ tiup dm template --local --user ubuntu
...
global:
  # # The OS user who runs the tidb cluster.
  user: "ubuntu"
...

=======

$ tiup dm template --local --user ubuntu --master-servers m1,m2,m3
...
global:
  # # The OS user who runs the tidb cluster.
  user: "ubuntu"
...
master_servers:
  - host: m1
  - host: m2
  - host: m3
...
```

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
Support rendering example topology.yaml with specifying variables.
```


makocchi-git:
Sorry. I fixed ioutil package.

Issue #1537, opened 2021-08-28
Server not started due to alertmanager failure


xpepermint:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?

I followed the guide at https://docs.pingcap.com/tidb/stable/production-deployment-using-tiup. I used the simple topology (single server) file and set the public server address for the `host:`. All components started successfully, but the `alertmanager` returned an error.

```yaml
global:
  user: "tidb"
  ssh_port: 22
  deploy_dir: "/tidb-deploy"
  data_dir: "/tidb-data"
server_configs: {}
pd_servers:
- host: PUBLIC_IP
tidb_servers:
- host: PUBLIC_IP
tikv_servers:
- host: PUBLIC_IP
monitoring_servers:
- host: PUBLIC_IP
grafana_servers:
- host: PUBLIC_IP
alertmanager_servers: # fails
- host: PUBLIC_IP
```

2. What did you expect to see?

All cluster components to start successfully.

3. What did you see instead?

The error of the `alertmanager` at /tidb-deploy/alertmanager-9093/log/alertmanager.log:

```
level=info ts=2021-08-28T12:44:04.536261203Z caller=main.go:140 msg="Starting Alertmanager" version="(version=0.17.0, branch=HEAD, revision=c7551cd75c414dc81df027f691e2eb21d4fd85b2)"
level=info ts=2021-08-28T12:44:04.536328631Z caller=main.go:141 build_context="(go=go1.12.4, user=root@932a86a52b76, date=20190503-09:10:07)"
level=error ts=2021-08-28T12:44:04.539124506Z caller=main.go:165 msg="unable to initialize gossip mesh" err="create memberlist: Could not set up network transport: failed to obtain an address: Failed to start TCP listener on \"3.68.95.140\" port 9094: listen tcp 3.68.95.140:9094: bind: cannot assign requested address"
```

4. What version of TiUP are you using (`tiup --version`)?

1.5.5

AstroProfundis:
When your server is behind a NAT gateway (e.g., most cloud servers inside a VPC), the public IP you could use to visit it is not assigned to the server, instead it's on the gateway. So it's not possible to bind an address not available on the system.

We do not have full support of NAT/VPC based servers, it's recommended to setup the TiUP Cluster console machine (the controller server) within the same VPC as the deployment servers, and use the private IP address to control them.

qqqdan:
Close as suggestion provided and no feedback.

PR #1608, opened 2021-11-02
add --external-url for alertmanager

reAsOn2010:
tiup deployed alertmanager creating alerts with unspecified external-url

the template generating alertmanager config

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```


AstroProfundis (approved):
Thank you for the contribute!

It would be better to add a ListenAddress for it, so that AlertManager could support VPC environment that listen IP and public access IP are different. (#1537)

But that's not necessarily a demand for this PR, I'd merge this, but it's also welcomed if you'd like to improve more on this.

PR #1639, opened 2021-11-24
Adjust for deprecation of `--monitor`

dveeden:

Replace `--monitor` in various places

### Check List 

Tests 

 - No code

Release notes:
```release-note
NONE
```

### DIFF
--- a/doc/user/playground.md
+++ b/doc/user/playground.md
@@ -22,23 +22,48 @@ The command line arguments for playground state:
 ```bash
 Usage:
   tiup playground [version] [flags]
+  tiup [command]
+
+Available Commands:
+  completion  generate the autocompletion script for the specified shell
+  display
+  help        Help about any command
+  scale-in
+  scale-out
 
 Flags:
-      --db int                   TiDB instance number (default 1)
+      --db int                   TiDB instance number
+      --db.Host host             Playground TiDB host. If not provided, TiDB will still use host flag as its host
+      --db.Port int              Playground TiDB port. If not provided, TiDB will use 4000 as its port
       --db.binpath string        TiDB instance binary path
       --db.config string         TiDB instance configuration file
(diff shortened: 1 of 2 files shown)


doc/user/playground.md
```diff
@@ -22,23 +22,48 @@ The command line arguments for playground state:
...
+  completion  generate the autocompletion script for the specified shell
+  display
+  help        Help about any command
+  scale-in
+  scale-out
 
 Flags:
-      --db int                   TiDB instance number (default 1)
+      --db int                   TiDB instance number
+      --db.Host host             Playground TiDB host. If not provided, TiDB will still use host flag as its host
```
AstroProfundis, line 36: This should be in lower case, it was a bug fixed in #1635

dveeden:
```suggestion
      --db.host host             Playground TiDB host. If not provided, TiDB will still use host flag as its host
```

doc/user/playground.md
```diff
@@ -22,23 +22,48 @@ The command line arguments for playground state:
...
+  display
+  help        Help about any command
+  scale-in
+  scale-out
 
 Flags:
-      --db int                   TiDB instance number (default 1)
+      --db int                   TiDB instance number
+      --db.host host             Playground TiDB host. If not provided, TiDB will still use host flag as its host
+      --db.Port int              Playground TiDB port. If not provided, TiDB will use 4000 as its port
```
dveeden, line 37:
```suggestion
      --db.port int              Playground TiDB port. If not provided, TiDB will use 4000 as its port
```

doc/user/playground.md
dveeden: This isn't yet fixed in v1.7.0 / ce8eb0a645cc3ead96a44d67b1ecd5034d112cf0

It is fixed in the master branch ( 9edf64ea0c1917a8aab2c355448471aa5e1419ee )

AstroProfundis (approved):

PR #1641, opened 2021-11-26
cluster: add additional prom scrape_configs

### LINKED ISSUE #1615: TiUP support user-defined prome configuration during reload
## Feature Request

Customer needs to change some metrics definition , such as remove some metrics collection.
After the customer update their prome configuration, the config will be reset when the cluster reload. 

TiUP  needs to provide a way to keep the user-defined configuration when reloading.

**Why the featue is needed:**

It is a common customer scenario. 
customer case (internal access): https://pingcap.feishu.cn/docs/doccnQEaIFtQklwMk4M2uUNPLUh

nexustar:

- close #1615 

user may use like this:

Screenshot from 2021-11-26 15-45-27

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/pkg/cluster/spec/monitoring.go
+++ b/pkg/cluster/spec/monitoring.go
@@ -17,6 +17,7 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
+	"os"
 	"path"
 	"path/filepath"
 	"reflect"
@@ -29,6 +30,7 @@ import (
 	"github.com/pingcap/tiup/pkg/cluster/template/scripts"
 	"github.com/pingcap/tiup/pkg/meta"
 	"github.com/pingcap/tiup/pkg/set"
+	"gopkg.in/yaml.v3"
 )
 
 // PrometheusSpec represents the Prometheus Server topology specification in topology.yaml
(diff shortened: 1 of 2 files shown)


AstroProfundis:
I don't think the nested multi-line config is a good idea.

We had been using nested multi-line format for the `server_config` and `xxx.config` fields in early versions of `tiup-cluster`, the result was getting issues and bug reports about "the config is incorrect" or "the config can not be applied" every week due to wrong indents user input.

There are many editors trying to automatically prepending indents when you edit the file, that's sweet in most cases, but not working well when there is anything nested and will be used to another file with different structure.

pkg/cluster/spec/monitoring.go
```diff
@@ -428,3 +437,26 @@ func (i *MonitorInstance) ScaleConfig(
...
+		for k, v := range addition {
+			job.(map[string]interface{})[k] = v
+		}
+	}
+	bytes, err = yaml.Marshal(result)
+	if err != nil {
+		return err
+	}
+	return os.WriteFile(source, bytes, 0644)
+}
```
breeswish, line 462: I think you need some unit tests for it.

AstroProfundis (approved):

breeswish (approved):

PR #1651, opened 2021-12-03
*: refactor logging to support custom settings for different session

AstroProfundis:
Using of package wide logger is not friendly to other packages to import, if any of the tiup code run inside goroutines they all have to use the same logging settings.

Set a logger for each task and manager, so that multiple instances can use different loggers.

### Check List 

Tests 

 - Integration test

Code changes

Side effects

 - Increased code complexity

Release notes:

```release-note
NONE
```


pkg/tui/progress/display_props.go
```diff
@@ -78,7 +78,7 @@ func (m Mode) String() string {
 	var s string
 	switch m {
 	case ModeSpinner:
-		s = "spinner"
+		s = "running"
```
nexustar, line 81: why, and why not edit other "spinner"

AstroProfundis: because they are really spinning...

AstroProfundis: ok i think the json marshal could also be adjusted

nexustar (approved):

PR #1659, opened 2021-12-09
cluster: check support check before scale-out the instance

### LINKED ISSUE #1621: Tiup cluster check should support check before expanding the instance
## Feature Request

Expansion of a PD instance can be successful if the PD instance already exists in another cluster and is running. As a result, the two TIDB clusters have the same PD instance.

When running tiup cluster scale-out command, you can check the ports and directories of the target machine, or add the -check option to the tiup cluster scale-out command.

**Why the featue is needed:**
O&m personnel may write wrong instance objects when performing capacity expansion operations. A check action can avoid this problem

srstack:
close #1621 

1.  check should support check with the scale-out topology file
```shell
tiup cluster check cluster-name  scale-out.yml  --cluster 
```
2.  Increase the detection of the existence of deployment directories, daily directories, data directories and systemdUnit file to prevent conflicts in tidb clusters deployed in different ways
image
3. Added prompts for scale-out operations
image

### Check List 

Tests 

Code changes

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```


AstroProfundis:
Should use `RootSSH` instead of `UserSSH` for pre-scale-out checks, as those host may have no deploy user created.

srstack:
> Should use `RootSSH` instead of `UserSSH` for pre-scale-out checks, as those host may have no deploy user created.

When checking an existing cluster, the ssh user will be forcibly set to metadata.User, which makes it impossible to log in with the root.
Now special treatment has been done for scale-out topology.

AstroProfundis (approved):

PR #1676, opened 2021-12-15
tiup/mirror: add ability to change owner of a component

AstroProfundis:
Change the owner of a component.

Use a process like rotating `root.json`:
 - On server side, the server admin run `tiup mirror transfer-owner` command to initialize a owner transfer process
 - The existence of new owner is checked, then open a temp HTTP server to serve the `component.json`
 - Client with the new owner's private key use `tiup mirror sign` to sign that temp URL
 - Server validate the signature received and update `component.json` along with `index.json`

The process is a pure administration operation, and could be used without the participate of old owner of the component, so that it also works if the old owner is absent for long time and can not be contacted.

The new owner must be granted before the transfer.

### Check List 

Tests 

Code changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/cmd/mirror.go
+++ b/cmd/mirror.go
@@ -74,6 +74,7 @@ of components or the repository itself.`,
 		newMirrorRenewCmd(),
 		newMirrorGrantCmd(),
 		newMirrorRotateCmd(),
+		newTransferOwnerCmd(),
 	)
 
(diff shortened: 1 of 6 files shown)


lucklove (approved):
LGTM

PR #1715, opened 2022-01-10
playground: fix cannot find nightly version

nexustar:
### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/pkg/repository/v1_repository.go
+++ b/pkg/repository/v1_repository.go
@@ -799,6 +799,30 @@ func (r *V1Repository) LocalComponentVersion(id, ver string, includeYanked bool)
 	return vi, nil
 }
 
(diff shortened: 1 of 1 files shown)


AstroProfundis (approved):

AstroProfundis:
We should also add a note on the "starting component" outputs to let users know which version of the component is actually running (as they may not with strictly the same version anymore)

nexustar:
> We should also add a note on the "starting component" outputs to let users know which version of the component is actually running (as they may not with strictly the same version anymore)

this fix is for v1.8.x, I will prepare another PR for that

srstack (commented):
LGTM

srstack (approved):

PR #1723, opened 2022-01-12
cluster: add significant warning destroying a cluster

### LINKED ISSUE #1722: Limit the use of tiup-cluster destroy
two question：
1. we need three confirmation，clean data 、 clean log and want continue？
image

2. @yiduoyunQ think，Do we need a switch to determine if destroy is enabled? The user's consideration is the influence of the command.Adding a layer of 'blocking' may allow customers to 'operate more accurately'

AstroProfundis:
Close #1722 

Add a significant warning when the users tries to destroy a cluster. It requires the user to input a long sentence to confirm (instead of a simple "Y/y" letter) so it's highly not possible for the user to destroy a cluster by mistake.

### Check List 

Tests 

图片

Release notes:

```release-note
NONE
```

### DIFF
--- a/pkg/cluster/manager/destroy.go
+++ b/pkg/cluster/manager/destroy.go
@@ -52,11 +52,15 @@ func (m *Manager) DestroyCluster(name string, gOpt operator.Options, destroyOpt 
 	}
 
 	if !skipConfirm {
-		if err := tui.PromptForConfirmOrAbortError(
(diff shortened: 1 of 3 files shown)


nexustar (commented):
the test script need to be modified duo the change of "y"

srstack (approved):
LGTM

nexustar (approved):

PR #1753, opened 2022-02-11
cluster:  completing the os type of the host

srstack:

Previously, the host would use the global os by default, but if it is a multi os mixed deployment, an exception will occur.

Completing the os type of the host  like full cpu architecture 

### Check List 

Tests 

Release notes:

```release-note
NONE
```


nexustar:
Will we support multi os mixed deployment?

srstack:
> Will we support multi os mixed deployment?
just  for mac.   ：(

AstroProfundis (approved):

AstroProfundis:
We don't support mixed OS in the cluster (the on-going macOS support will only be released with local deployment), but it's fine to make the code compatible with that, as long as our integrate tests on Linux pass.

PR #1785, opened 2022-03-09
add grafana embedding config

sunzhaoyang:

There is no need for repeated  development work, it can be directly embedded grafana in WebUI to view Cluster statistics.

So it's need to make grafana allow anonymous access. 

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/embed/examples/dm/minimal.yaml
+++ b/embed/examples/dm/minimal.yaml
@@ -26,6 +26,9 @@ monitoring_servers:
 
 grafana_servers:
   - host: 172.19.0.101
+  # config:       #  Enable this part if you want to use WebUI, make sure tiup dm -v newer than v1.9.0.
+    # auth.anonymous.enabled: true
+    # security.allow_embedding: true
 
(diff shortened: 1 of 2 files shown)


ljun0712:
From a security point of view, do not use anonymous login to access the monitoring data of the cluster. Monitoring data is very important, and access to it requires permission management and logging.
A more secure way can be considered: provide username/password

sunzhaoyang:
> From a security point of view, do not use anonymous login to access the monitoring data of the cluster. Monitoring data is very important, and access to it requires permission management and logging. A more secure way can be considered: provide username/password

Maybe we can tell users the risk and disable by default. Users could make the choice depend on the situation.

Security and convenience most of the time do not go hand in hand, we should let the user choose.

nexustar (approved):

PR #1819, opened 2022-03-26
cluster: display grafana URL on the top

Smityz:
I want to show grafana URL in the top, to remind users they can access Grafana directly like Dashborad
QQ20220325-210234@2x


pkg/cluster/manager/display.go
```diff
@@ -254,6 +261,23 @@ func (m *Manager) Display(name string, opt operator.Options) error {
...
+	for _, instance := range clusterInstInfos {
+		if instance.Role == "grafana" {
+			grafanaURL = append(grafanaURL, fmt.Sprintf("%s:%d", instance.Host, instance.Port))
+		}
+	}
+	if len(grafanaURL) == 0 {
+		return "", false
+	}
+	grafanaURLs := fmt.Sprintf("http://%s", grafanaURL[0])
+	for i := 1; i < len(grafanaURL); i++ {
```
AstroProfundis, line 275: This could be simplified by inserting `http://%s:%d` in L268, and use a `strings.Join()` here to concat them.

Smityz: done

PR #1853, opened 2022-04-20
fix: exporter and grafana are abnormal in prometheus

### LINKED ISSUE #1852: cluster:  enable tls, exporter and grafana are abnormal in prometheus
## Bug Report

1. What did you do?

MhLrMzQL2h

image

3. What did you see instead?

srstack:

close #1852 

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/embed/templates/config/prometheus.yml.tpl
+++ b/embed/templates/config/prometheus.yml.tpl
@@ -266,7 +266,11 @@ scrape_configs:
     scrape_interval: 30s
     metrics_path: /probe
     params:
+{{- if .TLSEnabled}}
+      module: [tls_connect]
+{{- else}}
       module: [tcp_connect]
+{{- end}}
     static_configs:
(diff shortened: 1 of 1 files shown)


nexustar (commented):
instead of use tcp on all port probe, I think we had better use another job

srstack:
> instead of use tcp on all port probe, I think we had better use another job

Using another job will involve some other configuration, which will become complicated.

srstack:
> instead of use tcp on all port probe, I think we had better use another job

ok

srstack:
> instead of use tcp on all port probe, I think we had better use another job

cc

nexustar (approved):

PR #1896, opened 2022-05-23
cluster: support RHEL8

### LINKED ISSUE #1887: Tiup support to deploy tidb 6.1 on  Red hat enterprise linux 8
## Feature Request

**Why the featue is needed:**

nexustar:

- close #1887

I do not know why,but it works.Perhaps that's been the story of SELinux

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```


nexustar:
> 



> If we suggest users to disable SELinux in `check`, I think it's okay not to workaround it.
> 
> Beside, is it possible to add new SELinux rules to allow TiDB components? Will it be too complex?

need to run `sudo chcon -t bin_t /tidb-deploy/pd-2379/scripts/run_pd.sh` for all component.Since not all systemd has SELinux,it may be too complex to add it to tiup-cluster code.If user do not accept disabling SELinux,he will need to use tiup cluster exec and 2-step scale-out.So I think we should use workaround in this PR

srstack (approved):

PR #1920, opened 2022-06-14
playground: use test-cluster as dashboard name

breezewish:

For now it is complicated to update the Dashboard JSON, for example:
https://github.com/tikv/tikv/issues/12796

One of the major changes need to take is that, playground initializes the datasource as "playground", so that the Grafana exported Dashboard JSON (containing Datasource as `${DS-PLAYGROUND}` instead of `${DS-TEST-CLUSTER}`) cannot be used by TiUP later again.

Change name to `Test-Cluster`.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/components/playground/grafana.go
+++ b/components/playground/grafana.go
@@ -145,7 +145,7 @@ func makeSureDir(fname string) error {
 	return os.MkdirAll(filepath.Dir(fname), 0755)
 }
 
-var clusterName string = "playground"
+var clusterName = "Test-Cluster"
 
 // dir should contains files untar the grafana.
 // return not error iff the Cmd is started successfully.


nexustar:
this PR may not work. In playground, we replace some vars to cluster_name https://github.com/pingcap/tiup/blob/e52c3bb95644f0081c235db60d7aed9144015824/components/playground/grafana.go#L82-L115
TEST-CLUSTER is not the only var we replace. maybe we should remove the call of this function instead

nexustar:
I guess it is ok to just use the origin datasource files. cc @AstroProfundis

breezewish:
@nexustar After replacement, the new variable name becomes DS_TEST-CLUSTER, which is identical with the original one.

AstroProfundis (approved):

PR #1928, opened 2022-06-16
README: a mis-spelling of `TiUp`

CbcWestwolf:
Just a case-sensitive fix.


AstroProfundis:
@CbcWestwolf Thank you for the fix, please also sign the CLA as guided in above comment so that we can merge this PR :)

AstroProfundis (approved):

PR #1970, opened 2022-07-01
pkg/utils: fix get free port from ipv6 host

gozssky:

`tiup playground` can not specify an ipv6 host.

Use the correct function to join host port.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
Fix playground cannot specify ipv6 host
```

### DIFF
--- a/pkg/utils/freeport.go
+++ b/pkg/utils/freeport.go
@@ -16,6 +16,7 @@ package utils
 import (
 	"fmt"
 	"net"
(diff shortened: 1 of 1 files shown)


nexustar:
Does TiDB support ipv6 now？

PR #1972, opened 2022-07-01
cdc rolling upgrade / scale-in ultilize two-phase-scheduling

### LINKED ISSUE #1971: TiUP support TiCDC rolling upgrade
## Feature Request

At the moment, when upgrade the TiCDC, it use the most simple way by replace the old instance by a new one.

After TiCDC will enable `SchedulerV3` in the near future, which support `2 phase scheduling` table schedule strategy, this can be used to help TiUP handle upgrade TiCDC in a much better way.

When the TiUP upgrade the TiCDC, TiCDC should move all tables replicating on the instance to other instance.

**Why the featue is needed:**

The strategy described above can alleviate TiCDC suffering from high replication latency, the so-called `checkpoint lag` caused by unplanned TiCDC instance shutdown.

3AceShowHand:

try close #1971

When rolling upgrade or scale in TiCDC instance, call `DrainCapture` to move all tables replicating on it to other running instance.

* introduce cdc open api to support cdc rolling upgrade strategy.
* Implement `RollingUpgradeInstance` interface for the TiCDC.
  * `upgrade` / `patch` / `scale-in` / `reload`  will use the feature.

### Check List 

Tests 

Code changes

Side effects

Related changes

Release notes:

```release-note
Support rolling upgrade TiCDC
```


3AceShowHand:
My development Golang version is `1.18.3`, I will fix the go mod file after the feature code finished.

go.mod
```diff
@@ -24,24 +24,21 @@ require (
...
 	github.com/otiai10/copy v1.7.0
 	github.com/pingcap/check v0.0.0-20211026125417-57bd13f7b5f0
-	github.com/pingcap/errors v0.11.5-0.20201126102027-b0a155152ca3
+	github.com/pingcap/errors v0.11.5-0.20211224045212-9687c2b0f87c
 	github.com/pingcap/failpoint v0.0.0-20220423142525-ae43b7f4e5c3
 	github.com/pingcap/fn v1.0.0
 	github.com/pingcap/kvproto v0.0.0-20220525022339-6aaebf466305
-	github.com/pingcap/log v1.1.0 // indirect
 	github.com/pingcap/tidb-insight/collector v0.0.0-20220111101533-227008e9835b
+	github.com/pingcap/tiflow v0.0.0-20220701111639-6d1341c93c68
```
overvenus, line 41: Please do not import tiflow, it may cause tiup binary bloat.

pkg/utils/http_client.go
```diff
@@ -155,6 +155,32 @@ func (c *HTTPClient) Post(ctx context.Context, url string, body io.Reader) ([]by
 	return checkHTTPResponse(res)
 }
 
+// PUT send a PUT request to the url and returns the response
+func (c *HTTPClient) PUT(ctx context.Context, url string, body io.Reader) ([]byte, int, error) {
```
overvenus, line 159:
```suggestion
func (c *HTTPClient) Put(ctx context.Context, url string, body io.Reader) ([]byte, int, error) {
```

pkg/cluster/api/cdcapi.go
```diff
@@ -0,0 +1,225 @@
...
+	var data []byte
+	_, err = tryURLs(endpoints, func(endpoint string) ([]byte, error) {
+		body, statusCode, err := client.client.PUT(client.ctx, endpoint, bytes.NewReader(body))
+		if err != nil {
+			if statusCode == http.StatusNotFound {
+				// old version cdc does not support `DrainCapture`, return nil to trigger hard restart.
+				return data, nil
+			}
+
+			if bytes.Contains(body, []byte("scheduler request failed")) {
```
overvenus, line 74: If tiup relies the magic string, we need to add a unit test to ticdc to make sure the string never change after since.

pkg/cluster/api/cdcapi.go
```diff
@@ -0,0 +1,225 @@
...
+	if err != nil {
+		return 0, err
+	}
+
+	var data []byte
+	_, err = tryURLs(endpoints, func(endpoint string) ([]byte, error) {
+		body, statusCode, err := client.client.PUT(client.ctx, endpoint, bytes.NewReader(body))
+		if err != nil {
+			if statusCode == http.StatusNotFound {
+				// old version cdc does not support `DrainCapture`, return nil to trigger hard restart.
```
overvenus, line 70: Could you add a debug log?

pkg/cluster/api/cdcapi.go
```diff
@@ -0,0 +1,225 @@
...
+func getAllCaptures(client *CDCOpenAPIClient) ([]*model.Capture, error) {
+	api := "/api/v1/captures"
+	endpoints := client.getEndpoints(api)
+
+	var response []*model.Capture
+
+	_, err := tryURLs(endpoints, func(endpoint string) ([]byte, error) {
+		data, err := client.client.Get(client.ctx, endpoint)
+		if err != nil {
+			return data, err
```
overvenus, line 211: How about old ticdc that does not support the API?

pkg/cluster/api/cdcapi.go
```diff
@@ -0,0 +1,225 @@
...
+	api := "/api/v1/status"
+	endpoints := c.getEndpoints(api)
+
+	// todo: only send request to the target capture.
+
+	var response model.ServerStatus
+	_, err := tryURLs(endpoints, func(endpoint string) ([]byte, error) {
+		data, err := c.client.Get(c.ctx, endpoint)
+		if err != nil {
+			return data, err
```
overvenus, line 182: How about old ticdc that does not support the API?

pkg/cluster/operation/action.go
```diff
@@ -591,6 +593,24 @@ func StopComponent(ctx context.Context,
...
 			}
+		case spec.ComponentCDC:
+			nctx := checkpoint.NewContext(ctx)
+			if !forceStop {
+				// when scale-in cdc node, each node should be stopped one by one.
+				cdc, ok := ins.(spec.RollingUpdateInstance)
+				if !ok {
+					panic("cdc should support rolling upgrade, but not")
+				}
+				err := cdc.PreRestart(nctx, topo, int(timeout), tlsCfg)
```
overvenus, line 604: What's the timeout? can we configure the value?

pkg/cluster/api/cdcapi.go
3AceShowHand: the magic string is a part of CDC RFC ERROR, https://github.com/pingcap/tiflow/blob/master/pkg/errors/cdc_errors.go#L51-L58

I think it's covered in https://github.com/pingcap/tiflow/blob/master/cdc/api/v1/api_test.go#L473, we can enhance this later.

pkg/cluster/operation/action.go
3AceShowHand: the timeout if `transfer-timeout`, default to `300s` in scale-in scenario.

overvenus: What about restart and stop scenario?

pkg/cluster/api/cdcapi.go
```diff
@@ -0,0 +1,281 @@
...
+	httpPrefix := "http"
+	if c.tlsEnabled {
+		httpPrefix = "https"
+	}
+	return fmt.Sprintf("%s://%s", httpPrefix, addr)
+}
+
+func (c *CDCOpenAPIClient) getEndpoints(cmd string) (endpoints []string) {
+	for _, addr := range c.addrs {
+		endpoint := fmt.Sprintf("%s/%s", c.getURL(addr), cmd)
```
overvenus, line 149: Please use `URL.Path = path.Join(URL.Path, cmd)`.

pkg/cluster/api/cdcapi.go
```diff
@@ -0,0 +1,281 @@
...
+func (c *CDCOpenAPIClient) DrainCapture(target string) (result int, err error) {
+	err = utils.Retry(func() error {
+		result, err = drainCapture(c, target)
+		if err != nil {
+			return err
+		}
+		return nil
+	}, utils.RetryOption{
+		Delay:   100 * time.Millisecond,
+		Timeout: 20 * time.Second,
```
overvenus, line 109: It should be configurable, otherwise TiCDC always get killed after 20s.

pkg/cluster/api/cdcapi.go
```diff
@@ -0,0 +1,266 @@
...
+				c.l().Debugf("cdc drain capture failed: %s", body)
+				return data, nil
+			}
+			if bytes.Contains(body, []byte("capture not exists")) {
+				c.l().Debugf("cdc drain capture failed: %s", body)
+				return data, nil
+			}
+			return data, err
+		}
+		return data, nil
```
overvenus, line 85: Data is always empty, and it cause `json.Unmarshal` fails.

pkg/cluster/api/cdcapi.go
```diff
@@ -0,0 +1,266 @@
...
+		return 0, err
+	}
+
+	var resp DrainCaptureResp
+	err = json.Unmarshal(data, &resp)
+	if err != nil {
+		return resp.CurrentTableCount, err
+	}
+
+	c.l().Infof("cdc drain capture finished, target=%+v, current_table_count=%+v, err=%+v", target, result, err)
```
overvenus, line 97: `result` is always **0**.

overvenus, line 97: `result` is always **0**.

pkg/cluster/operation/action.go
3AceShowHand: For the `Restart` and the`Stop`, the timeout is not set. all instances stopped by force, so it's useless.

nexustar (approved):
not familiar with TiCDC openapi, rest LGTM

PR #1987, opened 2022-07-14
cluster: support IngoreInitConfigComps

Smityz:
If we have a large number of tikv-server, it will cost a lot of time(several hours) to generate config

`./tiup-cluster scale-in Kvstore_UAT_0 --node <IP:port> -y --ignore-components tikv`

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```


Smityz:
```
  - Download blackbox_exporter: (linux/amd64) ... Done
  - Download node_exporter: (linux/amd64) ... Done
  - Download alertmanager: (linux/amd64) ... Error
Error: read manifest from mirror(https://tiup-mirrors.pingcap.com/) failed: invalid signature for file root.json: not enough signatures (2) for threshold 3 in root.json
```

I can't fix this unit test, can someone help me?

nexustar:
I think we should resolve that it takes a lot of time(several hours) to generate config, rather than provide a flag to ignore it

Smityz:
> I think we should resolve that it takes a lot of time(several hours) to generate config, rather than provide a flag to ignore it

I don't know why we should generate configs in all nodes when scale in. The config seems unchanged. Could you please explain it to me?

Smityz:
PTAL

AstroProfundis:
If any PD node is scaled in, we should re-generate configs for all TiKV and TiDB nodes as they are in the startup scripts. And the Prometheus config is always updated if any node is added or removed from the cluster.

I agree that we don't have to regenerate configs for all nodes in some cases, but that could be quite complex to implement, the current approach is a reasonable workaround.

Could you rename the `--ignore-components` argument to something like `--ignore-config-roles` to show that it is for configs? And I think it could be better to mark it as hidden as well.

Smityz:
Thanks for your explanation @AstroProfundis 
I think maybe we can disable regenerate configs when TiDB/TiKV scale in/out or prune? I think this operation is safe.

AstroProfundis:
Sorry for the delay...

> I think maybe we can disable regenerate configs when TiDB/TiKV scale in/out or prune?

I agree, and I think TiFlash is also safe to be ignored, but I'm not 100% sure about that...

Smityz:
In our production environment(only TiKV cluster), I have used this code many times and found nothing unusual, how about adding this feature as optional? Our cluster has hundreds of nodes, and init config for every node when scaling is really slow.

AstroProfundis:
I agree that adding it as an optional switch for users to decide what components should be ignored when updating configs could be reasonable.

> Could you rename the `--ignore-components` argument to something like `--ignore-config-roles` to show that it is for configs? And I think it could be better to mark it as hidden as well.

How about like this?

Smityz (commented):
Have updated name @AstroProfundis 
And I don't know what's the meaning of `mark it as hidden`

Smityz:
@AstroProfundis It's been a long time, Do you still interested in this PR?

PR #2010, opened 2022-08-07
RFC: Separate Component Version in Cluster

nexustar:
### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```


doc/rfcs/0001-separate-component-version-in-cluster.md
```diff
@@ -0,0 +1,44 @@
+# Separate Component Version in Cluster
+
+## Summary
+
+Add a version field to each component on topology file. Allow user to upgrade single component to given version.
+
+## Motivation
+
+- New component TiDB Dashboard need to deployed with TiDB but not release with TiDB. Need a way to specify Dashboard version when deployment and upgrade.
```
breezewish, line 9: I suspect whether TiDB Dashboard is really capable of a standalone version.

This means it is possible that user deploy a TiDB Dashboard with a very old TiDB, for example, TiDB 4.0 with TiDB Dashboard master. I guess this will just make TiDB Dashboard not working at all.

lonng: I think use different component for TiDB cluster should not be considered by TiUP PM. Because TiDB cannot ensure different versions can work well together. We should not do this if no one can guarantee the behaviors of a TiDB cluster containing the different versions of components.

Implementing this feature is very easy and the hard part is the different version compatibility.

AstroProfundis: There comes already a demand that TiDB-Dashboard want's to release standalone versions different with TiDB's, so we as a framework need to be prepared with such usage cases.

I do agree we should only enable this config for certain components, at this time, maybe only dashboard, in the `Specification` definitions.

doc/rfcs/0001-separate-component-version-in-cluster.md
```diff
@@ -0,0 +1,44 @@
...
+    ...
+	NodeExporterVersion     string    `yaml:"node_exporter_version,omitempty" default:"latest"`
+	BlackboxExporterVersion string    `yaml:"blackbox_exporter_version,omitempty" default:"latest"`
+    ...
+	}
+```
+
+4. if user not specify version, those version has default value latest which means use newest stable version in tiup repo.
+
+5. maybe add version field to all other component, deafult value is empty which means use global cluster version
```
AstroProfundis, line 40: As comments above, we should only add component specified version fields to those claims to work with other components on different versions.

breezewish: Maybe some version constraint is also needed. For example, TiDB Dashboard vX.X requires TiDB >= vX.X.

doc/rfcs/0001-separate-component-version-in-cluster.md
nexustar: > I think use different component for TiDB cluster should not be considered by TiUP PM. Because TiDB cannot ensure different versions can work well together. We should not do this if no one can guarantee the behaviors of a TiDB cluster containing the different versions of components.
> 
> Implementing this feature is very easy and the hard part is the different version compatibility.

tiup cluster is not only used in prod environment. I think we cloud only allow user to specify tidb-dashboard version by default. And add a flag like --allow-separate-vertion that allow user to specify version of any component to test or patch component.

doc/rfcs/0001-separate-component-version-in-cluster.md
nexustar: > Maybe some version constraint is also needed. For example, TiDB Dashboard vX.X requires TiDB >= vX.X and certain constraint break should not be allowed.

maybe we do not need it? if latest tidb-dashboard always support all exist tidb (>=6.2.0). just leave "latest" alisas as tidb-dashboard version. If user upgrade tidb cluster, tidb-dashboard upgrade to latest.If user do not upgrade tidb cluster, tidb-dashboard remains old version. We cloud just to warning user not to upgrade single component if he don't know what will happen.

nexustar: it is easy to add something like new dashboard do not support old tidb, but not easy to add information like old dashboard do not support new tidb

nexustar:
On TiDB v6.3.0, we will just make TiDB-Dashboard use the same version as TiDB. This RFC will not be implement recently.

kaaaaaaang (approved):

PR #2012, opened 2022-08-09
cdc check healthy when rolling upgrade

3AceShowHand:

* introduce cdc healthy open api, check for cdc cluster healthy before perform rolling upgrade.
* enable ticdc rolling upgrade since `v6.3.0`, which is not public yet.

### Check List 

Tests 

Code changes

Side effects

Related changes

Release notes:

```release-note
TiUP support TiCDC graceful rolling upgrade since 6.3.0
```

### DIFF
--- a/pkg/cluster/operation/operation.go
+++ b/pkg/cluster/operation/operation.go
@@ -30,6 +30,7 @@ const (
 
 type contextKey string
 
+// ContextKeyCurrentVersion is the key of current cluster version
 const ContextKeyCurrentVersion contextKey = "currentVersion"
 
 func getCurrentVersionFromContext(ctx context.Context) string {
--- a/pkg/cluster/operation/operation.go
+++ b/pkg/cluster/operation/operation.go
@@ -14,7 +14,6 @@
 package operator
 
 import (
-	"context"
 	"fmt"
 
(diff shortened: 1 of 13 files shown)


pkg/cluster/operation/operation.go
```diff
@@ -27,6 +28,24 @@ const (
 	EnvNamePDEndpointOverwrite  = "FORCE_PD_ENDPOINTS"
 )
 
+type contextKey string
+
+// ContextKeyCurrentVersion is the key of current cluster version
+const ContextKeyCurrentVersion contextKey = "currentVersion"
+
+func getCurrentVersionFromContext(ctx context.Context) string {
```
nexustar, line 36: I suggest just put cluster version on args of upgrade function. There is no need to put it to context. rest LGTM

3AceShowHand: I've tried it, but it causes the signature to change, which makes a lot of other code changes.
 If it's acceptable, it's ok to me.

nexustar: I think it is acceptable, since we never promise api stable in tiup v1.x

3AceShowHand: resolved

nexustar (approved):

srstack (approved):

Issue #1999, opened 2022-07-27
cluster/playground: Support component TiKV-CDC


pingyu:
## Feature Request

**Is your feature request related to a problem? Please describe:**

**Describe the feature you'd like:**

TiUP cluster & playground supports the [TiKV-CDC](https://github.com/tikv/migration/tree/main/cdc) component.

**Why the featue is needed:**

1. To deploy TiKV-CDC component by TiUP cluster.
2. To make it easy for users to try the TiKV-CDC component by TiUP playground.

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

PR #2000, opened 2022-07-27
playground: Supports the TiKV-CDC component

pingyu:
### What problem does this PR solve? 
Close #1999: playground: Support component [TiKV-CDC](https://github.com/tikv/migration/tree/main/cdc).

### What is changed and how it works?
Add codes to deal with deploy, scale-in and scale-out of the TiKV-CDC component.

### Check List 

Tests 

 - Integration test
 - Manual test (add detailed scripts or steps below)

deploy:
```
❯ bin/tiup-playground v6.2.0 --kvcdc 3 --kvcdc.version=v1.0.0-alpha --db 1 --kv 1 --pd 1 --tiflash 0 --tag clst30 --pd.port 20000
Playground Bootstrapping...
Start pd instance:v6.2.0
Start tikv instance:v6.2.0
Start tidb instance:v6.2.0
Start tikv-cdc instance:v1.0.0-alpha
Start tikv-cdc instance:v1.0.0-alpha
Start tikv-cdc instance:v1.0.0-alpha
Waiting for tidb instances ready
127.0.0.1:38211 ... Done
CLUSTER START SUCCESSFULLY, Enjoy it ^-^
To connect TiDB: mysql --comments --host 127.0.0.1 --port 38211 -u root -p (no password)
To view the dashboard: http://127.0.0.1:20000/dashboard
PD client endpoints: [127.0.0.1:20000]
To view the Prometheus: http://127.0.0.1:34991
To view the Grafana: http://127.0.0.1:39987
❯ 
❯ ^CPlayground receive signal:  interrupt
Wait prometheus(2794768) to quit...
Wait ng-monitoring(2794769) to quit...
Wait grafana(2794849) to quit...
Wait tikv-cdc(2793600) to quit...
Grafana quit
tikv-cdc quit
Wait tikv-cdc(2793601) to quit...
tikv-cdc quit
Wait tikv-cdc(2793611) to quit...
ng-monitoring quit
tikv-cdc quit
Wait tidb(2793599) to quit...
prometheus quit
tidb quit
Wait tikv(2793546) to quit...
tikv quit
Wait pd(2793509) to quit...
pd quit
```

`display`, `scale-in` and `scale-out`:
```
❯ bin/tiup-playground display --tag clst30
Pid      Role      Uptime
---      ----      ------
2670157  pd        29.357248287s
2670173  tikv      29.175895857s
2670180  tidb      29.012693018s
2670233  tikv-cdc  28.543214701s
[...]

PR #2022, opened 2022-08-15
cluster: Support TiKV-CDC component

pingyu:
Issue Number: #1999

Signed-off-by: pingyu <yuping@pingcap.com>

#1999 

TiUP cluster supports the [TiKV-CDC](https://github.com/tikv/migration/tree/main/cdc) component.

### Check List 

Tests 

 - Unit test
 - Integration test

```sh
➜  tiup-cluster deploy up v6.2.0 up.yaml
...
...
Cluster `up` deployed successfully, you can start it with command: `tiup-cluster start up --init`
➜  tiup-cluster start up -y
Starting cluster up...
...
...
Started cluster `up` successfully
➜  tiup-cluster display up
Cluster type:       tidb
Cluster name:       up
Cluster version:    v6.2.0
Deploy user:        root
SSH type:           builtin
...
...
Total nodes: 14
➜ tikv-cdc cli --pd 10.2.x.x:2379 capture list
[
  {
    "id": "3e2ede2a-13ad-4ded-8ac6-1c5f315dfd5e",
    "is-owner": false,
    "address": "10.2.x.x:8600"
  },
  {
    "id": "7b68a84e-63d5-4fd2-89a6-90aa15a7069c",
    "is-owner": false,
    "address": "10.2.x.x:8600"
  },
  {
    "id": "a05ae66d-6dcf-4c28-b0d2-0cab3a221dfc",
    "is-owner": true,
    "address": "10.2.x.x:8600"
  }
]
➜
```

Code changes

   - spec.TiKVCDCSpec, scripts.TiKVCDCScript, api.TiKVCDCOpenAPIClient
   - spec.ServerConfigs.TiKVCDC, spec.Specification.TiKVCDCServers, instance.ComponentTiKVCDC

Side effects

 - Increased code complexity

Related changes

   - TBD

Release notes:

```release-note
TiUP cluster supports the `TiKV-CDC` component.
```

### DIFF
--- a/pkg/cluster/spec/bindversion.go
+++ b/pkg/cluster/spec/bindversion.go
@@ -28,7 +28,8 @@ func TiDBComponentVersion(comp, version string) string {
 		ComponentPushwaygate,
 		ComponentCheckCollector,
 		ComponentSpark,
-		ComponentTiSpark:
+		ComponentTiSpark,
+		ComponentTiKVCDC: // TiKV-CDC use individual version.
 		return ""
 	default:
 		return version
--- a/pkg/cluster/spec/spec.go
+++ b/pkg/cluster/spec/spec.go
@@ -112,7 +112,7 @@ type (
 		Pump           map[string]interface{} `yaml:"pump"`
 		Drainer        map[string]interface{} `yaml:"drainer"`
 		CDC            map[string]interface{} `yaml:"cdc"`
-		TiKVCDC        map[string]interface{} `yaml:"tikv-cdc"`
+		TiKVCDC        map[string]interface{} `yaml:"kvcdc"`
 		Grafana        map[string]string      `yaml:"grafana"`
 	}
 
@@ -128,7 +128,7 @@ type (
 		PumpServers      []*PumpSpec          `yaml:"pump_servers,omitempty"`
 		Drainers         []*DrainerSpec       `yaml:"drainer_servers,omitempty"`
 		CDCServers       []*CDCSpec           `yaml:"cdc_servers,omitempty"`
-		TiKVCDCServers   []*TiKVCDCSpec       `yaml:"tikv-cdc_servers,omitempty"`
+		TiKVCDCServers   []*TiKVCDCSpec       `yaml:"kvcdc_servers,omitempty"`
 		TiSparkMasters   []*TiSparkMasterSpec `yaml:"tispark_masters,omitempty"`
 		TiSparkWorkers   []*TiSparkWorkerSpec `yaml:"tispark_workers,omitempty"`
 		Monitors         []*PrometheusSpec    `yaml:"monitoring_servers"`
--- a/pkg/cluster/spec/spec.go
+++ b/pkg/cluster/spec/spec.go
@@ -112,6 +112,7 @@ type (
 		Pump           map[string]interface{} `yaml:"pump"`
 		Drainer        map[string]interface{} `yaml:"drainer"`
 		CDC            map[string]interface{} `yaml:"cdc"`
+		TiKVCDC        map[string]interface{} `yaml:"tikv-cdc"`
 		Grafana        map[string]string      `yaml:"grafana"`
 	}
 
@@ -127,6 +128,7 @@ type (
 		PumpServers      []*PumpSpec          `yaml:"pump_servers,omitempty"`
 		Drainers         []*DrainerSpec       `yaml:"drainer_servers,omitempty"`
 		CDCServers       []*CDCSpec           `yaml:"cdc_servers,omitempty"`
+		TiKVCDCServers   []*TiKVCDCSpec       `yaml:"tikv-cdc_servers,omitempty"`
 		TiSparkMasters   []*TiSparkMasterSpec `yaml:"tispark_masters,omitempty"`
 		TiSparkWorkers   []*TiSparkWorkerSpec `yaml:"tispark_workers,omitempty"`
 		Monitors         []*PrometheusSpec    `yaml:"monitoring_servers"`
@@ -485,6 +487,7 @@ func (s *Specification) Merge(that Topology) Topology {
 		PumpServers:      append(s.PumpServers, spec.PumpServers...),
 		Drainers:         append(s.Drainers, spec.Drainers...),
 		CDCServers:       append(s.CDCServers, spec.CDCServers...),
+		TiKVCDCServers:   append(s.TiKVCDCServers, spec.TiKVCDCServers...),
 		TiSparkMasters:   append(s.TiSparkMasters, spec.TiSparkMasters...),
 		TiSparkWorkers:   append(s.TiSparkWorkers, spec.TiSparkWorkers...),
 		Monitors:         append(s.Monitors, spec.Monitors...),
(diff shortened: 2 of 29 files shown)


pingyu:
@nexustar PTAL, thanks~

haojinming (approved):
LGTM~

zeminzhou:
LGTM~

zeminzhou (approved):

pkg/cluster/spec/bindversion.go
```diff
@@ -28,7 +28,8 @@ func TiDBComponentVersion(comp, version string) string {
 		ComponentPushwaygate,
 		ComponentCheckCollector,
 		ComponentSpark,
-		ComponentTiSpark:
+		ComponentTiSpark,
+		ComponentTiKVCDC: // TiKV-CDC use individual version.
```
AstroProfundis, line 32: Is there any compatibility requirement or the user can always use the latest version of TiKV-CDC regardless of the TiDB version?

pkg/cluster/spec/spec.go
```diff
@@ -112,6 +112,7 @@ type (
 		Pump           map[string]interface{} `yaml:"pump"`
 		Drainer        map[string]interface{} `yaml:"drainer"`
 		CDC            map[string]interface{} `yaml:"cdc"`
+		TiKVCDC        map[string]interface{} `yaml:"tikv-cdc"`
```
AstroProfundis, line 115: Better use `tikv_cdc` to keep the same naming format with other components.

pkg/cluster/spec/spec.go
```diff
@@ -127,6 +128,7 @@ type (
 		PumpServers      []*PumpSpec          `yaml:"pump_servers,omitempty"`
 		Drainers         []*DrainerSpec       `yaml:"drainer_servers,omitempty"`
 		CDCServers       []*CDCSpec           `yaml:"cdc_servers,omitempty"`
+		TiKVCDCServers   []*TiKVCDCSpec       `yaml:"tikv-cdc_servers,omitempty"`
```
AstroProfundis, line 131: ditto, `tikv_cdc_servers`

tests/tiup-cluster/script/cmd_subtest.sh
```diff
@@ -5,15 +5,48 @@ set -eu
...
+                ;;
+            --tls)
+                test_tls=true
+                shift
+                ;;
+            --native-ssh)
+                native_ssh=true
+                shift
+                ;;
+            --staging)
```
AstroProfundis, line 35: This staging mirror is not used anywhere in this PR

tests/tiup-cluster/script/tikv_cdc.sh
```diff
@@ -0,0 +1,191 @@
...
+
+set -eu
+
+function tikv_cdc_test() {
+	mkdir -p ~/.tiup/bin/
+
+	version="nightly"
+	topo_name="tikv_cdc"
+	test_tls=false
+	mirror="https://tiup-mirrors.pingcap.com/"
```
AstroProfundis, line 11: `tiup mirror set --reset` will do the same as `tiup mirror set https://tiup-mirrors.pingcap.com`

tests/tiup-cluster/script/tikv_cdc.sh
```diff
@@ -0,0 +1,191 @@
...
+		--tikv-cdc-patch)
+			tikv_cdc_patch="$2"
+			shift
+			shift
+			;;
+		esac
+	done
+
+	# set mirror
+	tiup mirror set $mirror
```
AstroProfundis, line 43: as the staging mirror is not used, i think it's not necessary to build `tiup` and use it to manually set the mirror.

pkg/cluster/spec/tikv_cdc.go
```diff
@@ -0,0 +1,294 @@
...
+)
+
+// TiKVCDCSpec represents the TiKVCDC topology specification in topology.yaml
+type TiKVCDCSpec struct {
+	Host            string                 `yaml:"host"`
+	SSHPort         int                    `yaml:"ssh_port,omitempty" validate:"ssh_port:editable"`
+	Imported        bool                   `yaml:"imported,omitempty"`
+	Patched         bool                   `yaml:"patched,omitempty"`
+	IgnoreExporter  bool                   `yaml:"ignore_exporter,omitempty"`
+	Port            int                    `yaml:"port" default:"8600"`
```
AstroProfundis, line 37: The port should also be added to `portTypes` list in `pkg/cluster/spec/validate.go` at around L619, other wise the port conflict check is not working for it.

pkg/cluster/spec/spec.go
pingyu: The reason why I chose `tikv-cdc` is that:
1. I refer to this [PR](https://github.com/pingcap/tiup/pull/2017) that using `tidb-dashboard` & `tidb-dashboard_servers`.
2. The binary is named as `tikv-cdc`, so using `tikv_cdc` will be error-prone, e.g. `run_tikv-cdc.sh.tpl` will has both `tikv-cdc` & `tikv_cdc`.

Or how about using `kvcdc` ? Just as what we add to playground (#2000) ?

pkg/cluster/spec/bindversion.go
pingyu: No requirement. User can always use the latest version by now, and we will maintain the compatibility.

pkg/cluster/spec/spec.go
AstroProfundis: I think #2017 should also change to use identical name formatting, I'm absolutely sure that mixing `-` and `_` in the same key name will cause chaos to the end users.

`kvcdc` would be a good choice IMO.

pingyu: Okay~ I will change to `kvcdc`.

pkg/cluster/spec/tikv_cdc.go
pingyu: I found that the `Port` is already in the list for [`portTypes`](https://github.com/pingcap/tiup/blob/3671d82636e29dd71f86ba6fb3cf6eaec141dda9/pkg/cluster/spec/validate.go#L620) where `portConflictsDetect` get port value from. So the port conflict check should work.

And I add a test case to verify port conflict check in integrated test for tikv-cdc. PTAL~

tests/tiup-cluster/script/tikv_cdc.sh
pingyu: `tiup` is also used for installing `tikv-cdc` component for test case of `tiup cluster patch`. I remove the `tiup mirror` here, but keep the `tiup` target & function.

pkg/cluster/spec/spec.go
pingyu: Done. PTAL~

pingyu:
@AstroProfundis Thanks for your review ! All comments are addressed or replied, PTAL~

nexustar (approved):

pingyu:
Added cluster version check. PTAL, thanks~

PR #2039, opened 2022-09-14
ci: add retry in integrated tests

pingyu:
Signed-off-by: pingyu <yuping@pingcap.com>

Integrated tests occasionally failed due to external reasons (such as errors of component download, see [here](https://github.com/pingcap/tiup/actions/runs/3051118129/jobs/4919667358#step:7:93)).

At the same time, failed tests can only be rerun by who has write permissions. Other contributors can just wait, and have no idea that the tests would finally be passed or not.

So I think it would be better to do some retries on failure.

Add retries in integrated tests in gh workflows.

### Check List 

Tests 

 - Integration test

Code changes

Side effects

Related changes

Release notes:

```release-note
NONE
```


AstroProfundis:
The testing script can not be simply retried as the metadata of clusters are already written to disk. I think retry job in Github Actions would be a better way.

PR #2041, opened 2022-09-15
cluster: add support for displaying the grafana URLs in JSON output.

### LINKED ISSUE #2040: Missing Grafana URL output in json format
## Bug Report

1. What did you do?

A cluster named **test** and  there is a grafana instance in the cluster.

Then run `tiup cluster display test` and `tiup cluster display test --format json`

Both can give the result of grafana URL.

3. What did you see instead?

The Grafana URL is missing in the result in json format.

v1.10.2

onlyacat:
close #2040 

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```


nexustar:
I remember only first Grafana is used cc @AstroProfundis

AstroProfundis (approved):

AstroProfundis:
> I remember only first Grafana is used cc @AstroProfundis

Yes. Only the first Grafana in topology with the literal order is used.

Other Grafana instances should also be accessible but they all use the same datasource which is the first Prometheus in topology with the literal order.

PR #2043, opened 2022-09-20
[DNM] wait for leader to transfer back during rolling restart

cosven:
### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test (add detailed scripts or steps below)
 - No code

Code changes

 - Has exported function/method change
 - Has exported variable/fields change
 - Has interface methods change
 - Has persistent data change

Side effects

 - Possible performance regression
 - Increased code complexity
 - Breaking backward compatibility

Related changes

 - Need to cherry-pick to the release branch
 - Need to update the documentation

Release notes:

```release-note
NONE
```

PR #2051, opened 2022-10-08
wait for leader to transfer back during rolling restart

cosven:
Signed-off-by: cosven <yinshaowen241@gmail.com>

Before, there is a problem:

> Assume there are 3 tikv instances (1,2,3), and the total number of leaders is 300. When tikv1 is restarted, leaders will migrate to 2 and 3, which are 150 respectively. When restarting tikv2, the 150 leaders were migrated to tikv1 and tikv 3 at the same  time. In addition, considering that the newly restarted tikv may not be able to accept the new leader immediately, in this scenario, the leader will be strongly skewed to tikv3.

[This demo](https://github.com/pingcap/tiup/pull/2043) shows this approch can improve performance stability during rolling restart.

Performance degradation
|               | patched | default |
|---------------|---------|---------|
| mid pressure  | 6%      | 17%     |
| high pressure | 8%      | 22%     |

Total rolling restart time
|               | patched  | default  |
|---------------|----------|----------|
| mid pressure  | 135s     | 105~120s |
| high pressure | 165~180s | 150~165s |

image

This commit let tiup wait for leader to transfer back during rolling restart. In this way, the pressure should be relatively balance between tikv instances during rolling restart.

### Check List 

Tests 

 - [x] Unit test: It seems there is no need.
 - [x] Integration test:  It seems there is no need.
     - cases
       1. case1: tiup cluster do not need to wait when the total region count is small (less than 60).
       2. case2: tiup cluster need to wait until leaders are transfered back when the cluster is relatively big.
     - steps:
       1. Run `tiup cluster reload xxx -R tikv -y` to do rolling restart on one tidb cluster.
       2. Check if the leader count has recovered before next tikv is restarted.

Code changes

Side effects

   The duration of rolling restart may be longer than before.

Related changes

Release notes:

```release-note
Reduce the performance degradation during rolling restart.
```


lonng (approved):

lonng:
@cosven Please replace the image in the describtion which contains Chinese.

cosven:
I'll add another rule later: when the leader count is less than 50, just ignore recovering process, because PD does not balance leader when two tikv instances's leader count diff is less than 10.

AstroProfundis (approved):

pkg/cluster/api/pdapi.go
```diff
@@ -535,6 +536,110 @@ func (pc *PDClient) EvictStoreLeader(host string, retryOpt *utils.RetryOption, c
...
+//     Since PD considers it as balance when the leader count delta is less than 10, so
+//     these two conditions should be taken into consideration
+//
+//     - When the original leader count is less than 20, there is possibility that
+//     no leader will transfer back.
+//     For example: The target store's leader count is 19. Other stores' leader count are 9.
+//     There are 20 stores in total. In this case, there may be no leader to transfer back.
+//
+//     - When the leader count is less than 57, there is possibility that only less than 2/3
+//     leaders are transfered back. `(N-10-9 >= 2/3*N) -> (N>=57)`.
```
lhy1024, line 556:
```suggestion
//     leaders are transfered back. `(N-10-9 > 2/3*N) -> (N>57)`.
```

lhy1024: maybe we need to verify it by pd-simulator

cosven: I think N can be equal to 57. When it is 57, it should has at least 38 leader after balance and 38/57==2/3. :thinking: 

I'll take a look about the verification.

cosven: I think we just need to know the **accurate** value in our mind, and we can set the value to a larger number in tiup code. The tiup code does not need to be strongly coupled with the PD code. 

Moreover, clusters which have small number of leaders is supposed to has low pressure, and this enhancement may be unnecessary for them. Clusters in production env usually has thousands of leaders. 

So, I think we can adjust the code to the following. What do you think? @lhy1024 @AstroProfundis 
```golang
if originalCount < 200 {
    return nil
}
```

---------

In the future, tidb may support big sized regions. At that time, I think PD should provider some API to help other tools (like tiup/tidb-operators) integrate easily.

lhy1024: I think 200 is better than 57.

lhy1024: I am not sure whether `leaderTolerantSizeRatio`  will be changed by dynamic region. cc @bufferflies

cosven: Can we merge this PR first? /cc @lhy1024 @AstroProfundis 

IMO, even `leaderTolerantSizeRatio` is changed, the current threshold (200) is suitable.

lhy1024 (approved):

PR #2066, opened 2022-11-02
tiflash: check instance cpu flags when deploy | scale-out | upgrade

solotzg:

close https://github.com/pingcap/tiflash/issues/6075
rollback #2054

- use shell cmd `command -v lscpu >/dev/null 2>&1 && lscpu || true` to get cpu flags
- check cpu flags for tiflash instance when deploy | scale-out | upgrade

```
+ Detect CPU Arch Name
  - Detecting node 172.16.4.42 Arch info ... Done

+ Detect CPU OS Name
  - Detecting node 172.16.4.42 OS info ... Done

+ Detect CPU CPUFlags Name
  - Detecting node 172.16.4.42 CPUFlags info ... Done

Error: Fail to check CPU flags in host `172.16.4.42`. TiFlash requires `avx2` under `x86-64` platform since `v6.3.0`.
```

### Check List 

Tests 

 - [x] Unit test

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/components/cluster/command/upgrade.go
+++ b/components/cluster/command/upgrade.go
@@ -14,13 +14,18 @@
 package command
 
 import (
+	"path/filepath"
+
+	"github.com/pingcap/tiup/pkg/cluster/manager"
 	"github.com/pingcap/tiup/pkg/utils"
 	"github.com/spf13/cobra"
 )
 
 func newUpgradeCmd() *cobra.Command {
 	offlineMode := false
-
+	opt := manager.DeployOptions{
+		IdentityFile: filepath.Join(utils.UserHome(), ".ssh", "id_rsa"),
+	}
 	cmd := &cobra.Command{
 		Use:   "upgrade <cluster-name> <version>",
 		Short: "Upgrade a specified TiDB cluster",
@@ -38,7 +43,7 @@ func newUpgradeCmd() *cobra.Command {
 			teleCommand = append(teleCommand, scrubClusterName(clusterName))
 			teleCommand = append(teleCommand, version)
 
-			return cm.Upgrade(clusterName, version, gOpt, skipConfirm, offlineMode)
+			return cm.Upgrade(clusterName, version, gOpt, skipConfirm, offlineMode, opt)
 		},
 		ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
 			switch len(args) {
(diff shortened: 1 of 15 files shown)


AstroProfundis:
AVX2 is **not available** for `qemu64`, the default CPU type of QEMU, which makes most VM hosts not possible to pass this check.
```
processor	: 0
vendor_id	: GenuineIntel
cpu family	: 15
model		: 107
model name	: QEMU Virtual CPU version 2.5+
stepping	: 1
microcode	: 0x1
cpu MHz		: 2807.998
cache size	: 16384 KB
physical id	: 0
siblings	: 16
core id		: 0
cpu cores	: 16
apicid		: 0
initial apicid	: 0
fpu		: yes
fpu_exception	: yes
cpuid level	: 13
wp		: yes
flags		: fpu de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 ht syscall nx lm constant_tsc nopl xtopology cpuid tsc_known_freq pni cx16 x2apic hypervisor lahf_lm cpuid_fault pti
bugs		: cpu_meltdown spectre_v1 spectre_v2 spec_store_bypass l1tf mds swapgs itlb_multihit mmio_unknown
bogomips	: 5615.99
clflush size	: 64
cache_alignment	: 128
address sizes	: 40 bits physical, 48 bits virtual
```
I think it's a serious compatibility issue rather than a simple check.

components/cluster/command/upgrade.go
```diff
@@ -49,6 +54,9 @@ func newUpgradeCmd() *cobra.Command {
 			}
 		},
 	}
+	cmd.Flags().StringVarP(&opt.User, "user", "u", utils.CurrentUser(), "The user name to login via SSH. The user must has root (or sudo) privilege.")
```
AstroProfundis, line 57: This does not look good to me, adding the requirement of `sudo` user in `upgrade` is:
1. unnecessary, as the deploy user should have sudo permission;
2. introduces new dependency and changes the upgrade procedure of user

pkg/cluster/spec/tiflash.go
```diff
@@ -63,6 +63,7 @@ type TiFlashSpec struct {
 	ResourceControl      meta.ResourceControl `yaml:"resource_control,omitempty" validate:"resource_control:editable"`
 	Arch                 string               `yaml:"arch,omitempty"`
 	OS                   string               `yaml:"os,omitempty"`
+	CPUFlags             string               `yaml:"cpu_flags,omitempty"`
```
AstroProfundis, line 66: I don't understand what this field is used for

solotzg:
> AVX2 is **not available** for `qemu64`, the default CPU type of QEMU, which makes most VM hosts not possible to pass this check.

According to https://www.mankier.com/7/qemu-cpu-models#Description-Preferred_CPU_models_for_Intel_x86_hosts, is it feasible to set parameters like `-cpu Broadwell`.

components/cluster/command/upgrade.go
solotzg: If opt.User is "" rather than 'root' or other user name, we will get error like ssh handshake fail from https://github.com/pingcap/tiup/blob/7b324055e1b57e46fac715d38afeeef2ee5e6aff/pkg/cluster/manager/manager.go#L240-L242.

pkg/cluster/spec/tiflash.go
solotzg: It's used in func `func checkTiFlashCPUFlags(topo spec.Topology, version string) error` https://github.com/solotzg/tiup/blob/99e0aaf4d3f2e6598cce6011909022b7e22dd024/pkg/cluster/manager/basic.go#L310

nexustar:
I think it is necessary to add these check since we already have a check subcommand.If we do not provide a compatibility mode to allow use use tiflash without avx2, I think mention it on doc is enough. Just like we do not need to add check that user should not deploy TiDB to Windows.

AstroProfundis:
> > AVX2 is **not available** for `qemu64`, the default CPU type of QEMU, which makes most VM hosts not possible to pass this check.
> 
> According to https://www.mankier.com/7/qemu-cpu-models#Description-Preferred_CPU_models_for_Intel_x86_hosts, is it feasible to set parameters like `-cpu Broadwell`.

In most cases, end user have no control of what QEMU / KVM argument is used to start their VMs.

PR #2077, opened 2022-11-15
playground: Enable AutoTLS by default

dveeden:

TiDB Playground doesn't enable TLS by default. While TLS isn't needed in most cases to make the playground secure, it can help to detect issues with client code early on that could cause problems later when deployed on TiDB Cloud Serverless (which requires TLS).

if there is no config specified, create a temporary config that enables
AutoTLS.

https://docs.pingcap.com/tidb/stable/tidb-configuration-file#auto-tls

### Check List 

Tests 

Code changes

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
TiUP Playground enables AutoTLS by default
```

### DIFF
--- a/components/playground/instance/tidb.go
+++ b/components/playground/instance/tidb.go
@@ -16,6 +16,7 @@ package instance
 import (
 	"context"
 	"fmt"
+	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


dveeden:
@nexustar 

1. Is this ok or should we try to not set the config for versions that don't support autoTLS?
2. Should we put the config in `~/tiup/data/<name>/tidb-0/tidb.toml`  or is `/tmp/tidb-playground-config<random>` ok?

dveeden:
@nexustar PTAL

kaaaaaaang (approved):
works fine with tidb v4.0.0 and v6.1.0

nexustar (approved):

kaaaaaaang (approved):

kaaaaaaang:
some dev are using playground as a daily-develop tool and using it in ci. maybe it is is suitable to eable it by default

PR #2130, opened 2023-03-02
Allow users to run custom shell commands during the cluster rolling u…

darkelf21cn:
…pgrade

During the cluster rolling upgrade, I would like to patch the server operating system as well as online and offline servers in load balancers. With this PR, users can run their custom shell command during a cluster rolling upgrade, just before and after the component is restarted on each server.

This PR adds `--pre-upgrade-command` and `--post-upgrade-command` arguments to the `tiup cluster upgrade` command. Command in text format will be executed on each server, before the `PreRestart` and `PostRestart` functions perspectively. 

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```


AstroProfundis:
This is a good idea. I think we could mark it as "experimental" at present as I believe there could be more to add in the future.

For example, this could be extended to all "restart" and "reload" operations, as the `upgrade` calls `reload` to rolling restart the cluster, it should be usable for `reload` in nature. It could also be better if the user could upload a custom executable file (script or binary) to remote to execute before service start, and maybe the ability to only execute on selected nodes.

I'm not asking you to implement them all in this PR, but marking the feature as "experimental" could make it more flexible for us to improve it in the future as we don't need to keep compatibility. Or do you want to implement them all together?

darkelf21cn:
@AstroProfundis Thanks for the comment. I completely agree with you that user custom scripts would be useful in many operations. This PR was just a POC so I didn't design the API at a higher level. Here is the improved version. I added a new field `SSHCustomScripts` to the `Option` struct. So extending the custom ssh script to other operations will be easier. Furthermore, the custom script argument can be a file path or inline command now. Binary file execution is still not supported as it requires more changes. Last but not least, I have marked all those new arguments as experimental as you suggested.

darkelf21cn:
It seems like CI is broken

> Error: read manifest from mirror(https://tiup-mirrors.pingcap.com/) failed: invalid signature for file root.json: not enough signatures (2) for threshold 3 in root.json

nexustar (approved):

kaaaaaaang (approved):
it is simple but useful :)

PR #2133, opened 2023-03-07
display disaggregated mode

hehechen:
https://github.com/pingcap/tiup/issues/2135
When the label of TiFlash contains engine-role:write, it means that This TiFlash is a tiflash_write node. So tiflash_write should be displayed in tiup cluster display. But we don't support -R `tiflash_write` or `tiflash_compute` now, so just display `tiflash (write)` or `tiflash (compute)`.

### Check List 

Tests 

 - Unit test
 - Integration test
image

 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/pkg/cluster/manager/display.go
+++ b/pkg/cluster/manager/display.go
@@ -601,6 +601,12 @@ func (m *Manager) GetClusterTopology(dopt DisplayOption, opt operator.Options) (
 
 		// check if the role is patched
 		roleName := ins.Role()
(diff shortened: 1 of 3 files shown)


AstroProfundis (commented):
We need to test if `-R/--role` argument is working with the new patch.

AstroProfundis (approved):

PR #2138, opened 2023-03-13
Remove mark_cache_size default value of tiflash.toml after v5.4.0

hongyunyan:
ref https://github.com/pingcap/tiflash/pull/6861

Remove mark_cache_size default value of tiflash.toml after v5.4.0
Because tiflash add default_mark_cache_size = 5G in code from v5.4.0, so for the version after v5.4.0, we can directly remove this item in config.

### Check List 

Tests 

 - Unit test
 - Integration test
   use tiup-playground and tiup-cluster to start cluster in nightly and v5.3.0 to check the tiflash.toml
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/pkg/cluster/spec/tiflash.go
+++ b/pkg/cluster/spec/tiflash.go
@@ -377,6 +377,7 @@ func (i *TiFlashInstance) initTiFlashConfig(ctx context.Context, clusterVersion 
 		isStorageDirsDefined  bool
 		deprecatedUsersConfig string
 		daemonConfig          string
+		markCacheSize         string
 		err                   error
 	)
 	if isStorageDirsDefined, err = checkTiFlashStorageConfigWithVersion(clusterVersion, src); err != nil {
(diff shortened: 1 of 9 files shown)


kaaaaaaang (approved):

pkg/cluster/spec/tiflash.go
```diff
@@ -438,16 +439,17 @@ func (i *TiFlashInstance) initTiFlashConfig(ctx context.Context, clusterVersion
 	if tidbver.TiFlashNotNeedSomeConfig(clusterVersion) {
 		// For 5.4.0 or later, TiFlash can ignore application.runAsDaemon setting
```
breezewish, line 440: Please update the comment as well.

breezewish (approved):

pkg/cluster/spec/tiflash.go
hongyunyan: done

PR #2161, opened 2023-04-09
cluster: Allows setting other storage config for TiFlash when `storage.main` is not defined

breezewish:

TiUP Cluster will reject the deployment for the following TiFlash config:

```yaml
config:
  storage.remote.dir: /tmp/tiflash/remote
```

Actually this is fine: TiFlash does not require `storage.main` defined in this case.

Only check `storage.main` exists when `storage.latest` or `storage.raft` is defined.

### Check List 

Tests 

 - Unit test

Release notes:

```release-note
Allows setting other storage config for TiFlash when `storage.main` is not defined
```


JaySon-Huang (approved):
LGTM

kaaaaaaang (approved):

Issue #2170, opened 2023-04-11
cluster: wrong ngm config


kaaaaaaang:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?

2. What did you expect to see?

right pd address in config file like ["127.0.0.1:2379", "127.0.0.1:2339"]

3. What did you see instead?

wrong pd address in config file like ["127.0.0.1:2379" "127.0.0.1:2339"]

4. What version of TiUP are you using (`tiup --version`)?
v1.12.0

PR #2172, opened 2023-04-12
fix ngmonitoring.toml config

Mystery-cyf:
https://github.com/pingcap/tiup/issues/2170

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/pkg/cluster/spec/monitoring.go
+++ b/pkg/cluster/spec/monitoring.go
@@ -357,18 +357,20 @@ func (i *MonitorInstance) InitConfig(
 	}
 
 	if spec.NgPort > 0 {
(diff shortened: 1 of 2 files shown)


nexustar:
1. I think it is more suitable to use strings.Join(pds, ",")
2. You also need to add "[" and "]" on embed/config/ngmonitoring.toml.tpl since PDaddrs is string now

Cloud you mind fixing these ,or just use #2173 instead

Mystery-cyf:
Your approach is more efficient

PR #2197, opened 2023-05-06
skip destroy monitor when ignore_exporter is true

Smityz:
当我们设置`ignore_exporter: true'时，tiup不应该破坏节点导出器，因为它从未被部署过。


nexustar:
I suggest to move this if to "instCount[inst.GetManageHost()]--".So it will have the same behave like deploy.(If there are two instant on one machine, only one has "nomonitor" config, tiup cluster will deyloy and destroy monitor on this machine)

Smityz:
> I suggest to move this if to "instCount[inst.GetManageHost()]--".So it will have the same behave like deploy.(If there are two instant on one machine, only one has "nomonitor" config, tiup cluster will deyloy and destroy monitor on this machine)

updated

srstack (approved):

PR #2198, opened 2023-05-08
mkdir for raftdb-path

Smityz:
```
  - host: x.x.x.x
    port: 23000
    status_port: 23001
    data_dir: "/mnt/disk/1/x/tikv/23000"
    ignore_exporter: true
    config:
      server.labels: { host: "x.x.x.x" }
      raftstore.raftdb-path: "/mnt/disk/0/x/tikv/23000"
```
If we set `raftstore.raftdb-path` to a location different from `data_dir`, we need to create a folder for `raftstore.raftdb-path` in TiUP.
If community wants to merge this PR, I will add some unit test


pkg/cluster/spec/tikv.go
```diff
@@ -461,6 +462,15 @@ func (i *TiKVInstance) PostRestart(ctx context.Context, topo Topology, tlsCfg *t
 	return nil
 }
 
+func (i *TiKVInstance) ExtraDirs() []string {
+	configMap := reflect.Indirect(reflect.ValueOf(i.InstanceSpec)).FieldByName("Config").Interface().(map[string]interface{})
```
breezewish, line 466: How about:

```go
spec := i.InstanceSpec.(*TiKVSpec)
raftDir := spec.Config["raftstore.raftdb-path"]
```

Smityz: fixed

breezewish (approved):

breezewish:
Thanks!  @lucklove PTAL~

PR #2220, opened 2023-06-16
*: remove tcp_port for tiflash >= 7.1.0

zanmato1984:
Remove tiflash tcp_port config for tiflash >= 7.1.0.

1. For playground, remove tcp_port for tiflash < 7.1.0, as there are no compatibility and instance identity issues.
2. For ansibile and cluster, not render tcp_port config in tiflash config file for tiflash >= 7.1.0.

### Check List 

Tests 
 - No code

Code changes

Side effects
 - None

Related changes
 - None

Release notes:

```release-note
NONE
```

### DIFF
--- a/components/playground/instance/tiflash_pre7.go
+++ b/components/playground/instance/tiflash_pre7.go
@@ -146,7 +146,7 @@ func (inst *TiFlashInstance) checkConfigOld(deployDir, clusterManagerPath string
 	}()
 
 	// Write default config to buffer
-	if err := writeTiFlashConfigOld(flashBuf, version, inst.Port, inst.ServicePort, inst.StatusPort,
+	if err := writeTiFlashConfigOld(flashBuf, version, inst.Port, inst.TCPPort, inst.ServicePort, inst.StatusPort,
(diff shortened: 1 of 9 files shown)


components/playground/instance/tiflash_pre7.go
```diff
@@ -146,7 +146,7 @@ func (inst *TiFlashInstance) checkConfigOld(deployDir, clusterManagerPath string
 	}()
 
 	// Write default config to buffer
-	if err := writeTiFlashConfigOld(flashBuf, version, inst.TCPPort, inst.Port, inst.ServicePort, inst.StatusPort,
+	if err := writeTiFlashConfigOld(flashBuf, version, inst.Port, inst.ServicePort, inst.StatusPort,
```
Lloyd-Pottiger, line 149: I think we can not remove tcp port before tiflash 7.0

zanmato1984: Right, it'll cause single-node multi-instance conflict on 9000 port. Thanks for reminding.

Addressed.

Lloyd-Pottiger (approved):

PR #2314, opened 2023-11-14
playground: Only add --comments when needed

dveeden:

From https://dev.mysql.com/doc/relnotes/mysql/8.1/en/news-8-1-0.html

> Comments in the mysql client are now enabled by default. To disable them, start mysql with the --skip-comments option.
> 
> Our thanks to Daniël van Eeden for the contribution. (Bug #109972, Bug #35061087, WL #15597)

This means we can shorten the commandline to connect to TiDB if the client is newer than 8.1.0.

### Check List 

Tests 

 - Unit test

Side effects

 - Increased code complexity

Release notes:

```release-note
TiUP Playground now only prints `--comments` for older versions of MySQL Client.
```

### DIFF
--- a/components/playground/playground.go
+++ b/components/playground/playground.go
@@ -1489,7 +1489,7 @@ func mysqlCommand() (cmd string) {
 	if err != nil {
 		return
 	}
-	vMaj, vMin, _, err := parseMysqlCommand(string(mysqlVerOutput))
+	vMaj, vMin, _, err := parseMysqlVersion(string(mysqlVerOutput))
 	if err == nil {
 		if vMaj == 8 && vMin >= 1 { // 8.1.0 and newer
 			return "mysql"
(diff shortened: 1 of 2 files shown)


kaaaaaaang:
maybe user run playground in an machine without mysql client？

components/playground/playground.go
```diff
@@ -1468,3 +1471,55 @@ func logIfErr(err error) {
...
+	}
+	return
+}
+
+// parseMysqlCommand parses the output from `mysql --version` that is in `versionOutput`
+// and returns the major, minor and patch version.
+//
+// New format example: `mysql  Ver 8.2.0 for Linux on x86_64 (MySQL Community Server - GPL)`
+// Old format example: `mysql  Ver 14.14 Distrib 5.7.36, for linux-glibc2.12 (x86_64) using  EditLine wrapper`
+func parseMysqlCommand(versionOutput string) (vMaj int, vMin int, vPatch int, err error) {
```
mjonss, line 1506: Maybe Version instead of Command?
```suggestion
func parseMysqlVersion(versionOutput string) (vMaj int, vMin int, vPatch int, err error) {
```

components/playground/playground.go
```diff
@@ -1468,3 +1471,55 @@ func logIfErr(err error) {
...
+	cmd = "mysql --comments"
+	mysqlVerOutput, err := exec.Command("mysql", "--version").Output()
+	if err != nil {
+		return
+	}
+	vMaj, vMin, _, err := parseMysqlCommand(string(mysqlVerOutput))
+	if err == nil {
+		if vMaj == 8 && vMin >= 1 { // 8.1.0 and newer
+			return "mysql"
+		}
```
mjonss, line 1496: Is `--comments` default on on MariaDB CLI? From reading the knowledge base it does not look like it.

Also there is no harm in always using --comments, regardless of version, just not needed for MySQL CLI ver >= 8.1 right?

I wonder if parseMysqlCommand should also return the vendor, like MySQL or MariaDB?

mjonss (approved):
LGTM, but I assume some special handling for MariaDB is also needed?

dveeden:
> LGTM, but I assume some special handling for MariaDB is also needed?

No special handling for MariaDB is needed afaik. MariaDB strips comments by default in all versions so `--comments` is still needed. (ref: https://mariadb.com/kb/en/mariadb-command-line-client/#-c-comments )

Some commandline options in MariaDB are different (e.g. `--ssl-mode`), but I don't think we use any of these right now.

components/playground/playground.go
dveeden: Addressed in 03faa76fe16c2f09c6377ede6b4635ffaf44c7c5

mjonss (approved):

srstack (approved):

breezewish (approved):

xhebox (approved):

Issue #2292, opened 2023-10-20
tiup deployment fails on rocky 9.2


jebter:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?

tiup cluster deploy tidb-750 v7.5.0 ~/topology.yaml --user root -p
tiup is checking updates for component cluster ...
Starting component `cluster`: /home/tidb/.tiup/components/cluster/v1.12.1/tiup-cluster deploy tidb-750 v7.5.0 /home/tidb/topology.yaml --user root -p
Input SSH password:

+ Detect CPU Arch Name
  - Detecting node 172.16.6.50 Arch info ... Done

+ Detect CPU OS Name
  - Detecting node 172.16.6.50 OS info ... Done
Please confirm your topology:
Cluster type:    tidb
Cluster name:    tidb-750
Cluster version: v7.5.0
Role          Host         Ports                            OS/Arch       Directories
----          ----         -----                            -------       -----------
pd            172.16.6.50  2379/2380                        linux/x86_64  /data1/tidb-deploy/pd-2379,/data1/tidb-data/pd-2379
tikv          172.16.6.50  20160/20180                      linux/x86_64  /data1/tidb-deploy/tikv-20160,/data2/tikv-20160
tidb          172.16.6.50  4000/10080                       linux/x86_64  /data1/tidb-deploy/tidb-4000
tiflash       172.16.6.50  9000/8123/3930/20170/20292/8234  linux/x86_64  /data1/tidb-deploy/tiflash-9000,/data1/tiflash-9000
prometheus    172.16.6.50  9090/12020                       linux/x86_64  /data1/tidb-deploy/prometheus-9090,/data1/tidb-data/prometheus-9090
grafana       172.16.6.50  3000                             linux/x86_64  /data1/tidb-deploy/grafana-3000
alertmanager  172.16.6.50  9093/9094                        linux/x86_64  /data1/tidb-deploy/alertmanager-9093,/data1/tidb-data/alertmanager-9093
Attention:
    1. If the topology is not what you expected, check your yaml file.
    2. Please confirm there is no port/directory conflicts in same host.
Do you want to continue? [y/N]: (default=N) y
+ Generate SSH keys ... Done
+ Download TiDB components
  - Download pd:v7.5.0 (linux/amd64) ... Done
  - Download tikv:v7.5.0 (linux/amd64) ... Done
  - Download tidb:v7.5.0 (linux/amd64) ... Done
[...]

PR #2333, opened 2023-12-18
Add check for rocky linux

dveeden:

Rocky Linux 9.1 and later is listed in the documentation as supported, but the checks don't agree with this.

Add rocky linux to the check

### Check List 

Tests 

Release notes:

```release-note
NONE
```


dveeden:
Note that there is one open issue related to Rocky Linux: https://github.com/pingcap/tiup/issues/2292
However I don't think that changes the support status.

dveeden:
Related: https://tidbcommunity.slack.com/archives/C019S7HLP61/p1702885355480039

PR #2342, opened 2023-12-27
playground: tiproxy needs to override usercfg

### LINKED ISSUE #2339: Start multiple TiProxy by TiUP playground with a config file fails
## Bug Report

1. What did you do?

```shell
touch cfg
tiup playground v7.5.0 --tiproxy 2 --tiproxy.config cfg --tiflash 0
```

The cluster starts successfully

3. What did you see instead?

Only one TiProxy starts and another reports that the port is occupied.
This is because TiUP writes the port to the cfg file and all the TiProxy share the same file.
This is different for TiDB. TiUP won't write the TiDB config file.

v1.14.0

xhebox:

Close #2339 

previously, it doesnt override user config correctly leading to port conflicts when using custom configs for multiple tiproxy instances. TiDB pass ports by cli flags so then are fine.

Also, log-file is not passed

1. use different logic than TiDB to generate configs.
2. pass log-file

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/components/playground/instance/tiproxy.go
+++ b/components/playground/instance/tiproxy.go
@@ -20,7 +20,8 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/pelletier/go-toml"
+	"github.com/BurntSushi/toml"
+	"github.com/pingcap/tiup/pkg/cluster/spec"
 	tiupexec "github.com/pingcap/tiup/pkg/exec"
 	"github.com/pingcap/tiup/pkg/utils"
 )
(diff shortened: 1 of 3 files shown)


components/playground/instance/tiproxy.go
```diff
@@ -68,34 +69,33 @@ func (c *TiProxy) Start(ctx context.Context, version utils.Version) error {
...
-	if err := os.MkdirAll(filepath.Dir(configPath), 0755); err != nil {
-		return err
-	}
-	if err := os.WriteFile(configPath, b, 0644); err != nil {
+	enc := toml.NewEncoder(cf)
+	enc.Indent = ""
+	if err := enc.Encode(spec.MergeConfig(userConfig, map[string]any{
+		"proxy.pd-addrs":            strings.Join(endpoints, ","),
+		"proxy.addr":                utils.JoinHostPort(c.Host, c.Port),
+		"proxy.require-backend-tls": false,
```
djshow832, line 95: Do we need this line anymore?

PR #2346, opened 2023-12-29
playground: fix can not set runtime config in config file & set TiFlash logger level to debug

Lloyd-Pottiger:

1. Fix  can not set runtime config in config file specific by `--tiflash.config`

now:

image

image

image

image

2. Since TiFlash has changed the default logger level to **INFO**, explicitly set the TiFlash logger level to **DEBUG** in TiUP playground to debug more easily.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```


components/playground/instance/tiflash.go
```diff
@@ -139,6 +139,7 @@ func (inst *TiFlashInstance) Start(ctx context.Context, version utils.Version) e
 		fmt.Sprintf("--path=%s", filepath.Join(inst.Dir, "data")),
 		fmt.Sprintf("--listen_host=%s", inst.Host),
 		fmt.Sprintf("--logger.log=%s", inst.LogFile()),
+		"--logger.level=debug",
```
JaySon-Huang, line 142: can we overwrite this config when running playground？

Lloyd-Pottiger: Fixed in https://github.com/pingcap/tiup/pull/2346/commits/7ccb33d9f4dcb2acf312922ab99c989bdec45d0b

components/playground/instance/tiflash.go
```diff
@@ -131,26 +132,37 @@ func (inst *TiFlashInstance) Start(ctx context.Context, version utils.Version) e
...
-		fmt.Sprintf("--flash.proxy.advertise-addr=%s", utils.JoinHostPort(AdvertiseHost(inst.Host), inst.ProxyPort)),
-		fmt.Sprintf("--flash.proxy.status-addr=%s", utils.JoinHostPort(inst.Host, inst.ProxyStatusPort)),
-		fmt.Sprintf("--flash.proxy.data-dir=%s", filepath.Join(inst.Dir, "proxy_data")),
-		fmt.Sprintf("--flash.proxy.log-file=%s", filepath.Join(inst.Dir, "tiflash_tikv.log")),
-	)
-
-	var err error
+	}
+	runtimeConfig := [][]string{
+		{"tmp_path", filepath.Join(inst.Dir, "tmp")},
```
JaySon-Huang, line 139: I think tmp_path can be removed, TiFlash's default behavior "${path}/tmp" can handle it.

components/playground/instance/tiflash.go
```diff
@@ -131,26 +132,37 @@ func (inst *TiFlashInstance) Start(ctx context.Context, version utils.Version) e
...
+		{"flash.proxy.addr", utils.JoinHostPort(inst.Host, inst.ProxyPort)},
+		{"flash.proxy.advertise-addr", utils.JoinHostPort(AdvertiseHost(inst.Host), inst.ProxyPort)},
+		{"flash.proxy.status-addr", utils.JoinHostPort(inst.Host, inst.ProxyStatusPort)},
+		{"flash.proxy.data-dir", filepath.Join(inst.Dir, "proxy_data")},
+		{"flash.proxy.log-file", filepath.Join(inst.Dir, "tiflash_tikv.log")},
+	}
+	userConfig, err := unmarshalConfig(configPath)
+	if err != nil {
+		return errors.Trace(err)
+	}
```
JaySon-Huang, line 157: we should continue the following logic when the config file in `configPath` is not exists

Lloyd-Pottiger:
```go
	configPath := filepath.Join(inst.Dir, "tiflash.toml")
	if err := prepareConfig(
		configPath,
		inst.ConfigPath,
		inst.getConfig(),
	); err != nil {
		return err
	}
```

`configPath` always exists.

components/playground/instance/tiflash.go
Lloyd-Pottiger: removed https://github.com/pingcap/tiup/pull/2346/commits/8ec2aa8e13a072c5174920607b09e7c5be1b790a

components/playground/instance/tiflash.go
JaySon-Huang: got it

JaySon-Huang (approved):
LGTM

breezewish:
Just curious, in what case do we need to override the runtime config in the config file? AFAIK for TiKV and TiDB these runtime configurations (like port) are also specified via parameters so that it will not be overridden by config files.

Lloyd-Pottiger:
> Just curious, in what case do we need to override the runtime config in the config file? AFAIK for TiKV and TiDB these runtime configurations (like port) are also specified via parameters so that it will not be overridden by config files.

case like:

My tiup installed in /data1, but /data1 is full and /data2 is almost empty

PR #2353, opened 2024-01-08
Turn off THP added to boot

WalterWj:
Troubleshoot the problem that a large page of the server opens after the server restarts

Start The large page is opened：
image

tiup check apply：
image

the config:
image

You can see that THP has been turned off, and the shutdown is written into the Open startup item

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/pkg/cluster/manager/check.go
+++ b/pkg/cluster/manager/check.go
@@ -669,7 +669,13 @@ func fixFailedChecks(host string, res *operator.CheckResult, t *task.Builder) (s
 			fmt.Sprintf(`if [ -d %[1]s ]; then echo never > %[1]s/enabled; fi`, "/sys/kernel/mm/transparent_hugepage"),
 			"",
 			true)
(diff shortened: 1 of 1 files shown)


breezewish (approved):
LGTM

kaaaaaaang:
I think the automatic modification of /etc/rc.local has too much intrusiveness and may not be suitable for all operating systems and customer environments.

WalterWj:
> I think the automatic modification of /etc/rc.local has too much intrusiveness and may not be suitable for all operating systems and customer environments.

How about a judgment in the code? Do not modify this section if the rc.local file does not exist?

kaaaaaaang:
https://github.com/pingcap/tiup/issues/2360

PR #2354, opened 2024-01-08
cluster: fix tiproxy config and version

xhebox:

1, `require-backend-tls` is useless
2. now we use `nightly` version by default
3. fix a bug where configs are lost

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```


kaaaaaaang:
I do not think it is a good idea to use nightly by default in tiup-cluster

xhebox:
> I do not think it is a good idea to use nightly by default in tiup-cluster. For example，user may use tiup-cluster v1.14.1 years later.

Addressed. Fallback to `""` by default. Or "nightly", if either component version or cluster version is.

PR #2378, opened 2024-02-27
Optimized the active judgment command

hackersean:
when use GNU bash, version 5.2.21(1)-release (x86_64-redhat-linux-gnu) scale-out will be failed
```
[tidb@d86d4890486f ~]$ sudo -H bash -c "systemctl daemon-reload && if [[ $(systemctl is-active prometheus-9090.service) == "active" ]]; then systemctl reload prometheus-9090.service; fi"
Failed to connect to bus: No such file or directory
bash: -c: line 1: conditional binary operator expected
bash: -c: line 1: syntax error near `active'
bash: -c: line 1: `systemctl daemon-reload && if [[  == active ]]; then systemctl reload prometheus-9090.service; fi'
```

Optimized shell statement

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/pkg/cluster/module/systemd.go
+++ b/pkg/cluster/module/systemd.go
@@ -77,7 +77,7 @@ func NewSystemdModule(config SystemdModuleConfig) *SystemdModule {
 		systemctl, strings.ToLower(config.Action), config.Unit)
 
 	if config.CheckActive {
-		cmd = fmt.Sprintf("if [[ $(%s is-active %s) == \"active\" ]]; then %s; fi",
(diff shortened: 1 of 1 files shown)


AstroProfundis:
This looks like an issue of systemd on the target server, maybe your systemd service is not in a normal state or maybe you are using a `--user` scope?

PR #2464, opened 2024-10-28
[DMN] playground: add dm support for playground

siddontang:

Try to support DM in playground.

Notice, this is just a trail, a simple reference for the people who really will do this work later. 

```bash
tiup-playground nightly --dm-master 1 --dm-worker 1
```

create a DM mastesr and worker.

### Check List 

Tests 

Code changes

none

Side effects

none

Related changes

Release notes:

```release-note
Add DM support for TiUP playground
```

### DIFF
--- a/components/playground/instance/dm_worker.go
+++ b/components/playground/instance/dm_worker.go
@@ -1,0 +1,83 @@
+package instance
+
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/pingcap/tiup/pkg/utils"
+)
+
+type DMWorker struct {
+	instance
+	Process
+
+	masters []*DMMaster
+}
+
+var _ Instance = &DMWorker{}
+
+// NewDMWorker create a DMWorker instance.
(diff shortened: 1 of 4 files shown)


components/playground/instance/dm_worker.go
```diff
@@ -0,0 +1,83 @@
...
+
+	w.Process = &process{cmd: PrepareCommand(ctx, w.BinPath, args, nil, w.Dir)}
+
+	logIfErr(w.Process.SetOutputFile(w.LogFile()))
+
+	// try to wait for the master to be ready
+	// this is a very ugly implementation, but it may mostly works
+	// TODO: find a better way to do this,
+	// e.g, let master support a HTTP API to check if it's ready
+	time.Sleep(time.Second * 3)
```
siddontang, line 73: @Benjamin2037 seems that for DM worker, we should support Retry when the worker starts and tries to connect the leader of the Master. 

In the case of Playground, when the master starts, and the worker will also immediately start, and the worker will fail because at that time, there is no leader in DM master.

Benjamin2037: After discussing with the TiUP owner, we decided to use TiUP first to start the DM master and wait for all DM master starting processes to finish, and then TiUP starts the DM worker.

PR #2475, opened 2024-11-12
*: polish the trust key notice

breezewish:

Currently when installing TiUP, a misleading warning will be displayed:

image

Only display the warning, when JSON is accessed via insecure Internet.

And also carefully polished the words, to avoid being mis-interpreted as System root certificate.

### Check List 

Tests 

Release notes:

```release-note
NONE
```

### DIFF
--- a/pkg/localdata/profile.go
+++ b/pkg/localdata/profile.go
@@ -275,9 +275,9 @@ func (p *Profile) ResetMirror(addr, root string) error {
 
 	// Only cache remote mirror
 	if strings.HasPrefix(addr, "http") && root != localRoot {
-		if strings.HasPrefix(root, "http") {
-			fmt.Printf("WARN: adding root certificate via internet: %s\n", root)
-			fmt.Printf("You can revoke this by remove %s\n", localRoot)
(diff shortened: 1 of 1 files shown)


kaaaaaaang (commented):
/lgtm

srstack (approved):

pkg/cluster/template/install/local_install.sh.go
```diff
@@ -81,16 +81,17 @@ fi
 
 chmod 755 "$bin_dir/tiup"
 
-# telemetry is not needed for offline installations
-"$bin_dir/tiup" telemetry disable
```
xhebox, line 85: Should keep

xhebox (commented):
Install test will be fixed after merging

pkg/cluster/template/install/local_install.sh.go
breezewish: I synced it from `install.sh`, since there is no such things in `install.sh`

xhebox: Checked again, telemetry is disabled by default. So we don't need to write it here.

PR #2476, opened 2024-11-13
cluster: Update supported OS versions

dveeden:

`tiup cluster check ...` might report different information than https://docs.pingcap.com/tidb/stable/hardware-and-software-requirements#os-and-platform-requirements

### Check List 

Tests 

Release notes:

```release-note
NONE
```

### DIFF
--- a/pkg/cluster/operation/check.go
+++ b/pkg/cluster/operation/check.go
@@ -159,6 +159,11 @@ func checkSysInfo(opt *CheckOptions, sysInfo *sysinfo.SysInfo) []*CheckResult {
 	return results
 }
 
+// Try to keep this in sync with
+// https://docs.pingcap.com/tidb/stable/hardware-and-software-requirements#os-and-platform-requirements
+//
+// This information is in most cases based on the `ID` (Vendor) and `VERSION_ID` (Release) of /etc/os-release
+// See https://github.com/AstroProfundis/sysinfo/blob/tiup/os.go for details.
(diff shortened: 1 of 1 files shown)


pkg/cluster/operation/check.go
```diff
@@ -175,17 +180,31 @@ func checkOSInfo(opt *CheckOptions, osInfo *sysinfo.OS) *CheckResult {
...
 			return result
 		}
-	case "centos", "redhat", "rhel", "ol":
-		// check version
-		if ver, _ := strconv.ParseFloat(osInfo.Version, 64); ver < 7 {
-			result.Err = fmt.Errorf("%s %s not supported, use version 8 please",
+	case "centos":
+		// CentOS Linux is EOL
+		// CentOS Stream 9 and newer is still fine
+		if ver, _ := strconv.ParseFloat(osInfo.Version, 64); ver < 9 {
```
xhebox, line 199: Looks a little newer to me, maybe `7.3` or so? Or mark `result.warn`?

dveeden: You think we should still list CentOS Linux 7.3 as supported?

xhebox: > You think we should still list CentOS Linux 7.3 as supported?

I just worry existed tiup deployed on `CentOS7` will error.

xhebox (approved):

PR #2472, opened 2024-11-08
cluster: display dashboard server URLs

xhebox:
Signed-off-by: xhe <xw897002528@gmail.com>

### What problem does this PR solve? 

Close #2470

### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test (add detailed scripts or steps below)
 
aae6846c-a798-43d1-adb3-eb99b934c8cb

 - No code

Code changes

 - Has exported function/method change
 - Has exported variable/fields change
 - Has interface methods change
 - Has persistent data change

Side effects

 - Possible performance regression
 - Increased code complexity
 - Breaking backward compatibility

Related changes

 - Need to cherry-pick to the release branch
 - Need to update the documentation

Release notes:

```release-note
NONE
```

PR #2502, opened 2025-01-21
cluster: Remove duplicate display of dashboard URLs

dveeden:

Note the two lines with `Dashboard URL` here:

```
$ tiup cluster display dvetest1 
Cluster type:       tidb
Cluster name:       dvetest1
Cluster version:    v8.5.0
Deploy user:        tidb
SSH type:           builtin
Dashboard URL:      http://192.168.122.196:2379/dashboard
Dashboard URLs:     http://192.168.122.196:2379/dashboard
Grafana URL:        http://192.168.122.196:3000
ID                     Role        Host             Ports        OS/Arch       Status   Data Dir                    Deploy Dir
--                     ----        ----             -----        -------       ------   --------                    ----------
192.168.122.196:3000   grafana     192.168.122.196  3000         linux/x86_64  Up       -                           /tidb-deploy/grafana-3000
192.168.122.196:2379   pd          192.168.122.196  2379/2380    linux/x86_64  Up|L|UI  /tidb-data/pd-2379          /tidb-deploy/pd-2379
192.168.122.196:9090   prometheus  192.168.122.196  9090/12020   linux/x86_64  Up       /tidb-data/prometheus-9090  /tidb-deploy/prometheus-9090
192.168.122.196:4000   tidb        192.168.122.196  4000/10080   linux/x86_64  Up       -                           /tidb-deploy/tidb-4000
192.168.122.196:20160  tikv        192.168.122.196  20160/20180  linux/x86_64  Up       /tidb-data/tikv-20160       /tidb-deploy/tikv-20160
Total nodes: 5
```

Related: #2472

### Check List 

Tests 

Now only one line in the output: `Dashboard URL:      http://192.168.122.196:2379/dashboard`

Release notes:

```release-note
NONE
```

### DIFF
--- a/pkg/cluster/manager/display.go
+++ b/pkg/cluster/manager/display.go
@@ -832,7 +832,7 @@ func (m *Manager) displayDashboards(ctx context.Context, t *spec.Specification, 
 			}
 			if m.logger.GetDisplayMode() == logprinter.DisplayModeJSON && j != nil {
 				j.ClusterMetaInfo.DashboardURL = fmt.Sprintf("%s://%s/dashboard", scheme, addr)
-			} else {
+			} else if len(dashboardAddrs) == 1 {
 				fmt.Printf("Dashboard URL:      %s\n", color.CyanString("%s://%s/dashboard", scheme, addr))
 			}
 		}
@@ -845,7 +845,9 @@ func (m *Manager) displayDashboards(ctx context.Context, t *spec.Specification, 
 	}
 
 	if m.logger.GetDisplayMode() != logprinter.DisplayModeJSON || j == nil {
-		fmt.Printf("Dashboard URLs:     %s\n", strings.Join(dashboardAddrs, ","))
+		if len(dashboardAddrs) > 1 {
+			fmt.Printf("Dashboard URLs:     %s\n", strings.Join(dashboardAddrs, ","))
+		}
 	}
 
 	return nil


pkg/cluster/manager/display.go
```diff
@@ -844,9 +844,5 @@ func (m *Manager) displayDashboards(ctx context.Context, t *spec.Specification,
 		}
 	}
 
-	if m.logger.GetDisplayMode() != logprinter.DisplayModeJSON || j == nil {
-		fmt.Printf("Dashboard URLs:     %s\n", strings.Join(dashboardAddrs, ","))
```
xhebox, line 848: It is a new function that can display multiple dashboard address(dashboard_servers). So it is in valid to remove it.

dveeden: Maybe there should be an `if len(dashboardAddrs) > 1` around it instead? And then also a `if len(dashboardAddrs) == 1`  before line 836 ?

Also, does it use the right `scheme` if there are multiple?

dveeden: PTAL @xhebox

xhebox: Hmm, LGTM. But I don't know if it is wanted or not.

It is a new feature but `Dashboard URL: xxx` is an old feature. There are compatibility issues.

This feature is asked by the team of @sydpz, PTAL.

breezewish: How about displaying multiple dashboard URLs in `Dashboard URL`?

xhebox: > How about displaying multiple dashboard URLs in `Dashboard URL`?

I've considered, and I think it will break existed applications. For example, the team of TEM does rely on that.

dveeden: In the output or in the json structure?

xhebox: Output, of course. I did not modify json structure, iirc.

dveeden: Shouldn't humans rely on the human readable output and programs on the machine readable json? Maybe even over a REST interface instead of by calling a command.

Shouldn't TEM be modified to follow the output of tiup?

Another option would be to do something like git with Porcelain and Plumbing. Or something like `--output v1`...

Issue #2159, opened 2023-04-04
wrong data-dir owner


kaaaaaaang:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?

tiup cluster check xxx.yaml --enable-disk
2. What did you expect to see?

3. What did you see instead?
data-dir was created with root owner

4. What version of TiUP are you using (`tiup --version`)?

Issue #2497, opened 2025-01-21
use user tidb deploy instance, execute init error


geeklc:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?

use user tidb deploy instance。init operate error。

tiup cluster start tidb-test --init

execute this command, while show error：

![Image](https://github.com/user-attachments/assets/5e942f46-365c-434c-b0e0-c9c1c007fffe)

2. What did you expect to see?

3. What did you see instead?

and i get this error from pd logs：

![Image](https://github.com/user-attachments/assets/ab742fe0-8a2b-4780-9544-db5c6880bcfd)

execute command ll 

![Image](https://github.com/user-attachments/assets/c825801c-387e-4f81-9e80-25427565930c)

my topology.yaml is：

![Image](https://github.com/user-attachments/assets/8eb32b33-fefa-4128-9566-ed20121b1bde)

4. What version of TiUP are you using (`tiup --version`)?

![Image](https://github.com/user-attachments/assets/10a2ebb9-7d6b-458a-9a3b-cf04d61e2e56)

geeklc:
操作系统为：kylin10sp2arm

geeklc:
detail to see https://asktug.com/t/topic/1038534

xhebox:
Closed by #2511

PR #2511, opened 2025-02-18
cluster: remove created directories after fio test in `tiup cluster check`

djshow832:
Close https://github.com/pingcap/tiup/issues/2159, https://github.com/pingcap/tiup/issues/2497
In `tiup cluster check topology.yaml --enable-disk=true`, the parent directories of `tiup-fio-test` are created with root but are not removed after the test. The subsequent operations will fail because the directories exist but the user doesn't have permission.

Before creating `tiup-fio-test`, check which parent directory doesn't exist and should be removed.
After testing, remove the parent directory.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

1. Remove `/home/xxx/data` and execute `tiup cluster check --enable-disk=true`. After the test, check that `/home/xxx/data` is removed.
2. Create ``/home/xxx/data` and execute `tiup cluster check --enable-disk=true`. After the test, check that `/home/xxx/data` still exists but it is empty.

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
- Fix that the newly created directory is not removed after `tiup cluster check --enable-disk=true`.
```

### DIFF
--- a/pkg/cluster/task/check.go
+++ b/pkg/cluster/task/check.go
@@ -314,7 +314,7 @@ func (c *CheckSys) runFIO(ctx context.Context) (outRR []byte, outRW []byte, outL
 	// cleanup
 	dirToRm := testWd
 	if !checkDirExists {
-		dirToRm = filepath.Join(checkDir, testWd)
+		dirToRm = checkDir
 	}
 	_, stderr, err = e.Execute(
 		ctx,
--- a/pkg/cluster/task/check.go
+++ b/pkg/cluster/task/check.go
@@ -16,7 +16,6 @@ package task
 import (
 	"context"
 	"fmt"
-	"os"
 	"path/filepath"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


pkg/cluster/task/check.go
```diff
@@ -188,6 +188,19 @@ func (c *CheckSys) runFIO(ctx context.Context) (outRR []byte, outRW []byte, outL
 	testWd := filepath.Join(checkDir, "tiup-fio-test")
 	fioBin := filepath.Join(CheckToolsPathDir, "bin", "fio")
 
+	notExistDir := testWd
+	for {
+		parent := filepath.Dir(notExistDir)
```
xhebox, line 193: Is not the parent `checkDir`? Can not we just use `checkDir`?

djshow832: checkDir is the `data/pd` or `data/tikv` directory in my case, but typically I need to remove `data`.

PR #2513, opened 2025-03-06
cluster: prompt for instance restart

dveeden:

When restarting TiDB nodes in rapid succession and/or concurrently this can cause disruption for applications due to:
- Cold caches
- Loadbalancers that need to do health checks
- Transaction retries for aborted transactions

This adds a `--restart-timeout` option to `tiup cluster upgrade`. 

This then causes tiup to wait after instance restarts for either this timeout or a key press.

This allows the person that does the upgrade to verify the host that was restarted has become healthy before continuing. 

This changes the info that is shown at the start to show the concurrency  to make this more obvious to users.

### Check List 

Tests 

Code changes

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
A `--restart-timeout` option was added to allow more control over restart speed
```


dveeden:
With a `--restart-timeout 2m`:
```
# ps -C tidb-server -o cmd,stime
CMD                         STIME
bin/tidb-server -P 4000 --s 12:00
bin/tidb-server -P 4001 --s 12:02
bin/tidb-server -P 4002 --s 12:04
```

xhebox (approved):

xhebox:
Fix CI plz @dveeden

PR #2521, opened 2025-03-25
cluster: wait for PD finish loading regions after PD restart or upgrade

### LINKED ISSUE #2461: Ensure pd leader is ready to serve when rolling update pd
The PD leader needs to reload all regions' meta data into memory before it can serve, therefor for a big cluster, this process can take a relatively long time, such as 100 seconds or more. 
During rolling update, if we transfer the pd leader to a  PD node that has just been started, before load-region success, pd cannot provide services.
Here we suggest to wait some minutes before we try to transfer pd leader to a pd node that has just been started（restart the PD leader Node）.

djshow832:
close #2461

When checking PD health, also check pd/api/v2/ready to see if regions are loaded.
This applies to both upgrade and reload.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
- Wait for PD finish loading regions after PD restart or upgrade
```

### DIFF
--- a/pkg/cluster/api/pdapi.go
+++ b/pkg/cluster/api/pdapi.go
@@ -141,6 +141,7 @@ var (
 	pdRemoveTombstone    = "pd/api/v1/stores/remove-tombstone"
 	pdRegionsCheckURI    = "pd/api/v1/regions/check"
 	pdServicePrimaryURI  = "pd/api/v2/ms/primary"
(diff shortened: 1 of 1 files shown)


xhebox:
Resolved now

Issue #2508, opened 2025-02-06
DEB Package


dveeden:

ottok:
This is a subtask of #2506.

I am currently a bit baffled about these module dependencies that point to a subdirectory inside a specific branch in a module:
```
src/github.com/pingcap/tiup/components/client/main.go:27:2: cannot find package "github.com/gizak/termui/v3" in any of:
	/tmp/test/tiup/debian/build/src/github.com/pingcap/tiup/vendor/github.com/gizak/termui/v3 (vendor tree)
	/usr/lib/go-1.24/src/github.com/gizak/termui/v3 (from $GOROOT)
	/tmp/test/tiup/debian/build/src/github.com/gizak/termui/v3 (from $GOPATH)
src/github.com/pingcap/tiup/components/client/main.go:28:2: cannot find package "github.com/gizak/termui/v3/widgets" in any of:
	/tmp/test/tiup/debian/build/src/github.com/pingcap/tiup/vendor/github.com/gizak/termui/v3/widgets (vendor tree)
	/usr/lib/go-1.24/src/github.com/gizak/termui/v3/widgets (from $GOROOT)
	/tmp/test/tiup/debian/build/src/github.com/gizak/termui/v3/widgets (from $GOPATH)
src/github.com/pingcap/tiup/components/client/main.go:32:2: cannot find package "github.com/xo/usql/drivers/mysql" in any of:
	/tmp/test/tiup/debian/build/src/github.com/pingcap/tiup/vendor/github.com/xo/usql/drivers/mysql (vendor tree)
	/usr/lib/go-1.24/src/github.com/xo/usql/drivers/mysql (from $GOROOT)
	/tmp/test/tiup/debian/build/src/github.com/xo/usql/drivers/mysql (from $GOPATH)
src/github.com/pingcap/tiup/components/client/main.go:33:2: cannot find package "github.com/xo/usql/env" in any of:
	/tmp/test/tiup/debian/build/src/github.com/pingcap/tiup/vendor/github.com/xo/usql/env (vendor tree)
	/usr/lib/go-1.24/src/github.com/xo/usql/env (from $GOROOT)
	/tmp/test/tiup/debian/build/src/github.com/xo/usql/env (from $GOPATH)
src/github.com/pingcap/tiup/components/client/main.go:34:2: cannot find package "github.com/xo/usql/handler" in any of:
[...]

Issue #123, opened 2020-04-08
Be able to run commands without network access


nrc:
For CI and other automation, it is nice to be able to run commands and be able to guarantee they will not use the network. This implies nothing should be updated (local repositories are a grey area).

lonng:
@nrc Can the local mirror (`export TIUP_MIRRORS=/path/to/local`) resolve this issue?

nrc:
> Can the local mirror (export TIUP_MIRRORS=/path/to/local) resolve this issue?

Maybe. It depends on the specific ways that we design TiUp, e.g., if there is a manifest in the local mirror which can reference an online resource, then this isn't enough (I'm not sure if or why we would want that).

lonng:
This is why our manifest doesn't contain the URL of resources. 

> I'm not sure if or why we would want that

Do you mean why we need offline or other else?

nrc:
> Do you mean why we need offline or other else?

I meant why we'd want a reference to an online repo from an offline repo

lucklove:
We can add a `--offline` flag for TiUP, once it's enabled, TiUP and it's components should never access the network. In this case, if something missing(tarball, manifest), just report an error

innerr:
Still need it.

Here is a crying face, an angry emoj, and a begging picture.

ottok:
While working on #2508 I too have been struggling with this. I have been playing around with disabling `cmdCheckUpdate()`, which currently is a mandatory check at every command invocation, and it needs network. In the context of #2508, the program `tiup` is already installed and having a mechanism that checks for the latest version and downloads updates on every command invocation is excessive.

PR #2538, opened 2025-05-22
WIP: Skip online repo interaction for packaged build

ottok:
**THIS IS WORK-IN-PROGRESS, DO NOT MERGE**

This is not functional yet. E.g. `tiup list` segfaults.

However, with this applied, one can run `tiup --help` without triggering forces network lookups.

-----

Introduce boolean flag `IsPackagedBuild` in the environment package. This flag is intended to be set to `true` in builds of Debian, Fedora, Homebrew etc packages.

This change allows TiUP to run when installed via an external package manager. This also allows TiUP to run without requiring network access to the TiUP repository for initial setup or updates.

When this flag is true:

- The initial environment setup skips connecting to the online repository.

- The automatic update check before running a component is skipped.

- A message is printed to stderr indicating that online interaction is skipped.

- Functions related to updating components or self-updating TiUP return an error, as these operations are disabled in the package build.

See #123 and #2508.

See commit messages for details.

### Check List **TODO**

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

Release notes:

```release-note
NONE
```

### DIFF
--- a/pkg/environment/env.go
+++ b/pkg/environment/env.go
@@ -33,6 +33,10 @@ import (
 var (
 	// ErrInstallFirst indicates that a component/version is not installed
 	ErrInstallFirst = errors.New("component not installed")
+
+	// IsPackagedBuild is a flag to indicate if this is a packaged build (e.g., Debian, RPM, Homebrew).
+	// This should be set to true by the packaging process.
+	IsPackagedBuild = true // Temporarily set to true for testing packaged build behavior
 )
 
 // Mirror return mirror of tiup.
@@ -79,6 +83,16 @@ type Environment struct {
 func InitEnv(options repository.Options, mOpt repository.MirrorOptions) (*Environment, error) {
 	if env := GlobalEnv(); env != nil {
 		return env, nil
+	}
+
+	// Check if running in packaged build mode
+	if IsPackagedBuild {
+		fmt.Fprintln(os.Stderr, "Online version check and repository interaction skipped in packaged build.")
+		// Return a minimal environment that won't try to use the network repo
+		profile := localdata.InitProfile()
+		// We return a valid Environment struct but with v1Repo set to nil.
+		// Functions that use v1Repo must check for nil or IsPackagedBuild.
+		return &Environment{profile: profile, v1Repo: nil}, nil
 	}
 
 	initRepo := time.Now()
@@ -135,6 +149,9 @@ func (env *Environment) LocalPath(path ...string) string {
 
 // UpdateComponents updates or installs all components described by specs.
 func (env *Environment) UpdateComponents(specs []string, nightly, force bool) error {
+	if IsPackagedBuild {
+		return errors.New("component updates are disabled in this packaged build")
+	}
 	var v1specs []repository.ComponentSpec
 	for _, spec := range specs {
 		component, v := ParseCompVersion(spec)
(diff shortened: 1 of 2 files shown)


ottok:
@dveeden Could you suggest what is the best entry-point to this? I've read `docs/dev/README.md` etc but struggling with this. Initially I though it would be enough to just force `func cmdCheckUpdate` to bail out immediately, but seems the code base has a lot of assumptions regarding forces binary upgrades.

dveeden:
I'll have a look at this tomorrow as I'm attending a conference today

xhebox:
Maybe it will be easy to change mirror to local path(yes, we can set it to local filesystem) automatically. Then adding a fallback for unexisted localpath.

ottok:
What do you mean by local mirror? Which function/code path did you have in
mind?

I am trying to avoid using any mirror or copying binaries into users home
directories. Instead the existing system binaries should be used as-is.

xhebox:
> What do you mean by local mirror? Which function/code path did you have in mind? I am trying to avoid using any mirror or copying binaries into users home directories. Instead the existing system binaries should be used as-is.

It is possible to set `tiup mirror set /home/xxxx`, as long as `/home/xxx` is a valid mirror directory(containing manifests or so). In such case, `tiup` will not access internet.

If you hack through this path, e.g. `IsPackagedBuild ` will execute `tiup mirror set /unexisted_virtual_path` automatically. And `V1Repo` also handle this `/unexisted_virtual_path`, it will be more consistent. And it should not need to call `IsPackageBuild` everywhere. IMO, better than adding another layer on top of `V1Repo`.

BTW: `V2Repo` is under development currently. I hope eventually these parts can be simplified.

ottok:
> I'll have a look at this tomorrow as I'm attending a conference today

Could you please comment / provide guidance so it is easier for me to finish this with the optimal approach?

dveeden:
> > I'll have a look at this tomorrow as I'm attending a conference today
> 
> Could you please comment / provide guidance so it is easier for me to finish this with the optimal approach?

I think the local mirror that @xhebox  suggested would probably be a good approach. 

Disabling the version update check would be good.

However if we disable access to online repository info and components then the functionality of the tool is very limited.

Should we consider shipping a functional local mirror with the package as data?

ottok:
Isn't the mirror used to fetch the components that get installed in `~/components/` which include grafana, pd, playground, prometheus, tidb, tiflash and tikv. If we completely disable the mirror or use a local mirror that is empty, then users can't get any of these components, right?

That seems overkill. I just want to bypass the mandatory version check and download for the binaries that are already in the package. Running `tiup list` or `tiup client` should not have a mandatory mirror check. Those commands should just run and skip that stage.

I am currently testing Daniël's patch.

dveeden:
@ottok Yes, I agree with that.

ottok:
@dveeden Any comments about the overall direction of the latest version of this PR?

ottok:
There are a few places in the provided code where a nil pointer dereference (segfault) could occur when the IsPackagedBuild flag is true. When IsPackagedBuild is true, the v1Repo field in the Environment struct is explicitly set to nil. Any method call on this nil interface would then cause a panic.

I have an improved version of this commit on my Debian packaging branch, and I will update it here once I get the Debian packaging stabilized.

PR #2547, opened 2025-06-25
Update to github.com/r3labs/diff/v3

dveeden:

Follow up for https://github.com/pingcap/tiup/commit/85b624234d7fe44394e6c350e318211a901cbcab

This makes it easier to package tiup as debian package, see also #2508 

### Check List 

Tests 

 - [x] No code

Release notes:

```release-note
NONE
```


breezewish (approved):
Thank you!

PR #2558, opened 2025-07-11
Switch from yaml.v2 to yaml.v3

dveeden:

- Removed the replace rule in `go.mod`
- Upgraded from yaml.v2 to yaml.v3
- Both yaml.v2 and yaml.v3 are currently unmaintained
- ~~Naively replaced `UnmarshalStrict` with `Unmarshal`~~. See also https://github.com/go-yaml/yaml/issues/602 

### Check List 

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Code changes

Side effects

Related changes

Release notes:

```release-note
NONE
```


pkg/cluster/spec/validate_test.go
```diff
@@ -346,7 +346,7 @@ func (s *metaSuiteTopo) TestCountDir2(c *C) {
 	meta := ClusterMeta{}
 	yamlFile, err := os.ReadFile(file)
 	c.Assert(err, IsNil)
-	err = yaml.UnmarshalStrict(yamlFile, &meta)
+	err = yaml.Unmarshal(yamlFile, &meta)
```
xhebox, line 349: https://github.com/apptainer/apptainer/issues/473#issuecomment-1545697498

We better keep the strict behavior(it is used in cluster edit-config), something like this would do.

PR #2565, opened 2025-07-26
Remove ansible import functionality

dveeden:
### What problem does this PR solve? 

ref #2508

### Check List 

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] Manual test (add detailed scripts or steps below)
- [ ] No code

Code changes

- [ ] Has exported function/method change
- [ ] Has exported variable/fields change
- [ ] Has interface methods change
- [ ] Has persistent data change

Side effects

- [ ] Possible performance regression
- [ ] Increased code complexity
- [ ] Breaking backward compatibility

Related changes

- [ ] Need to cherry-pick to the release branch
- [x] Need to update the documentation

Release notes:

```release-note
Ansible import functionality has been removed
```

PR #2630, opened 2025-10-30
cluster: cleanup ansible import related things

dveeden:

Follow-up for https://github.com/pingcap/tiup/pull/2565 

This reduces a lot of complexity and code that isn't used or tested often.

Remove more Ansible import related things.

- This is keeping the `Imported` field, but sets the key to `-`.

Questions
Should we check if the imported field is set in the config and throw an error if that's the case?

### Check List 

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Code changes

- [x] Has exported function/method change
- [x] Has exported variable/fields change
- [x] Has persistent data change

Side effects

- [x] Breaking backward compatibility

Related changes

- [x] Need to update the documentation

Release notes:

```release-note
Clusters imported from an ansible installation are not supported anymore
```


dveeden:
@AstroProfundis thanks for the info. This is why I checked "Breaking backward compatibility" and set the release note to mention this.

Issue #2654, opened 2025-12-17
Should not treat version as float to compare


lance6716:
https://github.com/pingcap/tiup/blob/e525ada49ad0223167b9fd7e5902f568e5eca3c4/pkg/cluster/operation/check.go#L167-L250

This function treats OS versions as float. However, version 8.10 is larger than version 8.4, while float 8.10 is not.

4. What version of TiUP are you using (`tiup --version`)?

latest

PR #2656, opened 2025-12-17
treat version as string instead of float when compare

ghoshh:

ref #2654 

add a function to compare versions by splitting the version into integers to avoid the case like 8.4 > 8.10.

### Check List 

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Code changes

Side effects

Related changes

Release notes:

```release-note
NONE
```


pkg/cluster/operation/check.go
```diff
@@ -1020,3 +1020,30 @@ func CheckTimeZone(ctx context.Context, topo *spec.Specification, host string, r
 	}
 	return results
 }
+
+// compareVersion compares two version strings v1 and v2.
+// It returns 1 if v1 > v2, -1 if v1 < v2, and 0 if v1 == v2.
+func CompareVersion(v1, v2 string) int {
```
lance6716: 1. no need to be public function
2. add unit test to verify its behaviour

PR #2666, opened 2026-01-13
playground: fix config in CSE/NextGen modes

Lloyd-Pottiger:
This PR fixes a few playground configuration issues for CSE/NextGen modes.

- Record the bound component version on playground instances when `binpath != ""`
- Add missing CSE TiDB options (columnar store type, fulltext index, TiFlash addresses)
- Allow NextGen system TiDB to use multiple tikv-worker endpoints


breezewish:
We now have a https://github.com/pingcap/tiup/tree/master/components/playground-ng could you try to fix over it as well? :)

Issue #2325, opened 2023-12-04
cluster: Add option to configure ng-monitoring continuous_profiling


dveeden:
## Feature Request

When using TiUP Cluster to deploy a TiDB Cluster one can set the `ng_port` to get ng-monitoring deployed as part of the cluster. Then one can enable continuous profiling as described [here](https://docs.pingcap.com/tidb/stable/continuous-profiling#enable-continuous-profiling). However to automate deployment and make this persist on restarts of the cluster and cluster components it would be good to configure this in the configuration in the same way other cluster config is handled.

For ng-monitoring:
- There is a [`continuous_profiling` section](https://github.com/pingcap/ng-monitoring/blob/f223e14cc614cded58ec1f80be55d38845cee16e/config/config.go#L42) in the config.
- In [this section](https://github.com/pingcap/ng-monitoring/blob/f223e14cc614cded58ec1f80be55d38845cee16e/config/config.go#L403-L409) one can enable continuous profiling and set a retention time.

For tiup:
- The [template for ngmonitoring.toml](https://github.com/pingcap/tiup/blob/master/embed/templates/config/ngmonitoring.toml.tpl) doesn't have a `continuous_profiling` section.
- There doesn't seem to be a section for continuous profiling in [`NgMonitoringConfig`](https://github.com/pingcap/tiup/blob/c53895ca4f5ca58f70beb9bfe58a8956b44ab49e/pkg/cluster/template/config/ngmonitoring.go#L26-L36)

The workaround for now is to:
- Manually enable continuous profiling via the dashboard
- Send a HTTP POST to the dashboard or ng-monitoring service
- Possibly do this after cluster and/or component restarts.

```
 curl -s -X POST  -d '{"continuous_profiling":{"enable":true, "data_retention_seconds":259200}}' http://127.0.0.1:12020/config | jq .
{
  "status": "ok"
}
```

boltandrke:
This also affects configuring `storage.type` and `storage.path` for ng-monitoring.

For example, there is currently no way to set via topology YAML:

```toml
[storage]
type = "sqlite"
path = "/tidb-data/prometheus-9090/docdb_sqlite"
```

A general solution would be adding `ng_monitoring` to `server_configs` (similar to how tidb/tikv/pd configs work), so any ng-monitoring config can be set via topology YAML.

[...]

PR #2685, opened 2026-02-21
cluster: support ng-monitoring config via server_configs and per-instance overrides

boltandrke:
## What problem does this PR solve

Adds support for configuring ng-monitoring via topology YAML, addressing
the gap described in #2325.

Currently there is no way to set ng-monitoring configuration (e.g.
`storage.type`, `storage.path`, `continuous_profiling.*`) through
the topology YAML. The `NgMonitoringConfig` template is hardcoded with
no user-config merge step, unlike other components (PD/TiKV/TiDB) which
support `server_configs` + per-instance `config` overrides.

## What is changed and how it works

1. **`ServerConfigs.NGMonitoring`** — added `ng_monitoring map[string]any` to `ServerConfigs` for global ng-monitoring config
2. **`PrometheusSpec.NgMonitoringConfig`** — added `ng_monitoring_config map[string]any` for per-instance overrides
3. **`MonitorInstance.InitConfig`** — replaced template-based ngmonitoring.toml generation with map-based approach using `Merge2Toml`, following the same pattern as PD/TiKV/TiDB
4. **Example topology** — updated with documented `ng_monitoring` and `ng_monitoring_config` sections

### Usage

```yaml
server_configs:
  ng_monitoring:
    storage.type: "sqlite"
    storage.path: "/tidb-data/prometheus-9090/docdb_sqlite"
    continuous_profiling.enable: true
    continuous_profiling.profile_seconds: 5

monitoring_servers:
  - host: 10.0.1.21
    ng_port: 12020
    ng_monitoring_config:
      continuous_profiling.data_retention_seconds: 259200
```

## Check List

- [x] Code compiles (`go build ./...`)
- [x] All existing tests pass (`go test ./pkg/cluster/spec/...`)
- [x] New test added (`TestNGMonitoringServerConfig`)
- [x] Lint passes (`make check` — 0 issues)

Ref #2325

### DIFF
--- a/pkg/cluster/spec/server_config_test.go
+++ b/pkg/cluster/spec/server_config_test.go
@@ -69,7 +69,7 @@ monitoring_servers:
 		"advertise-address": "10.0.1.21:12020",
 		"log.path":          "/tidb-deploy/prometheus-9090/log",
 		"log.level":         "INFO",
-		"pd.endpoints":      []string{"10.0.1.10:2379"},
+		"pd.endpoints":      []string{"10.0.1.10:2379", "10.0.1.11:2379"},
 		"storage.path":      "/tidb-data/prometheus-9090",
 	}
 
@@ -87,6 +87,9 @@ monitoring_servers:
 	require.Contains(t, tomlStr, `enable = true`)                   // from global
 	require.Contains(t, tomlStr, `data_retention_seconds = 259200`) // from per-instance
 	require.Contains(t, tomlStr, `address = "0.0.0.0:12020"`)       // from base
+
+	// pd.endpoints must be a TOML array of quoted strings (matches old template behavior)
+	require.Contains(t, tomlStr, `endpoints = ["10.0.1.10:2379", "10.0.1.11:2379"]`)
 }
 
 func TestGetValueFromPath(t *testing.T) {


pkg/cluster/spec/monitoring.go
```diff
@@ -486,24 +487,37 @@ func (i *MonitorInstance) InitConfig(
...
-			LogDir:    paths.Log,
+		// Build base ng-monitoring config as a map so user overrides via
+		// server_configs.ng_monitoring and per-instance ng_monitoring_config
+		// are merged on top (same pattern as PD/TiKV/TiDB).
+		baseConfig := map[string]any{
+			"address":           utils.JoinHostPort(i.GetListenHost(), spec.NgPort),
+			"advertise-address": utils.JoinHostPort(i.GetHost(), spec.NgPort),
+			"log.path":          paths.Log,
+			"log.level":         "INFO",
+			"pd.endpoints":      pdAddrs,
```
dveeden:
```suggestion
			"pd.endpoints":      strconv.Quote(pdAddrs),
```

Maybe adding a quote here and using [`strconv.Quote()`](https://pkg.go.dev/strconv#Quote) might be better

boltandrke:
Thanks for the review @dveeden!

Regarding the `pd.endpoints` question: the rename from `pds` to `pdAddrs` is intentional — the old code built manually-quoted strings (`fmt.Sprintf("\"%s\"", addr)`) because the template interpolated them directly into `endpoints = [{{.PDAddrs}}]`. The new code collects plain addresses as a `[]string` because the TOML encoder (via `Merge2Toml`) handles serialization and quoting automatically.

`strconv.Quote()` would not work here since `pdAddrs` is `[]string`, not `string`. The Go TOML encoder already serializes `[]string{"10.0.1.10:2379", "10.0.1.11:2379"}` into the correct TOML array format:

```toml
[pd]
endpoints = ["10.0.1.10:2379", "10.0.1.11:2379"]
```

This matches the output produced by the old template.

I have pushed a new commit adding an explicit test assertion that verifies this behavior with multiple PD addresses, confirming the serialization is correct.

PR #2709, opened 2026-06-04
fix: auto-bootstrap root.json and create bin dir on first run

yahonda:

When tiup is installed via mise (https://mise.jdx.dev, a polyglot tool version manager similar to asdf) with:

    mise install github:pingcap/tiup
    mise use -g github:pingcap/tiup

the install.sh script is not executed, which means:
1. ~/.tiup/bin/ directory is not created
2. root.json is not downloaded from the mirror

This leaves the installation incomplete and requires the user to manually run 'tiup mirror set' before any tiup command works.

This PR fixes both issues:
- ResetMirror now ensures the bin directory exists before writing root.json
- InitEnv automatically bootstraps root.json from the default mirror if it's missing, so no manual 'tiup mirror set' is needed

In the future, we would like to register tiup in the mise registry so that users can simply run:

    mise install tiup
    mise use tiup

This fix is a prerequisite for that, as it removes the need for any post-install initialization steps.

### Check List

Tests
- [ ] Unit test
- [ ] Integration test
- [ ] No code

Code changes
- [x] Has exported function/method change

Side effects

Related changes

### DIFF
--- a/pkg/cluster/manager/check.go
+++ b/pkg/cluster/manager/check.go
@@ -686,7 +686,10 @@ func fixFailedChecks(host string, res *operator.CheckResult, t *task.Builder, sy
 	case operator.CheckNameTHP:
 		t.Shell(host,
 			fmt.Sprintf(
-				`if [ -d %[1]s ]; then echo never > %[1]s/enabled; fi && %s`,
+				// grubby only exists on RHEL-family distros; skip the persistent
+				// kernel argument when it's not available (e.g. Debian/Ubuntu)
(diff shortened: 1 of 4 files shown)


yahonda:
To isolate whether the CI failures are caused by this PR's changes or by pre-existing issues on master, this branch was rebased onto #2712 (SELinux/THP fixes for non-RHEL hosts). The 2 commits were also squashed into one. Once #2712 lands, I will rebase onto master again.

PR #2723, opened 2026-07-27
repository, tests: update testdata root.json to v7

bb7133:

Unit-test CI started failing on 2026-07-26 (e.g. in #2721, which touches unrelated files):

- `TestPollutedManifest` (both `pkg/repository` and `pkg/repository/v1manifest`) fails deterministically: the trusted root embedded in the test fixtures is a copy of the production **v6 root.json, which expired at 2026-07-26T11:18:30+08:00**. With the root expired, the trust chain breaks and the 'correct' `snapshot.json` in the polluted fixture can no longer be verified.
- The `TestUpstream` panic in the same run was transient: it fetches `timestamp.json` from the live mirror and hit the root-rotation/CDN-purge window; `assert.Nil` does not stop execution, so the subsequent `m.Signed` dereference panics on the nil manifest. It passes on re-run (verified locally).

Replace all copies of the production `root.json` in the repo with the rotated **v7** root (expires 2027-07-26). v7 only rotates the root role itself; sub-role keys are unchanged, so existing signed fixtures verify as-is:

- `pkg/repository/v1manifest/testdata/polluted/bin/root.json` (was v6)
- `pkg/repository/testdata/polluted/bin/root.json` (was v6)
- `pkg/repository/testdata/manifests/root.json` (was v6)
- `tests/tiup-cluster/root.json`, `tests/tiup-dm/root.json`, `tests/tiup/bin/root.json` (long-expired v4/v1 copies used by integration tests — refreshed to avoid the same trap later)

### Check List

Tests:
- Unit test: `go test ./pkg/repository/...` all green locally after the change (fails on master without it)

### Release notes
```release-note
NONE
```


zanmato1984 (approved):

bb7133:
Superseded by #2724 (same change rebased onto latest master; this PR's stale fork base caused merge conflicts).

PR #2732, opened 2026-08-27
cluster: fix Grafana VM datasource replacement

Defined2014:

close #xxx

### Check List 

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Code changes

Side effects

Related changes

Release notes:

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed dashboard provisioning so cluster names and datasource references are replaced correctly.
  * Prevented datasource names containing cluster names from being rewritten twice.
  * Ensured dashboard annotations and datasource configurations consistently use the provisioned datasource name.

* **Tests**
  * Improved coverage for dashboard provisioning and datasource replacement scenarios.

### DIFF
--- a/pkg/cluster/spec/grafana_test.go
+++ b/pkg/cluster/spec/grafana_test.go
@@ -282,6 +282,18 @@ func TestVictoriaMetricsDefaultDatasource(t *testing.T) {
 	err = os.MkdirAll(binDir, 0755)
 	require.NoError(t, err)
 
+	datasourceName := "test-cluster-vm"
+	dashboardReplacements := []struct {
+		commandPattern string
+		placeholder    string
+	}{
+		{`s/test-cluster/`, "test-cluster"},
+		{`s/Test-Cluster/`, "Test-Cluster"},
+		{`s/\${DS_.*-CLUSTER}/`, "${DS_TEST-CLUSTER}"},
+		{`s/DS_.*-CLUSTER/`, "DS_TEST-CLUSTER"},
+		{`s/\${DS_LIGHTNING}/`, "${DS_LIGHTNING}"},
+		{`s/DS_LIGHTNING/`, "DS_LIGHTNING"},
+	}
 	// Create a mock for the execute function to handle the dashboard copy command
 	origExecutor := &mockExecutor{
 		executeFunc: func(ctx context.Context, cmd string, sudo bool, timeouts ...time.Duration) ([]byte, []byte, error) {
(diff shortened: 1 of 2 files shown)


nolouch (approved):

pkg/cluster/spec/grafana_test.go
```diff
@@ -282,60 +283,27 @@ func TestVictoriaMetricsDefaultDatasource(t *testing.T) {
...
-							return nil, nil, err
-						}
-					}
-				}
+			if !strings.Contains(cmd, dashboardsDir) {
+				return nil, nil, nil
 			}
-			return nil, nil, nil
+
+			output, err := osexec.CommandContext(ctx, "sh", "-c", cmd).CombinedOutput()
```
bb7133: [P1] Keep this test portable across supported developer platforms. This executes the Linux-targeted dashboard command on the host. On macOS, BSD `sed -i` requires a backup suffix, unlike GNU sed; the `find -exec` invocation still exits successfully, all dashboard substitutions remain unchanged, and the assertions below fail. I reproduced this on exact head `6825fcc9` with `go test ./pkg/cluster/spec -run 'TestVictoriaMetricsDefaultDatasource|TestGrafanaDatasourceConfig' -count=1 -v`. Please avoid executing this GNU-specific remote command directly on Darwin (for example, gate the command-level test to Linux explicitly, or use a portable test helper/mock that preserves and verifies the production command order).

bb7133 (changes requested):
The production ordering change fixes the `test-cluster-vm-vm` rewrite, but the new regression test introduces a deterministic failure on macOS. Please make the test portable before merging.

pkg/cluster/spec/grafana_test.go
Defined2014: Done

bb7133 (approved):
Re-reviewed exact head eea1b10b6a914a5080752f18c8525a7aba19d971. The previous portability blocker is resolved: the regression test no longer executes GNU-style sed -i on the developer host, and the mock follows the actual initDashboards command order while verifying the generated datasource and dashboard references. The focused tests, focused race run, full pkg/cluster/spec tests, go vet, and diff check pass on Darwin/arm64. The remaining playground red check is the known 127.0.0.1:9527 startup/control-listener flake also seen unchanged on master, outside this Grafana diff; I reran it separately.

PR #2740, opened 2026-09-08
cluster: use the first PD for standalone tidb-dashboard

mayjiang0203:

When TiUP deploys a standalone TiDB Dashboard to a cluster with multiple PD
nodes, it generates a `--pd` argument containing all PD URLs joined by commas.
TiDB Dashboard treats `--pd` as a single endpoint and does not split the
comma-separated list, so it cannot connect to PD (e.g.
`dial tcp: address ...: too many colons in address`).

Ref: https://github.com/pingcap/tidb-dashboard/issues/1920

Generate the standalone Dashboard startup command with only the first PD URL,
because Dashboard's PD HTTP API client only supports a single `baseURL`. This
is a workaround until Dashboard supports multiple PD endpoints; once it does,
restore joining all PD URLs.

- Add `dashboardPDEndpoint(pdServers, enableTLS)`, which returns the first PD's
  advertised client URL (empty when there are no PD servers).
- Replace `strings.Join(pds, ",")` in `DashboardInstance.InitConfig` with the
  helper.
- After scaling in PD nodes, warn the user when a standalone tidb-dashboard may
  still point to a removed PD endpoint, and show the restart command
  (`tiup cluster restart <name> -R tidb-dashboard`) to pick up the new endpoint.

### Check List 

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Manual test steps:
- Deploy a cluster with multiple PD nodes and a standalone tidb-dashboard, then
  confirm `run_tidb-dashboard.sh` contains a single `--pd` URL and Dashboard can
  connect to PD.
- Scale in the PD node that the standalone tidb-dashboard points to, and confirm
  the scale-in output warns to restart the dashboard.

Code changes

Side effects

Related changes

Release notes:

```release-note
Fix standalone TiDB Dashboard failing to connect to PD when the cluster has
multiple PD nodes, and warn to restart it after scaling in the PD node it
points to.
```

### DIFF
--- a/pkg/cluster/manager/scale_in.go
+++ b/pkg/cluster/manager/scale_in.go
@@ -106,6 +106,14 @@ func (m *Manager) ScaleIn(
 		return err
 	}
 
+	deletedNodes := set.NewStringSet(nodes...)
+	scaledInPD := false
+	topo.IterInstance(func(inst spec.Instance) {
+		if deletedNodes.Exist(inst.ID()) && inst.ComponentName() == spec.ComponentPD {
+			scaledInPD = true
+		}
+	})
+
 	b, err := m.sshTaskBuilder(name, topo, base.User, gOpt)
 	if err != nil {
 		return err
@@ -161,6 +169,15 @@ func (m *Manager) ScaleIn(
 	}
 
 	m.logger.Infof("Scaled cluster `%s` in successfully", name)
+	if scaledInPD {
+		if dash := spec.FindComponent(topo, spec.ComponentDashboard); dash != nil && len(dash.Instances()) > 0 {
+			m.logger.Warnf("%s", color.YellowString(
+				"\nSince PD node(s) were scaled in, the standalone tidb-dashboard connects to a single PD endpoint "+
+					"that may have been removed. If it can no longer reach PD, restart it to pick up a new endpoint:\n\t%s",
+				color.GreenString("%s restart %s -R %s", tui.OsArgs0(), name, spec.ComponentDashboard),
+			))
+		}
+	}
 
 	return nil
 }
--- a/pkg/cluster/spec/dashboard.go
+++ b/pkg/cluster/spec/dashboard.go
@@ -18,7 +18,6 @@ import (
 	"crypto/tls"
 	"fmt"
 	"path/filepath"
-	"strings"
 	"time"
 
 	"github.com/pingcap/tiup/pkg/cluster/ctxt"
@@ -198,10 +197,6 @@ func (i *DashboardInstance) InitConfig(
 	enableTLS := topo.GlobalOptions.TLSEnabled
 	spec := i.InstanceSpec.(*DashboardSpec)
 
-	pds := []string{}
-	for _, pdspec := range topo.PDServers {
-		pds = append(pds, pdspec.GetAdvertiseClientURL(enableTLS))
-	}
 	cfg := &scripts.DashboardScript{
 		// -h, --host string              listen host of the Dashboard Server
 		Host:        i.GetListenHost(),
@@ -211,7 +206,7 @@ func (i *DashboardInstance) InitConfig(
 		LogDir:      paths.Log,
 		Port:        spec.Port,
 		NumaNode:    spec.NumaNode,
-		PD:          strings.Join(pds, ","),
+		PD:          dashboardPDEndpoint(topo.PDServers, enableTLS),
 		TLSEnabled:  enableTLS,
 	}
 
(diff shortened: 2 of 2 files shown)


kaaaaaaang (approved):

zanmato1984 (changes requested):
Thanks for fixing this. The implementation looks reasonable, but this behavior needs automated regression coverage before merge.

This PR changes both the generated standalone Dashboard startup command and the PD scale-in recovery path, while the patch currently has no test coverage. Please add tests covering at least:

- multiple PD servers select only the first advertised client URL;
- an empty PD list returns an empty endpoint;
- TLS URL generation and an explicit advertised client address are preserved;
- scaling in a PD with a standalone Dashboard exercises the regenerated endpoint and warning path;
- scaling in a non-PD node does not emit the Dashboard warning.

If the manager flow is too coupled to test directly, please extract the warning decision into a focused helper and test it together with Dashboard script generation. Manual testing alone will not protect these configuration and recovery semantics from regression.

mayjiang0203:
@zanmato1984 Thanks for the review and the detailed suggestions! Automated tests have been added covering the points you raised.

`pkg/cluster/spec/dashboard_test.go`
- `TestDashboardPDEndpoint`: multiple PD servers select only the first advertised client URL; an empty PD list returns an empty endpoint.
- `TestDashboardScriptSinglePD`: renders the generated `run_tidb-dashboard.sh` and asserts a single `--pd` endpoint (no comma-joined list), and that TLS URL generation and an explicit advertised client address are preserved.

`pkg/cluster/manager/scale_in_test.go`
- Extracted the warning decision into focused helpers (`isScaledInPD`, `hasStandaloneDashboard`) as you suggested.
- `TestIsScaledInPD` / `TestHasStandaloneDashboard`: scaling in a PD with a standalone dashboard exercises the warning path; scaling in a non-PD node does not.

Also verified end-to-end on a live 3-PD cluster: the standalone dashboard is generated with a single `--pd`, and the restart warning is emitted after scaling in the PD it points to.

Please take another look when you have a chance.

zanmato1984 (changes requested):
Thanks for adding the tests. They pass locally and cover the extracted helpers, but two production paths from the previous review are still not exercised:

1. `TestDashboardScriptSinglePD` manually constructs `scripts.DashboardScript` and sets `PD` by calling `dashboardPDEndpoint` itself. It would still pass if `DashboardInstance.InitConfig` stopped wiring the helper into the generated startup script. Please exercise `DashboardInstance.InitConfig` with the existing mock executor (or an equivalent production-level entry point) and assert the transferred `run_tidb-dashboard.sh`.

2. `TestIsScaledInPD` and `TestHasStandaloneDashboard` test the two predicates separately, but they never execute the combined warning branch in `ScaleIn` or assert the actionable restart command. Removing the warning, changing `&&` to `||`, or breaking the command text would leave both tests green. Please expose a focused combined warning decision/message helper or test the manager path, covering both PD + standalone Dashboard (warning emitted with the restart command) and non-PD scale-in (no warning).

The current Codecov report is consistent with these gaps: the `DashboardInstance.InitConfig` wiring line and the scale-in warning block remain uncovered (8 changed lines total). Once those production paths are covered, the original blocking concern should be resolved.

zanmato1984:
To clarify why I am keeping `CHANGES_REQUESTED`: the issue is not the number of test files, but whether they fail when the production path regresses.

### Dashboard startup-script path

`TestDashboardScriptSinglePD` currently constructs its own `scripts.DashboardScript` and repeats the intended wiring in the test:

```go
PD: dashboardPDEndpoint(tt.pds, tt.enableTLS),
```

It never calls `DashboardInstance.InitConfig`, which contains the production wiring. If production were accidentally changed back to a comma-joined PD list, this test would still pass because the test continues to call the helper directly.

Please invoke `DashboardInstance.InitConfig` with the existing `mockExecutor` (or an equivalent production entry point), then inspect the transferred `run_tidb-dashboard.sh`. This makes the test fail when the actual wiring regresses.

### Scale-in warning path

`TestIsScaledInPD` and `TestHasStandaloneDashboard` verify two predicates independently, but do not execute or assert this production behavior:

```go
if scaledInPD && hasStandaloneDashboard(topo) {
    m.logger.Warnf(...)
}
```

The current tests would remain green if the warning block were deleted, `&&` were changed to `||`, or the suggested restart command were broken. Please test a combined warning decision/message helper or capture the manager logger, covering:

- PD + standalone Dashboard: warning contains the expected `restart ... -R tidb-dashboard` command;
- non-PD + Dashboard: no warning;
- PD + no standalone Dashboard: no warning.

The eight uncovered changed production lines reported by Codecov correspond to these two unexercised paths; the coverage number is supporting evidence, not the goal by itself.

mayjiang0203:
@zanmato1984 
Thanks for the clarification. Updated in 7dacc610:
- TestDashboardScriptSinglePD now calls DashboardInstance.InitConfig with the existing mock executor and checks the transferred startup script.
- TestDashboardScaleInWarning exercises the combined decision/message helper used by ScaleIn, checking the restart command for PD scale-in with a standalone Dashboard and no warning for non-PD scale-in or no standalone Dashboard.

The relevant package tests and CI unit tests/lint pass. Could you please take another look?

zanmato1984 (approved):
Re-reviewed at `7dacc610d0d8f7bca7e445312ddb4c0ae583859f`. The previous blocking concerns are resolved:

- the Dashboard test now exercises `DashboardInstance.InitConfig` and inspects the transferred startup script, so production wiring regressions are detected;
- the scale-in warning decision and message are now implemented in a production helper with positive and negative coverage, including the restart command.

`go test -count=10 ./pkg/cluster/spec ./pkg/cluster/manager` passes locally, and the reported CI checks are green.

